PR #21, opened 2016-05-28
Implementation of set/get

romaninsh:
This PR implements:

    $m['foo'] = 123;

functionality of the basic Active Record pattern. While the loading of an active record is not supported yet, you can work with the values already.

For reviewer, please check:

 - [x] Code looks OK
 - [x] Test scripts present
 - [x] Well-documented
 - [x] No outstanding questions or issues or TODO items

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     ],
     "require": {
         "php": ">=5.5.0",
-        "atk4/dsql": "*"
+        "atk4/dsql": "*",
+        "atk4/core": "dev-feature/implement-containers"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
     "require": {
         "php": ">=5.5.0",
         "atk4/dsql": "*",
-        "atk4/core": "dev-feature/implement-containers"
+        "atk4/core": "1.0"
     },
     "require-dev": {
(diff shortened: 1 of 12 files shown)


romaninsh:
I've finished with this PR, so it's fully ready for review and merge. Working on "Persisting" next.

composer.json
```diff
@@ -14,7 +14,8 @@
     ],
     "require": {
         "php": ">=5.5.0",
-        "atk4/dsql": "*"
+        "atk4/dsql": "*",
+        "atk4/core": "1.0"
```
skondakov: Why do we need atk core in agile data?

src/Model.php
```diff
@@ -1,33 +1,329 @@
-<?php
+<?php // vim:ts=4:sw=4:et:fdm=marker
```
skondakov: Maybe these vim settings shouldn't be in git?

skondakov:
Seems good to me, except the 2 comments

composer.json
romaninsh: because we're using implementation for hooks, containers and initializers:

http://agile-core.readthedocs.io/en/latest/

src/Model.php
romaninsh: :) we love vim, right !

romaninsh:
I'm ticking off the check-lists above, @skondakov

PR #26, opened 2016-06-20
Feature/general improvements

romaninsh:
This PR implements several general fixes, such as improving relationship between model and persistence drivers. This PR is necessary before Join implementation.


composer.json
```diff
@@ -15,7 +15,7 @@
     "require": {
         "php": ">=5.5.0",
         "atk4/dsql": "*",
-        "atk4/core": "1.0"
+        "atk4/core": "*"
```
skondakov: So it's no more forked with specific version? Is this ok?

docs/hooks.rst
```diff
@@ -0,0 +1,30 @@
...
+
+.. important::
+
+Please never use ``$this`` inside your hook to refer to the model. The model
+is always passed as a first argument. If you ever use ``$this`` then your
+model will perform very weirdly when cloned::
+
+    $m->addHook('beforeSave', function($m) {
+        $this['name'] = 'John';  // WRONG!
+        $m['surname'] = 'Smith'; // GOOD
```
skondakov: Should be $m['name'] for the purpose of the example I think

skondakov:
Seems good to me overall, just check the comments please and you can push.

composer.json
romaninsh: we need 1.0.1 version, which is now stable. Will keep both repos in sync.

docs/hooks.rst
romaninsh: i re-checked and it actually is correct with the surname.

PR #35, opened 2016-06-30
Implemented SQL conditions

romaninsh:
Finally, implementation of conditions is complete. And It's AWESOME!

resolve #28

PR #36, opened 2016-06-30
Feature/implement relations

romaninsh:
- [x] Accept #35 first
 - [x] wait for hasOne to be added

Moved todo into #37, ready for review.

Relations add traversal between models.

Docs: https://github.com/atk4/data/blob/feature/implement-relations/docs/relations.rst#dealing-with-many-to-many-relations

### DIFF
--- a/docs/relations.rst
+++ b/docs/relations.rst
@@ -83,7 +83,57 @@ hasMany Relation
 
 .. php:method:: hasMany($link, $model);
 
-There are several ways how to link models with hasMany
+There are several ways how to link models with hasMany::
+
+    $m->hasMany('Orders', new Model_Order());  // using object
+
+    $m->hasMany('Order', function($m, $r) {    // using callback
+        return new Model_Order();
+    });
+
+    $m->hasMany('Order');                      // will use factory new Model_Order
+
+
+Dealing with many-to-many relations
+-----------------------------------
+
+It is possible to perform relation through an 3rd party table::
+
+    $i = new Model_Invoice();
+    $p = new Model_Payment();
+
+    // table invoice_payment has 'invoice_id', 'payment_id' and 'amount_allocated'
+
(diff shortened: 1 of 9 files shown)


docs/conditions.rst
```diff
@@ -0,0 +1,208 @@
...
+nested and consist of objects as well as actions::
+
+
+    $m->addCondition(
+        $m->expr('[age] inbetween [min_age] and [max_age]'), 
+        [
+            'min_age'=>$m->action('min', ['age']), 
+            'max_age'=>$m->expr('(20 + [])', [20])
+        ]
+    );
```
skondakov: This one looks great - this way one can use a connection specific expression.

docs/relations.rst
```diff
@@ -0,0 +1,296 @@
...
+    $u = new Model_User($db)
+    $a = new Model_Address($db);
+
+    $u->hasOne('address_id', $a)
+        ->addFields([
+            'address_1',
+            'address_2',
+            'address_3',
+            ['notes', 'type'=>'text']
+        ]);
```
skondakov: Please provide a resulting query here

src/Field.php
```diff
@@ -21,6 +21,11 @@ function __construct($defaults = []) {
         }
     }
 
+    public function get()
+    {
+        return $this->owner[$this->short_name];
```
skondakov: Hm, why not just $this; ?

skondakov:
Seems good to me.

src/Field.php
romaninsh: because we need the value of the field, not the actual field itself. This is same as 4.3 btw.

docs/relations.rst
romaninsh: clarified in the further branch

PR #43, opened 2016-07-05
Cleanups to make 0.5 release more stable

romaninsh:
I've ran into problems when recording tutorial, so I'm fixing those minor bugs and will release them in a next minor release.

 - [x] implement insert()
 - [x] implement join::addFields(['foo','bar']);
 - [x] implement insertRows()
 - [x] implement reload()

### DIFF
--- a/src/Field_SQL_One.php
+++ b/src/Field_SQL_One.php
@@ -8,7 +8,7 @@ class Field_SQL_One extends Field_One
     function addField($field, $their_field)
     {
         $this->owner->addExpression($field, function($m) use ($their_field) {
-            return $m->refLink($this->link)->action('fieldValues', [$their_field]);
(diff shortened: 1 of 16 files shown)


src/Field_SQL_One.php
```diff
@@ -8,7 +8,7 @@ class Field_SQL_One extends Field_One
     function addField($field, $their_field)
     {
         $this->owner->addExpression($field, function($m) use ($their_field) {
-            return $m->refLink($this->link)->action('fieldValues', [$their_field]);
+            return $m->refLink($this->link)->action('field', [$their_field]);
```
skondakov: Definitely looks more clear this way

skondakov:
Looks good to me!

PR #45, opened 2016-07-06
Implement setLimit(), setOrder() and export().

romaninsh:
Review and merge #43 before merging this.

### DIFF
--- a/src/Persistence_SQL.php
+++ b/src/Persistence_SQL.php
@@ -127,6 +127,19 @@ class Persistence_SQL extends Persistence {
         }
     }
 
+    protected function setLimitOrder($m, $q)
+    {
+        if ($m->limit && $m->limit[0]) {
+            $dsql->limit($m->limit[0], $m->limit[1]);
+        }
+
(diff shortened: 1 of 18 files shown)


src/Persistence_SQL.php
```diff
@@ -127,6 +127,23 @@ public function initQueryFields($m, $q)
         }
     }
 
+    protected function setLimitOrder($m, $q)
+    {
+        if ($m->limit && ($m->limit[0] || $m->limit[1])) {
+            if ($m->limit[0] === null) {
+                // really, SQL?
+                $m->limit[0] = '18446744073709551615';
```
skondakov: Should this be hardcoded?

skondakov:
Looks good to me, except the comment above.

src/Persistence_SQL.php
romaninsh: it's a stupid SQL limitation, and that number shouldn't be there in the frist place. 

See http://stackoverflow.com/questions/255517/mysql-offset-infinite-rows

PR #46, opened 2016-07-06
Implement iterator

romaninsh:
- [x] merge #43 and #45 first.


skondakov:
Code looks good and UT coverage is very good as well. Its good to go!

PR #55, opened 2016-07-20
Fix aliases on related entities.

romaninsh:
When related entity relies on the same table it is possible to run into problem when SQL is confused about which table to use.

``` sql
select
   name, 
  (select name from item 
   where item.parent_id = item.id
  ) parent_name
from item
```

To avoid this problem this PR introduces use of aliases in related entities. The related table alias does not have to be passed and it will be automatically deducted from field:

With this patch recursive models produce queries like this:

``` sql
select
    `pp`.`id`,`pp`.`name`,`pp`.`age`,`pp_i`.`parent_item_id`,
    (select `parent`.`name` 
     from `item` `parent`
     left join `item2` as `parent_i` on `parent_i`.`item_id` = `parent`.`id` 
     where `parent`.`id` = `pp_i`.`parent_item_id`
     ) `parent_item`,
    (select sum(`child`.`age`) from `item` `child`
     left join `item2` as `child_i` on `child_i`.`item_id` = `child`.`id` 
     where `child_i`.`parent_item_id` = `pp`.`id`
    ) `child_age`,`pp`.`id` `_i` 
from `item` `pp`left join `item2` as `pp_i` on `pp_i`.`item_id` = `pp`.`id`
```


skondakov:
Seems good to me! I've also tested the examples in the section.

PR #61, opened 2016-07-21
Implement hasOne() relationship ID binding.

romaninsh:
When populating model that contains hasOne() relation, sometimes you want to also populate the related entity. This should nicely place the ID of a new record into the main table:

```
$m = new Model_User($db);

$m['name'] = 'John';
$m->ref('address_id')->save(['address'=>'street']);
$m->save();
```

This should also save $m['address_id'];

### DIFF
--- a/src/Field_One.php
+++ b/src/Field_One.php
@@ -132,9 +132,15 @@ class Field_One
         $m = $this->getModel($defaults);
         if ($this->owner->loaded()) {
             if ($this->their_field) {
-                return $m->loadBy($this->their_field, $this->owner[$this->our_field]);
+                return $m->tryLoadBy($this->their_field, $this->owner[$this->our_field])
+                    ->addHook('afterSave', function($m){ $this->owner[$this->our_field] = $m[$this->their_field]; })
+                    ->addHook('afterDelete', function($m){ $this->owner[$this->our_field] = null; })
+                    ;
             } else {
-                return $m->load($this->owner[$this->our_field]);
+                return $m->tryLoad($this->owner[$this->our_field])
+                    ->addHook('afterSave', function($m){ $this->owner[$this->our_field] = $m->id; })
+                    ->addHook('afterDelete', function($m){ $this->owner[$this->our_field] = null; })
+                    ;
             }
         } else {
             $m = clone $m; // we will be adding conditions!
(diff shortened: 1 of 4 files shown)


src/Field_One.php
```diff
@@ -132,9 +132,21 @@ public function ref($defaults = [])
         $m = $this->getModel($defaults);
         if ($this->owner->loaded()) {
             if ($this->their_field) {
-                return $m->loadBy($this->their_field, $this->owner[$this->our_field]);
+                return $m->tryLoadBy($this->their_field, $this->owner[$this->our_field])
+                    ->addHook('afterSave', function ($m) {
```
DarkSide666: Do we need afterSave hook here even if referenced model can be loaded? This is good if we allow to change `id` of model, but ... do we allow that?

romaninsh: That is a very good point. I'm not sure which one is better.

tests/RelationSQLTest.php
```diff
@@ -253,4 +253,51 @@ public function testAggregateHasMany()
...
+        $this->assertEquals('Joe contact', $u['address']);
+        $this->assertEquals('Joe contact', $u->ref('contact_id')['address']);
+
+        $u->load(2);
+        $u->ref('contact_id')->save(['address' => 'Peters new contact']);
+
+        $this->assertNotEquals(null, $u['contact_id']);
+        $this->assertEquals('Peters new contact', $u->ref('contact_id')['address']);
+
+        $u->save()->reload();
```
DarkSide666: reload() is needed to reload hasOne->addField fields?

src/Field_One.php
DarkSide666: I don't think agile/data supports changing model id. Can we do:
```
$m->load(1);
$m['id'] = 2;
$m->save(); // update table set id=2 where id=1 ??? do this work or will create insert instead?
```

DarkSide666: I don't think agile/data supports changing model id. Can we do:
```
$m->load(1);
$m['id'] = 2;
$m->save(); // update table set id=2 where id=1 ??? do this work or will create insert instead?
```

romaninsh: OK, i think i want to keep as-is for the following pattern:

$user->ref('contact_id')->load(123)->save();

which allows to easily update to a different record.

tests/RelationSQLTest.php
romaninsh: currently those methods will always guarantee to create a new model when called, so no need to worry about them just yet.

PR #63, opened 2016-07-21
Intelligent model reloading after save

romaninsh:
When saving the model in Agile Data some field values may not be "true" after save, such as SQL expressions or fields that have default values in the database. 

This PR implements intelligent reloading of the models after they are saved. Includes docs.

### DIFF
--- a/docs/expressions.rst
+++ b/docs/expressions.rst
@@ -105,7 +105,7 @@ You can use a callback method when defining expression::
 Model Reloading after Save
 --------------------------
 
-Whe you add SQL Expressions into your model, that means that some of the fields
+When you add SQL Expressions into your model, that means that some of the fields
 might be out of sync and you might need your SQL to recalculate those expressions.
 
 To simplify your life, Agile Data implements smart model reloading. Consider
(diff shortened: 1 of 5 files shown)


docs/expressions.rst
```diff
@@ -101,3 +101,92 @@ You can use a callback method when defining expression::
...
+expressions. This time you can explicitly enable reload after save::
+
+    $m = new Model_Math($db, ['reload_after_save' => true]);
+    $m['a'] = 4;
+
+    $m->save();
+
+    echo $m['a']+$m['b']; // outputs 14
+
+.. note:: If your model is using reload_after_save, but you wish to insert
```
DarkSide666: What if I want to update without reload and model is using reload_after_save ? Is that possible? Like ->saveAndUnload() in atk.

src/Model.php
```diff
@@ -733,7 +745,9 @@ public function save($data = [])
             $this->id = $this->persistence->insert($this, $data);
             $this->hook('afterInsert', [$this->id]);
 
-            //$this->hook('beforeInsert', array(&$source));
+            if ($this->reload_after_save !== false) {
```
DarkSide666: shouldn't here be `===true` instead of `!==false` ?

docs/expressions.rst
romaninsh: Answered here: https://github.com/atk4/data/pull/65/files#diff-954e56aad15105e727c7776fcd9ee369R183

src/Model.php
romaninsh: `!==` false is correct. will reload if it's either null or true.

PR #69, opened 2016-07-21
[1.1] Allow updating ID of a model and properly reference fields in update.

romaninsh:
When performing an update in MySQL, we can have multiple conditions set both on the main table and joined table. 

Currently there is something weird with the way how it's done, but UPDATE statement includes join() even though joined tables are updated separately by Join_SQL.

Because of that fields must be prefixed. This works in MySQL but sqlite does not like joins during update.  And since we don't need joins, we should get rid of them and rebuild action('update') without joins and take out table prefixes from the field expressions too.

This PR is work in progress to clean all this up and finally make it possible to perform this:

```
$m->load(10);
$m['id'] = 12;
$m->save();

echo $m->id; // 12
```


DarkSide666:
Looks OK,but need to add some test cases ;)

DarkSide666:
Also should not set editable=false in Model->init() method as it is now!

PR #70, opened 2016-07-22
Add descriptions and few fixes

DarkSide666:
Mostly descriptions for methods and properties are added.
But there are a few fixes too - not much, but still.

### DIFF
--- a/src/ConnectionInterface.php
+++ b/src/ConnectionInterface.php
@@ -6,6 +6,8 @@ interface ConnectionInterface
 {
     /**
      * Create new instance of the model.
+     *
+     * @param string $model_name Name of the model
      */
     public function add($model_name);
 }
--- a/src/ConnectionInterface.php
+++ b/src/ConnectionInterface.php
@@ -1,7 +1,12 @@
 <?php
 
+// vim:ts=4:sw=4:et:fdm=marker:fdl=0
+
 namespace atk4\data;
 
+/**
+ * Class description?
+ */
 interface ConnectionInterface
 {
     /**
--- a/src/Model.php
+++ b/src/Model.php
@@ -489,12 +489,13 @@ class Model implements \ArrayAccess, \IteratorAggregate
     // }}}
 
     // {{{ ArrayAccess support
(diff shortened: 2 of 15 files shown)


src/ConnectionInterface.php
```diff
@@ -1,11 +1,18 @@
 <?php
 
+// vim:ts=4:sw=4:et:fdm=marker:fdl=0
+
 namespace atk4\data;
 
+/**
+ * Class description?
```
romaninsh: i wonder if we even use this.

src/Model.php
```diff
@@ -826,18 +1057,30 @@ public function getIterator()
...
+     * @return $this
+     */
     public function each($method)
     {
         foreach ($this as $rec) {
             if (is_string($method)) {
                 $rec->$method();
-            } else {
-                $method($rec);
+            } elseif (is_callable($method)) {
```
romaninsh: warning, this will fail when 'system' variable is passed, because it will think it is executable. Search for 'Callable' for a proper implementation.

src/ConnectionInterface.php
DarkSide666: No, currently we don't use this interface anywhere. Can I delete it completely in this PR?

src/Model.php
```diff
@@ -912,12 +1173,18 @@ public function join($foreign_table, $defaults = [])
...
+     * @param array  $defaults
+     *
+     * @return Join
+     */
     public function leftJoin($foreign_table, $defaults = [])
     {
-        if (!is_array($defaults)) {
-            $defaults = ['master_field' => $defaults];
-        }
-
```
romaninsh: this is causing a problem.

romaninsh:
```
leftJoin('table', 'field'); 
```

DarkSide666: Yes, that's my mistake.
It looked that `join()` have same checking so I removed this "duplicated" one, but I didn't notice, that `$defaults['weak'] = true;` line which changes situation :(

PR #71, opened 2016-07-25
Improve joined model update/delete.

romaninsh:


romaninsh:
Current problem. $m->delete() and $m->save() rely on action('update') and
action('delete'). Those are not really compatibel anymore.

Single-record save() and delete() are properly augmented by Join_SQL to ensure
that related entities are deleted. More importantly, it might attempt to reload record
after to make sure it is within the same set. 

$m->save() invokes update with a fixed ID. Iit updates joined
tables separately. Because the record has been loaded successfully
before and will be reloaded after updates (in some cases) at a risk of
failed transaction, we don't have to invoke action here.

Action is more suited for all-set update and if you use joins with
SQLite you're on your own, as it does not support join with update.

Same goes for deletion.

With this PR Persistence_SQL will no longer use action for update/delete but will instead create a basic query. This solves prolbems:

 - ID field update
 - updating records in joined tables
 - updating records that are duplicated across joined tables
 - deleting records with condition on joined table field

(test-scripts for the above cases are not supplied)

PR #72, opened 2016-07-25
Preserves "Dirty" flags for afterSave hooks.

romaninsh:
when using afterSave hooks you might be interested to look at the dirty fields. Unfortunately using the "reload after save" mechanics will cause it to reload the record which resets dirty fields.

This fix:
 - will memorize dirty fields before reloading (A)
 - will memorize dirty fields after reloading (B)
 - will restore dirty fields A then call hooks.
 - will merge with dirty fields B after hooks.

Goal of this PR:
 - make sure you get all dirty fields inside afterSave hook regardless if "reload after save" is On/Off.
 - if you perform any changes, in afterSave - do not forget those.
 - if afterLoad performs any changes as a part of reloading, keep those in our dirty array

No test-scripts provided for the above cases.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -886,6 +886,8 @@ class Model implements \ArrayAccess, \IteratorAggregate
      *
      * @return $this
      */
+    public $_dirty_after_reload = [];
+
     public function save($data = [])
     {
         if (!$this->persistence) {
@@ -899,6 +901,8 @@ class Model implements \ArrayAccess, \IteratorAggregate
         if ($this->hook('beforeSave') === false) {
             return $this;
         }
+
+        $this->_dirty_after_save = [];
 
         $is_update = $this->loaded();
         if ($is_update) {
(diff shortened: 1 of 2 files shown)


src/Model.php
```diff
@@ -886,6 +886,8 @@ public function tryLoadBy($field, $value)
      *
      * @return $this
      */
+    public $_dirty_after_reload = [];
```
DarkSide666: should be defined at top of Model class file.

src/Model.php
```diff
@@ -900,6 +902,8 @@ public function save($data = [])
             return $this;
         }
 
+        $this->_dirty_after_save = [];
```
DarkSide666: not defined in class file - at top.

src/Model.php
romaninsh: i thought since it's $_xx and we don't want anyone to use it but can't make it "private" due to our other class referring to it and also because our style checkers didn't complain, then we can leave here.

src/Model.php
romaninsh: yeah it was initially caled dirty_after_save and i renamed it into dirty_after reload, let me check if this is necessary.

romaninsh: ok, it's used elsewhere in the file, so we need this.

DarkSide666: ok, will recheck after #73

PR #80, opened 2016-07-30
Persistence requires name trait

DarkSide666:
Persistence requires name trait otherwise we can get exception from Container trait _addContainer method which uses $name property in this line:
```
        $element->name = $this->_shorten($this->name.'_'.$element->short_name);
```

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -24,35 +24,35 @@ class Model implements \ArrayAccess, \IteratorAggregate
      *
      * @var string
      */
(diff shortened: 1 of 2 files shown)


romaninsh:
I thought persistence was a container and container does not need name trait..

romaninsh:
Ok, as per http://agile-core.readthedocs.io/en/develop/container.html?highlight=nametrait#ObjectTrait - NameTrait is needed on containers.

PR #81, opened 2016-08-01
Field new method set() which set field value in model

DarkSide666:


romaninsh:
- [ ] please update documentation.

DarkSide666:
not sure where to put this in docs.

PR #85, opened 2016-08-02
nothing serious :)

DarkSide666:


DarkSide666:
Will create new PR soon which will include these changes

PR #86, opened 2016-08-03
Feature/rename classes etc.

DarkSide666:
* rename classes Field_One, Field_Many and Field_SQL_One to Relation_x
* related changes in docs

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -143,7 +143,11 @@ class Field
             $arr['join'] = $this->join;
         }
 
-        return $arr;
+        if ($this->editable) {
+            $arr['editable'] = $this->editable;
+        }
+
+         return $arr;
     }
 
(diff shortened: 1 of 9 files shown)


src/Field.php
```diff
@@ -143,6 +143,10 @@ public function __debugInfo()
             $arr['join'] = $this->join;
         }
 
+        if ($this->editable) {
+            $arr['editable'] = $this->editable;
+        }
+
```
romaninsh: I don't think we need this. I try to keep only the essential information there. Join is quite impactful when you're debugging, but editable does not amount for much.

I suppose it's ok, for now but possibly will want to clean this up later.

DarkSide666: Ok, but currently while developing it's very useful

PR #87, opened 2016-08-03
Implement Field->mandatory property

DarkSide666:


romaninsh:
I do t think that mandatory should be in AD. It can. Be defined in ATK inside custom field class. Will accept but need to move this property out properly.

romaninsh:
whops forgot to actually accept ;)

PR #88, opened 2016-08-11
Fix deep traversal

DarkSide666:
**Fixes deep traversal for hasOne.**

If owner model is loaded, then it simply loads referenced model (like before).
But if owner model is not loaded and we use Persistence_SQL, then we add condition with subselect to referenced model.

Previously:
```
$order->unload()->addCondition('amount', '>', 5);
$order->ref('user_id')->action('select')->render();
// select `id`,`name` from `user` - WITHOUT conditions
```

Now:
```
$order->unload()->addCondition('amount', '>', 5);
$order->ref('user_id')->action('select')->render();
// select `id`,`name` from `user` where `id` in (select `user_id` from `order` where `amount` > :a)
```

**Also**
* fixes Field_One_SQL -> Field_SQL_One class/file name
* multiple fixes in comments

### DIFF
--- a/src/Field_Many.php
+++ b/src/Field_Many.php
@@ -5,6 +5,8 @@
 namespace atk4\data;
 
 /**
- * Obsolete. Left for compatibility
+ * Obsolete. Left for compatibility.
  */
-class Field_Many extends Relation_Many {}
+class Field_Many extends Relation_Many
+{
+}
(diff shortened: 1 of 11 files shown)


romaninsh:
Can you make please add a few tests for inserting into hasOne() with reloading? (you need to define expression)

romaninsh:
Added ticket for this #89

Issue #82, opened 2016-08-02
Reverse Join fails to update record


romaninsh:
It appears that the use of $join->id in reveres join is incorrect.

Look at this file: https://github.com/atk4/data/blob/develop/src/Join_SQL.php#L211

With reverse join you are joining like this:

    $db->join('contact.user_id');

`foreign_field` is user_id, and suppose user_id is 2. When you add new record, $join->id is [set to lastInsertID](https://github.com/atk4/data/blob/develop/src/Join_SQL.php#L184):

    $this->id = $insert->connection->lastInsertID();

so it holds the value of contact.id, let's say it's 1 for our new record. If you execute delete after, the where() clause can cause deletion of incorrect record:

    $delete->where($this->foreign_field, $this->id);

which would render to "where user_id = 1".

 - [x] create failing test-case for this bug
 - [x] implement a fix

romaninsh:
This should be fixed by #107

Issue #98, opened 2016-08-13
Nested join doesn't use correct field in condition


DarkSide666:
`$user->join('contact_info')->join('address')`

This tries to use `user.address_id` field in last join condition instead of `contact_info.address_id`.

PR #99, opened 2016-08-13
Feature/fix join kind

### LINKED ISSUE #97: Join_SQL->kind doesn't pass to DSQL and it always use left join

DarkSide666:
fixes #97 and #98 

* Join_SQL forgot to set join kind for DSQL query at all.
* Join_SQL forgot that in case of nested joins it should look up table alias not in owner model, but in "previous" join foreign_table/foreign_alias.

### DIFF
--- a/src/Join_SQL.php
+++ b/src/Join_SQL.php
@@ -9,10 +9,10 @@ namespace atk4\data;
  */
 class Join_SQL extends Join implements \atk4\dsql\Expressionable
 {
-    public $foreign_alias;
     /**
      * A short symbolic name that will be used as an alias for the joined table.
      */
+    public $foreign_alias;
 
     /**
      * By default this will be either "inner" (for strong) or "left" for weak joins.
--- a/src/Join_SQL.php
+++ b/src/Join_SQL.php
@@ -106,9 +106,7 @@ class Join_SQL extends Join implements \atk4\dsql\Expressionable
         if ($this->on) {
             $query->join(
                 $this->foreign_table.' '.$this->foreign_alias,
-                $this->on instanceof \atk4\dsql\Expression ?
-                $this->on :
-                $query->expr($this->on),
+                $this->on instanceof \atk4\dsql\Expression ? $this->on : $query->expr($this->on),
                 $this->kind
             );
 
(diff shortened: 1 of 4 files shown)


src/Join_SQL.php
```diff
@@ -119,9 +118,14 @@ public function initSelectQuery($model, $query)
...
+                    // if join is nested, then use previous join table alias
+                    ? (isset($this->join->foreign_alias) ? $this->join->foreign_alias : $this->join->foreign_table)
+                    // otherwise use owner model table alias
+                    : (isset($this->owner->table_alias) ? $this->owner->table_alias : $this->owner->table)
+                ).'.'.$this->master_field
+            ),
+            $this->kind
         );
 
         if ($this->reverse) {
```
DarkSide666: Line next to this still looks suspicious. short_name = $this->join? really?
I have a feeling that something like I fixed few lines above should be there - with $this->join->foreign_alias/table if $this->join is set.
Please look into this before merging!

romaninsh:
Does this resolve #82 ? Can we work a bit more to clean up multi-level joins here?

src/Join_SQL.php
romaninsh: there might be  a bit confusion. join is normally same as short_name. It was short_name originally. I think i started using ->join at some point or other way around but didn't complete. I think i will merge all of your PRs into single one, finish with joins add a bunch of test-scripts and then merge.

romaninsh:
Noticed the lack of tests... ;)

DarkSide666:
I'm not sure about #82 - have not tested that.

tests/JoinSQLTest.php
```diff
@@ -378,4 +378,113 @@ public function testDoubleSaveHook()
...
+                2 => ['id'=>2, 'name'=>'US'],
+                3 => ['id'=>3, 'name'=>'India'],
+        ], ];
+        $this->setDB($a);
+
+        $db = new Persistence_SQL($this->db->connection);
+        $m_u = new Model($db, 'user');
+        $m_u->addField('contact_id');
+        $m_u->addField('name');
+        $j = $m_u->join('contact');
```
DarkSide666: Is this really reverse join?

PR #107, opened 2016-08-19
Big cleanup in the join logic and addition of big tests

romaninsh:
This PR introduces few necessary tools to run more sophisticated tests and then fixes a few issues that were spotted when using Agile Data in commercial projects.

## SMBO TestSuite

SMBO is an accounting app. A few models have been mocked up with the similar logic and presented in tests/smbo/lib. Test scripts use those models to test a deeper logic.

## Added TransferTest

In SMBO there is an entity called Document. Any document with doc_type='payment' is considered Payment. Some of those payments can be Transfers.

Transfers exist in pairs - one transfer decrease amount on source account, the other transfer increases the amount on the destination account.

Creating a transfer will create the other leg too and cross-link those documents. 

## Implementing Cloning support.

When transfer document is created and we are willing to create a destination record, it is actually mirror-matched to our own record. Although we could copy all the properties, some users may want to do a "clone" here.

I have improved support for model cloning. The new clone must have no associations with the source model. Previously something wasn't linking up correctly, so when creating a transfer, some ID values were shared incorrectly.

## Join refactoring

Previously joins were linked between 2 physical fields. When creating a join they were updating "insert" query of the parent to add the field in there.

Now joins are refactored in a way where they are joined between a Model field and a physical field of joined table. So in theory you should be able to:

    $this->getRef('user_id')->addField('user_county_id', 'country_id');
    $this->join('country', 'user_country_id');

Instead of beforeInsertQuery it now uses beforeInsert and same thing with modification.


src/Model.php
```diff
@@ -237,6 +237,19 @@ public function __construct($persistence = null, $defaults = [])
...
     }
 
+    public function __clone()
+    {
+        // we need to clone some of the elements
+        if ($this->elements) {
+            foreach ($this->elements as $id => $el) {
+                if ($el instanceof Join) {
+                    $this->elements[$id] = clone $el;
+                    $el->owner = $this;
```
DarkSide666: I guess there is bug. $el variable will be lost after foreach anyway. Probably $this->elements[$id]->owner should be there instead of $el->owner.

romaninsh: you are correct!

romaninsh: strangely - changing this logic breaks the build.

README.md
```diff
@@ -42,6 +43,43 @@ object off your model which you can tweak.
...
+Most ORM today are built by extending Query Builder. While this gives a benefit to build "Queries"
+out of your entity this in a one-way operation. Additionally things you CAN build into a query
+are quite limited.
+
+
+
+
+### Introducing Actions
+
+ORM/Active Record is not designed for relational databases. There are too much potential that ORM
```
DarkSide666: should be "... is not only designed ..." instead of "... is not designed ..." otherwise reader will think that AD will not work for SQL bases at all :)

romaninsh: ok, valid point.

README.md
```diff
@@ -42,6 +43,43 @@ object off your model which you can tweak.
...
+abstracts. The way how modern ORM get around it by offering ability to generate "Query-in-progress"
+object off your model which you can tweak.
+>>>>>>> develop
```
romaninsh: crap, what's this.

src/Model.php
DarkSide666: it shouldn't be like that

Issue #90, opened 2016-08-13
Create standard type set


romaninsh:
Types such as 'string', 'date', 'int' and 'boolean' (or bool) should be standart and all persistences should know how to deal with those types. 

 - [x] define list of types on a wiki page
 - [x] define normalisation rules and write documentation
 - [x] create test-script
 - [x] implement for SQL persistence
 - [ ] imlpement for Array persistence

https://github.com/atk4/data/wiki/Type-Normalization

DarkSide666:
string = text, varchar,varchar2, char etc.
date = date, datetime
int = any numeric value, int, smallint, longint, float, numeric, ?timestamp?
boolean = bool, tinyint(1), int, enum etc.

romaninsh:
This list can be useful too: http://php.net/manual/en/mongo.types.php

DarkSide666:
so there can also be these types:
* array - array, list (mostly for NoSQL persistence)
* binary - BinaryData for NoSQL, BLOB for SQL

romaninsh:
After thinking - the type conversion will be handled by Persistence SQL just when the data comes off the iterator before it's passed into $model->data

 - [ ] When query is being built also build a "transformation" consisting of [field=>fx, ..] array.
 - [ ] When results are fetched, pass them through transformator
 - [ ] Allow fields to define custom transformators

romaninsh:
if we use "int" we should use "bool" btw.

romaninsh:
definition of having a type - if field has a type then persistence will normalize it.

I suppose if we add 'array' into allowed typess, then MySQL will have to serialize it.

romaninsh:
Fully fixed in #130

PR #109, opened 2016-08-22
Feature/type casting

romaninsh:
This PR implements a proper type-casting when storing / loading data from SQL databases. All the types are implemented as requested in #90 

Additionally a few additional fixes / features were include:

## Strict Fields

By default Agile Data will throw exception if you try to `set()` or `get()` value for a non-existant field. It's possible to switch off Strict fields

 - [ ] Add documentation for strict fields

## Fixed bug in isDirty()

Previously if default field Value was null, then isDirty would return false when it should return true. Fixed.

## Added check for read-only fields

Now you can no longer change values of read-only fields

## Removed hook('normalize')

This was half-arsed implementation of typecasting. It's now obsolete and must be removed from everywhere (please check in case i missed)

 - [x] Check if hook('normalize') is removed from source, docs and tests.

## Type-casting for expressions

If you set field value to expression or null, then type-casting will not touch it. 

 - [ ] Implement "mandatory" to verify for non-null values.

## addTitle() fields now support sub-query lookups.

Isn't it nice to be able to define:

   $m->hasOne('currency_id', new Currency())->addTitle();

and then be able to just:

   $m->insert('title' ['currency'=>'EUR']);

Now you can do this, no need to even add anything!


src/Field.php
```diff
@@ -49,13 +57,22 @@ class Field
...
+
+    /**
      * Setting this to true will never actually store
      * the field in the database. It will action as normal,
      * but will be skipped by update/insert.
      *
      * @var bool
      */
-    public $never_persist = false;
+    public $never_save = false;
```
DarkSide666: What's the difference between $never_save and $readonly. They both look the same for me.

DarkSide666:
Reviewed. Questions below:

* naming: we have `typecastSaveToPersistence` and `typecastLoadToPHP`.
I propose renaming `typecastLoadToPHP` to `typecastLoadFromPersistence`.

* I'm not sure about `FieldTest->testPersist` is it correct or not. Name field is set as never_persist, but at the end of the test you still can change this field value easily with beforeSave hook. Is that OK?

* Not sure what you are testing in `TypecastingTest->testCastingExpressions()`. How this is related with typecasting? It looks similar to `FieldTest->testTitle()`.

* What is purpose of `_Template.php` class? It's not used anywhere.

src/Field.php
romaninsh: readonly (renamed to read_only in next PR) will not let you even change the value. Calling set('ro_field', foo) will throw exception.

romaninsh:
Answers to your comments above:

PR #112, opened 2016-08-22
[1.1] Improve multi-persistence, multi-record and multi-object operations

romaninsh:
One of the core features for 1.1 is improved support for multiple permissions as well as improved handling when working with multiple records. This PR adds a lot of functionality on top of existing logic. When reviewing make sure each section has proper documentation and tests.

### newInstance()

Creates a Vanilla copy of your model.

### asModel()

Casts your model into a different class without  reloading.

### saveAs()

Saves your model under a different class.

### saveAndUnload()

Saves your model then unload. Will not try to reload.

### withPersistence()

Returns new instance of your model associated with given persistence (and copy dirty values).

A lot of documentation have been updated to provide various examples how the new features can be used for:

 - creating MemCache for load-by-id caching strategy with SQL write-back.
 - create read/write replica support
 - store active record into different DB every time it is updated
 - implement caching of logged-in user inside session

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -3,7 +3,7 @@
 Advanced Topics
 ===============
 
-Agile Data allow you to implement various tricks. 
+Agile Data allow you to implement various tricks.
 
 
 Audit Fields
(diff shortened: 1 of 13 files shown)


DarkSide666:
PR reviewed, can be merged in.

There are no tests for these features, but I have created issue #119 for this.

PR #130, opened 2016-08-29
Various bugfixes in preparation for 1.1

romaninsh:
Wouldn't that be great if there was a simple method that would convert the $data from the business realm in the way how Model prepares them for update() / insert() operations into an actual array that can be passed into DSQL ?

Now that's possible with. typecastSaveToPersistence does exactly that. Not only it converts the types for all the fields, but it also makes use of 'actual' property in SQL fields.

## Documentation is important

I have added a full documentation for Field and Field_SQL classes.

## Tests are added

Trying to change `read_only` field causes Exception. Trying to save with Mandatory field equal to NULL also causes Exception. Trying to set field with `enum` property to an incorrect value similarly generates an Exception.

## Other Fixes

Fixed `date` and `time` fields that will no longer adjust themselves for TIMEZONE changes. The `datetime` is still stored in UTC.

## Joined fields

Sending joined fields into their respective table also typecasts them.


src/Persistence_SQL.php
```diff
@@ -325,7 +325,7 @@ public function initQueryConditions($m, $q)
      *
      * @return array
      */
-    public function typecastLoadToPHP($m, $row)
+    public function typecastLoadFromPersistence($m, $row)
```
DarkSide666: so, you like my variant of method name better? :)

src/Persistence_SQL.php
```diff
@@ -398,78 +414,110 @@ public function typecastLoadToPHP($m, $row)
...
-                    $format = ['date' => 'Y-m-d', 'datetime' => 'Y-m-d H:i:s', 'time' => 'H:i:s'];
+                break;
+            case 'date':
+            case 'datetime':
+            case 'time':
+                $class = isset($f->dateTimeClass) ? $f->dateTimeClass : 'DateTime';
+
+                $format = ['date' => 'Y-m-d', 'datetime' => 'Y-m-d H:i:s', 'time' => 'H:i:s'];
+
+                if (is_numeric($value)) {
```
DarkSide666: yeay >60 is gone, thanks! is_numeric should take care of everything

PR #139, opened 2016-08-31
Backward compatibility of Reference classes, changes in addTitle, syntax in docs

### LINKED ISSUE #76: Update docs SQL syntax
Correct syntax highlighter should be set in SQL blocks in docs.
http://build-me-the-docs-please.readthedocs.io/en/latest/Using_Sphinx/ShowingCodeExamplesInSphinx.html#showing-source-code-examples-in-sphinx

### LINKED ISSUE #137: Reference_One->addTitle() changes
* addTitle() should return new Expression field not Reference_One instance;
* addTitle() should have ability to pass $defaults to expression field;
* ~~addTitle() shouldn't set read_only=false, expressions should always be read_only=true. And no need for never_save=true here if we already set read_only=true;~~
* addTitle() should by default advice UI to not show title field in forms as editable field. This can be overwritten by passing $defaults in addTitle() method.

DarkSide666:
* added Relation_* classes for backward compatibility
* changes in addTitle() - fixes #137 
* syntax of SQL blocks in docs - fixes #76

### DIFF
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -51,7 +51,7 @@ Operations
 
 Most database drivers will support the following additional operations::
 
-    >, <, >=, <=, !=, in
+    >, <, >=, <=, !=, in, not in
 
 The operation must be specified as second argument::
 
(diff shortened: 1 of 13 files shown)


romaninsh:
Please implement withField() as discussed and then this can be merged!

Issue #131, opened 2016-08-29
better implementation for `fmt` and currency for money.


romaninsh:
Currently, is not fully implemented.

romaninsh:
With  #144 implementing fmt/currency is much simpler. Documentation even refers to Advanced section where 'Field_Currency' could be implemented. 

But the destination article is missing. Also custom formats has mention in the documentation, but does not seem to reflect in the code and does not have any test-scripts.

DarkSide666:
related discussion: https://github.com/atk4/data/issues/147

romaninsh:
After much thought I decided to drop the 'fmt' entirely and leave money implementation as-is. Let's focus on releasing 1.1 now.

PR #144, opened 2016-09-06
Implement Strict Types

romaninsh:
Previously when you execute `$m['date'] = time()` the value would be stored as integer until the save. Save would then convert it and subsequential load would now store field as DateTime.

This pattern creates uncertainty. Will your "date" field be int or DateTime? how can you rely on that?

This PR introduces strict_types. Now when you set `$m['date'] = time()` it will be converted into DateTime right away. No more ambiguity. 

It also makes it simpler for persisting data, because we no longer need to look into various cases.

A full documentation included!

TODO (before this can be accepted):

 - [ ] no tests or implementation for "load normalization"
 - [ ] See #131 for some problems with Currency and Date formats.


docs/persistence.rst
```diff
@@ -106,6 +106,365 @@ database record will also have its ID changed. Here is example::
...
+
+    $m->addField('is_married', [
+        'type' => boolean',
+        'enum' => ['Yes', 'No']
+    ]);
+
+    $m['is_married'] = 'Yes';  // type-casts into true
+    $m['is_married'] = true;   // better way!
+
+    $m->save();   // stores as "Yes".
```
DarkSide666: stores as "Yes" or 1 in SQL persistence ?

romaninsh: Yes in persistence.

src/Model.php
```diff
@@ -456,7 +470,18 @@ public function set($field, $value = null)
 
         $f = $this->hasElement($field);
 
-        // set hook here
+        try {
+            if ($this->hook('normalize') !== false && $f) {
```
DarkSide666: I guess we have to pass $field and $value to normalize hook here.

src/Model.php
```diff
@@ -456,7 +470,18 @@ public function set($field, $value = null)
...
+                $value = $f->normalize($value);
+            }
+        } catch (Exception $e) {
+            if (method_exists($e, 'addMoreInfo')) {
```
DarkSide666: Let's add correct atk4/core version dependency in composer.json and then we can get rid of this condition.

DarkSide666:
review done, bunch of fixes commited, but there are still few questions.

src/Model.php
romaninsh: yes, that's correct.

src/Model.php
romaninsh: OK. I'll do that .

docs/persistence.rst
romaninsh: that's correct. But now we need to swap Yes/No in example.

src/Model.php
DarkSide666: changed to `if ($f && $this->hook('normalize', [$f, $value]) !== false)`

docs/persistence.rst
DarkSide666: Will switch Yes/No in docs, source and tests because [0=>'Yes', 1=>'No'] is misleading.
It should better be 0=false=No and 1=true=Yes.

PR #151, opened 2016-09-09
Few fixes

DarkSide666:
* set field type to 'string' for addTitle generated field
* boolean fields should normalize integers to boolean otherwise we always need to set enum [0,1] even if we just want to use/store 0|1. This is also needed to be compatible with atk4 toolkit form fields
* fixing empty string and null value issue once and for all (I hope) !!!
* Reference_SQL_One->addField now allows to pass additional properties to field object

### DIFF
--- a/src/Persistence_SQL.php
+++ b/src/Persistence_SQL.php
@@ -429,6 +429,16 @@ class Persistence_SQL extends Persistence
             return $callback($value, $f, $this);
         }
 
+        if ($value === null) {
+            return null;
+        }
+
+        // only string type fields can use empty string as legit value, for all
+        // other field types empty value is the same as no-value, nothing or null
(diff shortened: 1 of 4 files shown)


src/Persistence_SQL.php
```diff
@@ -429,6 +429,16 @@ public function typecastLoadField(Field $f, $value)
...
         }
 
+        if ($value === null) {
+            return;
+        }
+
+        // only string type fields can use empty string as legit value, for all
+        // other field types empty value is the same as no-value, nothing or null
+        if ($f->type != 'string' && $value === '') {
+            return;
```
romaninsh: i would use "return null" here just to be clear that it returns null.

DarkSide666: yeah I did like that at first, but StyleCI said that it's not good :(

PR #157, opened 2016-09-16
Import types from hasOne()->addField()

romaninsh:
Previously the type would be lost, but now we can actually look inside the related model and fetch the type form there.

This is useful as you don't have to worry about messing up.


src/Reference.php
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /**
+     * If you have set $their_field property correctly, then calling this method
+     * will traverse into a related model and fetch the type of their field. This
+     * method will be extra careful to avoid any loops.
+     */
+    public function guessFieldType($their_field = null)
+    {
+        if (!is_scalar($their_field)) {
```
DarkSide666: probably should use `is_string` instead of `is_scalar` here

src/Reference.php
```diff
@@ -0,0 +1,237 @@
...
+    public function ref($defaults = [])
+    {
+        return $this->getModel($defaults);
+    }
+
+    /**
+     * If you have set $their_field property correctly, then calling this method
+     * will traverse into a related model and fetch the type of their field. This
+     * method will be extra careful to avoid any loops.
+     */
```
DarkSide666: should add @param and @return in description

src/Reference.php
```diff
@@ -0,0 +1,237 @@
...
+
+        // see where we shouldn't traverse
+        if (isset($this->owner->_do_not_traverse_into)) {
+            $dnti = $this->owner->_do_not_traverse_into;
+
+            if (isset($dnti[$cl])) {
+                return;
+            }
+        }
+        $dnti[$cl] = true;
```
DarkSide666: $dnti is undefined here if $this->owner->_do_not_traverse_into is not defined at all in model

src/Reference.php
```diff
@@ -0,0 +1,237 @@
...
+
+        return $f ? $f->type : null;
+    }
+
+    // {{{ Debug Methods
+
+    /**
+     * List of properties to show in var_dump.
+     */
+    protected $debug_fields = ['our_field', 'their_field'];
```
DarkSide666: Should be $__debug_fields instead of $debug_fields

src/Reference.php
```diff
@@ -0,0 +1,237 @@
...
+     * Returns array with useful debug info for var_dump.
+     *
+     * @return array
+     */
+    public function __debugInfo()
+    {
+        $arr = [
+            'ref'     => $this->link,
+            'model'   => $this->model,
+        ];
```
DarkSide666: probably $link and $model can be defined in $__debug_fields too and removed from here.
$__debug_fields can even use array keys if you want to do like this:
```
protected $__debug_fields = ['ref'=>'link', 'model', 'our_field', 'their_field'];
public function __debugInfo()
{
    $arr = [];
    foreach ($this->__debug_fields as $k => $v) {
        $k = is_numeric($k) ? $v : $k;
        if (isset($this->$v)) {
            $arr[$k] = $this->$v;
        }
    }
    return $arr;
}
```

src/Reference_One.php
```diff
@@ -253,65 +119,32 @@ public function ref($defaults = [])
...
-            if (isset($this->$key)) {
-                $arr[$key] = $this->$key;
-            }
-        }
-
-        return $arr;
-    }
-
-    // }}}
+    protected $debug_fields = ['our_field', 'their_field', 'type', 'system', 'never_save', 'never_persist', 'read_only', 'ui', 'join'];
```
DarkSide666: should be $__debug_fields instead of $debug_fields.
$link and $model should be added in $__debug_fields too if you agree with my other comment above in Reference::__debug_info method

src/Reference_SQL_One.php
```diff
@@ -117,6 +125,20 @@ function ($m) {
         return $ex;
     }
 
+    public function ref($defaults = [])
+    {
+        $m = parent::ref($defaults);
+
+        // If model is not loaded, then we are probably doing deep traversal
```
DarkSide666: only "probably" but that's not always the case.

How can I get all orders of active clients?
```
$c = new Client();
$c->addCondition('active', true);
$o = $c->ref('Order');
```
what will this return?

romaninsh: something like
```
select * from order where client_id in (select id form client where active = 1)
```

src/Reference.php
romaninsh: please add either now or later.

src/Reference.php
romaninsh: keep it is_scalar. If you spot any problems, create a test-script first, then we can change.

src/Reference.php
romaninsh: ok, agreed.

src/Reference.php
romaninsh: please keep debug info as light as possible. We don't want too much info in there. 2-3 properties is more than enough. I was thinking about trimming it further down.

romaninsh: you can still dump them, it's about what's most important when you use 'var_dump' so i only show stuff that can help identify the problem.

src/Reference.php
```diff
@@ -210,7 +214,7 @@ public function guessFieldType($their_field = null)
     /**
      * List of properties to show in var_dump.
      */
-    protected $debug_fields = ['our_field', 'their_field'];
+    protected $__debug_fields = ['ref' => 'link', 'model', 'our_field', 'their_field'];
```
romaninsh: we actually can simply use 'link' here, so that people are not confused.

src/Reference_SQL_One.php
DarkSide666: then it's OK

src/Reference.php
DarkSide666: OK, but I already merged :)

PR #161, opened 2016-09-16
typecasting not working on tryLoad()

romaninsh:


src/Persistence_SQL.php
```diff
@@ -687,63 +681,56 @@ public function load(Model $m, $id)
...
     /**
-     * Loads any one record.
+     * Tries to load any one record.
      *
      * @param Model $m
      *
      * @return array
      */
-    public function loadAny(Model $m)
+    public function tryLoadAny(Model $m)
```
DarkSide666: tryLoadAny can also use tryLoad($m) to avoid code duplication. That can be possible if in tryLoad method you make $id parameter optional and add condition:
```
if ($id !== null) $load->where($m->getElement($m->id_field), $id);
```

romaninsh: We can do that later ;) this should be good enough fix for now.

DarkSide666: OK I agree

PR #168, opened 2016-09-22
fixes #167

### LINKED ISSUE #167: Enum values doesn't correctly validate
enum can be defined as
```
->enum([
    'legal' => 'Juridiska persona', // non-numeric key, should use key as value
    'individual', // numeric key, should use value as value
])
```
in such case Model->set() throws exception because it only checks enum array values not keys

DarkSide666:


DarkSide666:
Will use this solution instead. Agile data doesn't need to know anything about visible titles of values.

```
$v = [
    'legal' => 'Juridiska persona',
    'individual' => 'Fiziska persona',
];
$field
    ->enum(array_keys($v))
    ->ui(['valueList' => $v]);
```

PR #172, opened 2016-09-29
if condition is added, then field should not be editable in UI

DarkSide666:
Consider case:
```
$client->load(1);
$crud->setModel($client->ref('Order'));
```
currently when you click "Add" in CRUD, then in orders edit form you still see client_id field as editable field, because even if it has system=true, it also has ui[editable]=true by default and can be edited.

This PR restricts editing in UI if there is condition on field.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -710,6 +710,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
 
         if ($f) {
             $f->system = true;
+            $f->ui['editable'] = false;
             if ($operator === '=' || func_num_args() == 2) {
                 $v = $operator === '=' ? $value : $operator;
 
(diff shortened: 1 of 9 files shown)


src/Model.php
```diff
@@ -710,6 +710,7 @@ public function addCondition($field, $operator = null, $value = null)
 
         if ($f) {
             $f->system = true;
+            $f->ui['editable'] = false;
```
romaninsh: I think unless you set editable to true/false explicitly, it falls back to 'system' here:

https://github.com/atk4/data/blob/develop/src/Field.php#L266

Does that not work?

DarkSide666: That works, but for Reference_One fields editable=true is set as default here: https://github.com/atk4/data/blob/develop/src/Reference_One.php#L53

romaninsh: I see and it's passed into field here: https://github.com/atk4/data/blob/develop/src/Reference_One.php#L84

I guess we shouldn't set the default there. Do you think it would work if UI would be null?

romaninsh: No why? It's not system by default so it should show. I need to do tests.

PR #173, opened 2016-09-30
Feature/minor improvements in references

romaninsh:
Also fixes build problem.

### DIFF
--- a/src/Reference_SQL_One.php
+++ b/src/Reference_SQL_One.php
@@ -14,17 +14,24 @@ class Reference_SQL_One extends Reference_One
      *
      * Returns Expression in case you want to do something else with it.
      *
-     * @param string|Field $field
-     * @param string|null  $their_field
-     * @param array        $defaults    Properties
+     * @param string|Field|array $field or [$field, ..defaults]
+     * @param string|null        $their_field
      *
      * @return Field_SQL_Expression
      */
-    public function addField($field, $their_field = null, $defaults = [])
+    public function addField($field, $their_field = null)
     {
+        if (is_array($field)) {
+            $defaults = $field;
+            $field = $defaults[0];
+        } else {
+            $defaults = [];
+        }
+
         if ($their_field === null) {
             $their_field = $field;
         }
+
 
         $e = $this->owner->addExpression($field, array_merge([
             function ($m) use ($their_field) {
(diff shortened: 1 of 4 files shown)


src/Reference_SQL_One.php
```diff
@@ -14,25 +14,33 @@ class Reference_SQL_One extends Reference_One
...
+     * @param string|null        $their_field
      *
      * @return Field_SQL_Expression
      */
-    public function addField($field, $their_field = null, $defaults = [])
+    public function addField($field, $their_field = null)
     {
+        if (is_array($field)) {
+            $defaults = $field;
+            $field = $defaults[0];
```
DarkSide666: probably should check if $defaults[0] is set at all

src/Reference_SQL_One.php
```diff
@@ -43,20 +51,39 @@ function ($m) use ($their_field) {
...
      *
      * @return $this
      */
-    public function addFields($fields = [])
+    public function addFields($fields = [], $defaults = [])
     {
         foreach ($fields as $field => $alias) {
-            if (is_numeric($field)) {
-                $this->addField($alias);
+            $d = $defaults;
```
DarkSide666: not needed row

src/Reference_SQL_One.php
```diff
@@ -43,20 +51,39 @@ function ($m) use ($their_field) {
...
+
+            if (is_array($alias)) {
+                $d = array_merge($defaults, $alias);
+                $alias = $alias[0];
```
DarkSide666: shouldn't we first check if $alias[0] is set at all?

src/Reference_SQL_One.php
romaninsh: it's a bit meaningless not to have there, but ok.

src/Reference_SQL_One.php
romaninsh: why not

romaninsh: adressed.

src/Reference_SQL_One.php
romaninsh: fixed.

src/Reference_SQL_One.php
romaninsh: ok, you are right, fixed.

PR #176, opened 2016-10-03
Implement Model::addRef()

### LINKED ISSUE #92: add ability to create simple relations with addRelation()
Currently using hasOne() and hasMany() you can define related models. That's cool, but sometimes you want to add a custom relation.

Basically when calling $m->ref('AuditLog'); it would respond with a related model. Technically you could create a method  $m->getAuditLog();  but we want to keep all relations separate in case we want to iterate through them (for example by visualising in UI or exploring relations).

Here is invocation:

```
$m->addRelation('AuditLog', function($m) {
    return (new Model_Audit($big_data_db))->addCondition('user_id', $m->id);
});
```

And usage after:

```
$log = $m->load(123)->ref('AuditLog');
$log->addAction(..);
```

or in UI framework:

```
$crud->addRef('AuditLog');
```

romaninsh:
Resolves #92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 1.1.2
+
+Minimum stability fixes after the release #172, #173
+
+
 # 1.1.1
 
 * disabled type fetching from related entities when using ref->addField due to performance degradation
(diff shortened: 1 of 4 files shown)


romaninsh:
i somewhat messed up with merges here. ;)

PR #177, opened 2016-10-04
Remove unwanted brackets

SaitamaSama:


romaninsh:
Thanks, appreciate the fix ;)

PR #178, opened 2016-10-04
adds afterSaveTransaction and afterDeleteTransaction hooks

DarkSide666:
Sometime it is useful to have hook which executes after model save/delete is commited.
For example, if you need to execute some action after model inserts a record, but this action relies on data inserted in afterSave hook.

Not-tested example - just to give you idea:
```
$order->addHook('afterInsert', function ($m) {
    // automatically create new invoice when new order is inserted
    // this executes in same $order->save() transaction
    $order->ref('Invoice')->set([$some_data])->save();

    // and now we also want to get any unpaid invoice and do something with that
    $total = $order->ref('UnpaidInvoice')->loadAny()->get('total');
    // here ref() will add condition on id of invoice which we created few lines above
    // and loadAny() will not find this record, because it's not yet commited.
    $m->set('total', $total)->save();
});
```

so last action should be done after save is commited, but previously we couldn't hook in that spot.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1167,7 +1167,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
             $this->set($data);
         }
 
-        return $this->atomic(function () use ($data) {
+        $a = $this->atomic(function () use ($data) {
             if ($this->hook('beforeSave') === false) {
                 return $this;
             }
@@ -1248,6 +1248,10 @@ class Model implements \ArrayAccess, \IteratorAggregate
 
             return $this;
         });
+
+        $this->hook('afterSaveTransaction', [$a]);
+
+        return $a;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


romaninsh:
I think queuing actions after transaction is committed is not a very good design. I need to think about implications and possible use for this.

I think it should be more like a queue:

```
$db->afterCommit($dsql);
```

execute DSQL or call-back after transaction is finished. Also hook should be done inside persistence, not inside the model, because model can be wrapped yet in other transaction and that would make results unpredictable.

DarkSide666:
OK, you're right. Doing something after transaction is not in scope of model and in my particular case issue was somewhere else - using saveAndUnload() and relying on model field value which was not reloaded. Changing used method to save() solved my issues.

PR #184, opened 2016-10-06
Add serialize support, change typecasting

### LINKED ISSUE #182: Improve encoding practices during saving
Currently when you save a field, it executes `typecastSaveField` that must convert PHP value into database-friendly value. For example for MongoDB it would convert date from DateTime into MongoDate, for SQL it would convert it into "STRING".

In some cases we want to encode / serialize data. For exmaple we want to store "array" into a field or even some arbitrary data (could be array, could be string could be boolean). Also what about storing StdObject? Another case is storing binary data in base64? and what about encrypting data?

If you do not define type, then by definition AgileData will not perform any conversion, but that's not what we want. 

I think there should be another mechanism to encode data.

```
$this->addField('data', ['serialize'=>true]);
```

the above would cause data to be `serialize()`d before it actually hits the database. Upon load the data will be un-serialized. Also we could define several serialization mechanisms:

```
$this->addField('data', ['serialize'=>'base64']);
```

or

```
$this->addField('data', ['serialize'=>'json']);
```

There are however few questions that stop me from implementing this:
## Combining type and serialization

What if I want to use DateTime type and serialize it into SQL?

```
$this->addField('ts', ['type'=>'datetime', 'serialize'=>true]);
```

Unfortunately when used with SQL this would convert datetime into string first, then serialize string and then store. Not really what user migth have wanted. I think DateTime object should be serialized without typecasting.

On other hand, what abut this:

```
$this->addField('ts', ['type'=>'datetime', 'serialize'=>'base64']);
```

base64 can only store strings so attempting to serialize date will cause a problem. Would serialize be different depending on if we store inside Mondo or SQL ? It shouldn't be, right? Creating multiple 'serialize' / 'encode' etc parameters seems like an overkill already.
## json_decode - should it use true/false?

json_encode / decode cannot distinguish between "StdClass" objects and arrays. So you would have to specify which one you want. Do we create 2 different serialization types because of that?
## conflicting with types

Currently we have type "struct" that stores array as json. It does not really allow to store non-array values. I think it should be renamed into 'array' anyway, but tehn would 'struct' shore StdClass objects? what about storing scalar values?

DarkSide666:
- fully implemented `serialize`
- changes in `typecast`
- most of logic moved from Persistence_SQL to Persistence class
- removed `loadCallback` and `saveCallback` properties. Now you should use `$field->typecast = [$typecast_save_callback, $typecast_load_callback]` instead which does the same thing.

ToDo:
- [ ] change current and add a lot of new tests for typecasting and serialization in multiple combinations
- [ ] documentation need changes and few additional paragraphs about typecasting and serialization
- [ ] implement `_typecastSaveField` and `_typecastLoadField` for Persistence_Array

as result this should fix #182


romaninsh:
Those errors in test scripts are really weird, like:

```
3) atk4\data\tests\FolderTest::testRate
atk4\data\tests\AgileExceptionWrapper: Use of obsolete field type abbreviation. Use "integer", "string", "boolean" etc.
```
I don't think that this test-script was added by this PR, so why does it fail suddenly.

PR #187, opened 2016-10-10
Fix incompatible type matching

romaninsh:
Currently, if the field value is "4" and you change it to 4 (int) then this makes field dirty. To avoid this, hotfix was added: 33377f7f52f017581cf566c21bc1f9488af089cc

Unfortunately this resulted in some bad side-effects (not found by tests) reported here: https://github.com/atk4/data/pull/184#commitcomment-19325848

This PR is properly addressing the issue, by using strict comparison unless we are dealing with stringified numbers. Also it adds several tests, so if there are more test-cases that are affected, then please add.

The original problem was caused because hasOne('blah', ); will use non-integer fields even for persistence SQL where `id` fields use int() by default. Still, we can't be fully sure about that and without typecasting SQL response comes back as string.

### DIFF
--- a/tests/BusinessModelTest.php
+++ b/tests/BusinessModelTest.php
@@ -104,6 +104,9 @@ class BusinessModelTest extends TestCase
         $m = new Model();
         $m->addField('name');
         $m->data = ['name' => 5];
+        $m['name'] = 5;
+        $this->assertEquals([], $m->dirty);
+
         $m['name'] = 10;
         $this->assertEquals(['name' => 5], $m->dirty);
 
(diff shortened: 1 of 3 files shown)


romaninsh:
I have added some test-script for typecasting of non-string type's "" into null.

The point here is that any non-string type with value of "" is invalid and is probably a null.

romaninsh:
Those cases seem to be cleaning up after improper use of casting, so technically we could say those are unnecessary, but create a lot of problems around Agile Data.

If those fixes create a problem, please provide examples.

tests/BusinessModelTest.php
```diff
@@ -114,6 +117,37 @@ public function testDirty()
...
+        $m['name'] = 6;
+        $this->assertNotEquals([], $m->dirty);
+        $m['name'] = 5;
+        $this->assertEquals([], $m->dirty);
+
+        $m->data = ['name' => 4.28];
+        $m['name'] = '4.28';
+        $this->assertEquals([], $m->dirty);
+        $m['name'] = '5.28';
+        $this->assertNotEquals([], $m->dirty);
```
DarkSide666: why name didn't become dirty here ???

DarkSide666: ah sorry :D

PR #191, opened 2016-10-18
Feature/id default type

romaninsh:
As per documentation, we use undefined type to preserve value of the field.

Persistence_SQL currently uses 'integer' as default type and the same thing is with global persistence. It's safer not to set any type, because this way we will preserve the original type of id, INT or HASH.

Another effect of this is that hasOne() which defines fields without type by default now won't update "123" to 123 and consider it dirty.


DarkSide666:
looks good

PR #193, opened 2016-10-28
More tests and cleanup of Timezone handling.

romaninsh:
Solves some problems using date when in a different timezones.

This also adds option $field->timezone that can instruct your persistence layer on how to better store the date.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -38,6 +38,12 @@ class Field
     public $enum = null;
 
     /**
+     * For "datetime" type, this can specify how the value is persisted in
+     * the database that lacks timezone support.
+     */
+    public $timezone = null;
+
+    /**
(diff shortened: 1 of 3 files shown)


romaninsh:
note - if you don't spot lack of documentation for the new property, you did review poorly!

DarkSide666:
I guess most of this is already supported in https://github.com/atk4/data/pull/184.
Property is called `persist_timezone` and there is even more properties, like `dateTimeZoneClass` etc.
See https://github.com/atk4/data/pull/184/files and search for term "timezone".
Only thing that is really missing there is a bunch of test-cases.

DarkSide666:
Already implemented in https://github.com/atk4/data/pull/184

PR #194, opened 2016-11-01
Feature/csv persistence

romaninsh:
Still work in progress.


romaninsh:
Documentation is missing.

src/Persistence_CSV.php
```diff
@@ -0,0 +1,281 @@
...
+    public function insert(Model $m, $data)
+    {
+        if (!$this->mode) {
+            $this->mode = 'w';
+        } elseif ($this->mode == 'r') {
+            throw new Exception(['Currently reading records, so writing is not possible.']);
+        }
+
+        if (!$this->handle) {
+            $this->saveHeader();
```
DarkSide666: no such method exist

src/Persistence_CSV.php
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        $id = $this->generateNewID($m, $table);
```
DarkSide666: $table doesn't exist

DarkSide666:
Documentation is there, but there is a lot missing in loading and saving.
* saving in CSV file is not implemented as such.
* not sure why we need $mode attribute and how it will be used

src/Persistence_CSV.php
romaninsh: added.

src/Persistence_CSV.php
romaninsh: fixed.

romaninsh:
Dark, to answer your question about mode - the idea is to prevent user from writing into persistence if he has read from it.

Perhaps we could require user to specify if he wants to read or write to the file, but that can be improved later.

PR #195, opened 2016-11-01
Feature/yet another fix for dirtiness

romaninsh:
In my latest update i have changed to use '===' to compare types for dirtiness and cleaned up no-type handling. Still it's not a perfect way, and objects are never equal:

    $a['date'] = clone $a['date'];

This results in dirty field, and it shouldn't be.

This fix addresses this by comparing types first and if types match, use loose comparison. I think it's the best way to compare values.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -497,7 +497,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
 
         $current_value = array_key_exists($field, $this->data) ? $this->data[$field] : $original_value;
 
-        if ($value === $current_value) {
+        if (gettype($value) == gettype($current_value) && $value == $current_value) {
             // do nothing, value unchanged
             return $this;
         }
--- a/src/Model.php
+++ b/src/Model.php
@@ -529,7 +529,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
         }
 
         if (array_key_exists($field, $this->dirty) && (
-            $this->dirty[$field] === $value
+            gettype($this->dirty[$field]) == gettype($value) && $this->dirty[$field] == $value
         )) {
             unset($this->dirty[$field]);
         } elseif (!array_key_exists($field, $this->dirty)) {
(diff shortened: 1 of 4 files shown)


src/Persistence_SQL.php
```diff
@@ -458,6 +458,9 @@ public function _typecastLoadField(Field $f, $value)
                 // datetime only - set from persisting timezone
                 if ($f->type == 'datetime' && isset($f->persist_timezone)) {
                     $v = $dt_class::createFromFormat($format, $v, new $tz_class($f->persist_timezone));
+                    if ($v === false) {
+                        throw new Exception(['Incorrectly formatted datetime', 'format' => $format, 'value' => $value, 'field' => $f]);
+                    }
                     $v->setTimeZone(new $tz_class(date_default_timezone_get()));
                 } else {
                     $v = $dt_class::createFromFormat($format, $v);
```
DarkSide666: should add if ($v===false) part here too i guess

tests/TypecastingTest.php
```diff
@@ -338,4 +338,45 @@ public function testTimestamp()
...
+                    'date'     => 'foobar',
+                ],
+            ], ];
+        $this->setDB($a);
+        $db = new Persistence_SQL($this->db->connection);
+
+        $m = new Model($db, ['table' => 'types']);
+        $m->addField('ts', ['actual' => 'date', 'type' => 'date']);
+        $m->loadAny();
+        $m['ts'] = new \DateTime();
```
DarkSide666: should be `= new \DateTime('2016-11-01')` or test will fail tomorrow :)

src/Model.php
```diff
@@ -497,7 +497,7 @@ public function set($field, $value = null)
 
         $current_value = array_key_exists($field, $this->data) ? $this->data[$field] : $original_value;
 
-        if ($value === $current_value) {
+        if (gettype($value) == gettype($current_value) && $value == $current_value) {
```
DarkSide666: so `set('1')` for type=integer field will make it dirty?

romaninsh: added test for this. It works OK.

romaninsh: because of normalisation.

src/Persistence_SQL.php
romaninsh: I can't see where the link is pointing to.

romaninsh: ok, that's addressed!

PR #198, opened 2016-11-11
Add implementation for refModel()

romaninsh:
A small PR adding refModel(). This method is handy when you simply want to fetch the model from a reference, without it doing any additional conditioning.


romaninsh:
Documentation is missing..

DarkSide666:
No documentation,but besides that looks OK.

PR #210, opened 2016-12-20
We should allow model to override action (like UnionModel does)

romaninsh:
A few minor tweaks that makes implementation of advanced patterns (such as UnionModel) smoother.

see https://agiletoolkit.org/data/extensions/report for more information on UnionModel features.


romaninsh:
this is already included in #211

PR #211, opened 2016-12-20
Added implementation for safe handling of ReadOnly models and Models without ID

romaninsh:
## Add tests for model without ID
In some cases the model will be used with data that does not have ID field defined. Although technically model would have worked previously, the code wasn't fully tested.

## Add implementation for read_only
Model can be $m->read_only = true, in which case save/delete/insert operations will fail.

### DIFF
--- a/src/Persistence_SQL.php
+++ b/src/Persistence_SQL.php
@@ -590,7 +590,11 @@ class Persistence_SQL extends Persistence
      */
     public function tryLoad(Model $m, $id)
     {
-        $load = $this->action($m, 'select');
+        if (!$m->id_field) {
+            throw new Exception(['Unable to load field by "id" when Model->id_field is not defined.', 'id'=>$id]);
+        }
+
+        $load = $m->action('select');
         $load->where($m->getElement($m->id_field), $id);
         $load->limit(1);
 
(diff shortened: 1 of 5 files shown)


src/Persistence_SQL.php
```diff
@@ -590,7 +590,11 @@ public function action($m, $type, $args = [])
      */
     public function tryLoad(Model $m, $id)
     {
-        $load = $this->action($m, 'select');
+        if (!$m->id_field) {
+            throw new Exception(['Unable to load field by "id" when Model->id_field is not defined.', 'id'=>$id]);
+        }
+
+        $load = $m->action('select');
```
DarkSide666: No need to call Model->action here. It's slightly faster (one function call less) to call persistence->action instead as it was before.

src/Persistence_SQL.php
```diff
@@ -590,7 +590,11 @@ public function action($m, $type, $args = [])
      */
     public function tryLoad(Model $m, $id)
     {
-        $load = $this->action($m, 'select');
+        if (!$m->id_field) {
```
DarkSide666: Shouldn't this be in Model->tryLoad or Persistence->tryLoad instead of Persistence_SQL?
Is there any use case when model can be loaded without knowing its id?

src/Persistence_SQL.php
```diff
@@ -657,7 +662,7 @@ public function load(Model $m, $id)
      */
     public function tryLoadAny(Model $m)
     {
-        $load = $this->action($m, 'select');
+        $load = $m->action('select');
```
DarkSide666: this does the same thing as before, but uses one extra function call ...

romaninsh: except that it allows model to override actions, which is what union model does.

src/Persistence_SQL.php
romaninsh: again, allowing model to override action is important.

src/Persistence_SQL.php
DarkSide666: I see. My bad to not notice that.

PR #213, opened 2017-01-04
Feature/various cleanups

romaninsh:
- Documentation added for Persistence_SQL
 - Lots of documentation bugs resolved (as reported by sphinx, e.g. broken references)
 - added support for field aliasing when executing action() - field, fx and count


romaninsh:
Need to release those fixes in order for https://github.com/atk4/ui/pull/29 not to fail.

PR #217, opened 2017-03-09
Aggregate using IFNULL() by default.

romaninsh:
When we declare "hasMany" relationship and use aggregation, what happens if no related records exist? SUM() without any values returns null, but that's not what we want.

This PR adds a new action 'fx0' that's identical to 'fx' in all ways, but will wrap expression into "IFNULL([], 0)". This way aggregate functions will always return non-null values and will not break any further calculations.

### DIFF
--- a/src/Persistence_SQL.php
+++ b/src/Persistence_SQL.php
@@ -562,7 +562,6 @@ class Persistence_SQL extends Persistence
 
             case 'fx':
             case 'fx0':
-            case 'fx00':
                 if (!isset($args[0], $args[1])) {
                     throw new Exception([
                         'fx action needs 2 arguments, eg: ["sum", "amount"]',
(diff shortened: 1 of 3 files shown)


src/Persistence_SQL.php
```diff
@@ -572,10 +573,16 @@ public function action($m, $type, $args = [])
                 $field = is_string($args[1]) ? $m->getElement($args[1]) : $args[1];
                 $this->initQueryConditions($m, $q);
                 $m->hook('initSelectQuery', [$q, $type]);
+
+                $expr = "$fx([])";
+                if ($type == 'fx0') {
+                    $expr = "ifnull($expr, 0)";
```
DarkSide666: This is wrong. Instead of:
```
$expr = "$fx([])";
if ($type == 'fx0') {
    $expr = "ifnull($expr, 0)";
}
```
should be simply (ifnull should be inside sum):
```
$expr = ($type == 'fx0' ? "$fx(ifnull([]), 0)" : "$fx([])");
```

PR #225, opened 2017-04-09
Implement 'required' property

### LINKED ISSUE #223: normalize() must not be called twice
Introduced in 6919e967, normalize() during typecasting shouldn't be executed. 

Need to be removed, because model should only store normalized values.

romaninsh:
- Added new "required" property
 - Required is also mandatory
 - Numeric/float/money values are normalized better by taking out all non-essential characters (e.g. currency symbol)
 - Field exceptions cleaned up
 - Removed double-normalize when saving row to DB, resolves #223 
 - Loading 'time' from DB will properly reset "date" on the DateTime instead of using current date.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -221,9 +221,15 @@ class Field
             if (!$this->owner->strict_types) {
                 return $value;
             }
+
             if ($value === null) {
+                if ($this->required) {
+                    throw new Exception('may not be null');
+                }
+
                 return;
             }
+
             $f = $this;
 
             // only string type fields can use empty string as legit value, for all
--- a/src/Field.php
+++ b/src/Field.php
@@ -231,14 +231,14 @@ class Field
             // other field types empty value is the same as no-value, nothing or null
             if ($f->type && $f->type != 'string' && $value === '') {
                 if ($this->required && empty($value)) {
-                    throw new Exception('may not be a zero');
+                    throw new Exception('may not be a empty');
                 }
 
                 return;
             }
 
             switch ($f->type) {
-            case null:
+            case null: // loose comparison, but is OK here
                 if ($this->required && empty($value)) {
                     throw new Exception('must not be empty');
                 }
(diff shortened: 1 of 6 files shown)


src/Field.php
```diff
@@ -106,14 +106,26 @@ class Field
...
+     * it's an empty value.
+     *
      * Can contain error message for UI.
      *
      * @var bool|string
      */
     public $mandatory = false;
 
     /**
+     * Required field must have non-empty value. A null value can still be
```
DarkSide666: `required` should not accept null values because `empty(null)===true`

src/Field.php
```diff
@@ -204,96 +218,133 @@ public function __construct($defaults = [])
...
-        case 'money':
-            if (!is_numeric($value)) {
-                throw new Exception('Field value must be numeric');
+            $f = $this;
+
+            // only string type fields can use empty string as legit value, for all
+            // other field types empty value is the same as no-value, nothing or null
+            if ($f->type && $f->type != 'string' && $value === '') {
+                if ($this->required && empty($value)) {
+                    throw new Exception('may not be a zero');
```
DarkSide666: wrong message "zero"

src/Field.php
```diff
@@ -204,96 +218,133 @@ public function __construct($defaults = [])
...
+
+                return;
             }
-            $value = round($value, 4);
-            break;
-        case 'boolean':
-            if (is_bool($value)) {
+
+            switch ($f->type) {
+            case null:
```
DarkSide666: Switch statement use loose comparison, so `null` can't be used here because it will be the same as `""` or `0`.
See also http://stackoverflow.com/a/7091795/1466341

src/Field.php
```diff
@@ -222,23 +221,29 @@ public function normalize($value)
...
             }
+
             $f = $this;
 
             // only string type fields can use empty string as legit value, for all
             // other field types empty value is the same as no-value, nothing or null
             if ($f->type && $f->type != 'string' && $value === '') {
                 if ($this->required && empty($value)) {
-                    throw new Exception('may not be a zero');
+                    throw new Exception('may not be a empty');
```
romaninsh: Typo

PR #226, opened 2017-04-09
fix caption for hasOne fields

DarkSide666:


src/Reference_One.php
```diff
@@ -61,6 +61,15 @@ class Reference_One extends Reference
     public $read_only = false;
 
     /**
+     * Defines a label to go along with this field. Use getCaption() which
+     * will always return meaningfull label (even if caption is null). Set
+     * this property to any string.
+     *
+     * @var string
+     */
+    public $caption = null;
```
romaninsh: Are we not extending field?

tests/FieldTest.php
```diff
@@ -107,17 +107,17 @@ public function testMandatory3()
...
-        $m->addField('foo2', ['caption'=>'My Foo']);
-        $this->assertEquals('My Foo', $m->getElement('foo2')->getCaption());
+        $f = $m->addField('foo2', ['caption'=>'My Foo']);
+        $this->assertEquals('My Foo', $f->getCaption());
 
-        $m->addField('foo3', ['ui'=>['caption'=>'My Foo']]);
-        $this->assertEquals('My Foo', $m->getElement('foo2')->getCaption());
+        $f = $m->addField('foo3', ['ui'=>['caption'=>'My Foo']]);
+        $this->assertEquals('My Foo', $f->getCaption());
     }
```
romaninsh: I think tests were OK Before. But is there test for relations?

src/Reference_One.php
DarkSide666: No. We extend Reference which is separate class from Field. That's why we have to redefine these properties here.

tests/FieldTest.php
DarkSide666: Tests were OK. I just made them a bit shorter :)

There is no test for caption in relations. Will do in a short while.

PR #232, opened 2017-06-12
Feature/implement validation

romaninsh:
Implements a consistent validation for models.

For examples, see https://github.com/atk4/data/blob/3ab7741e8d0956671c3c309d9f55ac963faefa1f/src/ValidationException.php

TODO: documentation

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -329,7 +329,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
      * You may also use format:
      *  ['field'=>['must not have character [ch]', 'ch'=>$bad_character']] for better localization of error message.
      *
-     * Always use 
+     * Always use
      *   return array_merge(parent::validate($intent), $errors);
      *
      * @param string $intent By default only 'save' is used (from beforeSave) but you can use other intents yourself.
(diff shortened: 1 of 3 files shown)


src/Model.php
```diff
@@ -1249,6 +1275,9 @@ public function save($data = [], Persistence $to_persistence = null)
         }
 
         return $this->atomic(function () use ($to_persistence) {
+            if (($errors = $this->validate('save')) !== []) {
+                throw new ValidationException($errors);
```
DarkSide666: Probably you should add 2nd parameter here like this `throw new ValidationException($errors, 'save');`

src/ValidationException.php
```diff
@@ -0,0 +1,40 @@
...
+            foreach ($errors as $field=>$error) {
+                return parent::__construct([
+                    $error,
+                    'field'=> $field,
+                ]);
+            }
+        }
+
+        return parent::__construct([
+            'Incorrectly use of ValidationException, argument should be an array',
```
DarkSide666: "Incorrect use ..." not "Incorrectly use ..." :)

src/Model.php
romaninsh: tried to think of a good reason why, but just cant find

src/ValidationException.php
romaninsh: fixed

src/Model.php
DarkSide666: because ValidationException use and show 2nd parameter :)

PR #240, opened 2017-09-03
Feature/integrate core 1 3

romaninsh:
Actually nothing much was done here, simply added trait.


romaninsh:
I'll work on improving test coverage..

PR #246, opened 2017-09-11
Fix array persistence data error

### LINKED ISSUE #245: Array Persistence without table name ignores first record
Consider this example:

``` php
    $my_array = [
        ['name'=>'Vinny', 'surname'=>'Sihra', 'birthdate'=>new \DateTime('1973-02-03')],
        ['name'=>'Zoe', 'surname'=>'Shatwell', 'birthdate'=>new \DateTime('1958-08-21')],
        ['name'=>'Darcy', 'surname'=>'Wild', 'birthdate'=>new \DateTime('1968-11-01')],
        ['name'=>'Brett', 'surname'=>'Bird', 'birthdate'=>new \DateTime('1988-12-20')],
    ];
    $goodData = ['data'=>$my_array];

    $m = new Model(new Persistence_Array($goodData), 'data');
    $m->addField('name');
```

When model initialized like that, executing foreach works well:

``` php
    foreach($m as $junk) {
        echo $m['name']."\n";
    }
```

This will result in output:

```
Vinny
Zoe
Darcy
Brett
```

Next change those two lines:

``` php
    $goodData = $my_array;
    $m = new Model(new Persistence_Array($goodData));
```

The Persistence_Array supports no-table arrays like that, however it fails to load values for the first column resulting in output:

```
  (null here)
Zoe
Darcy
Brett
```

DarkSide666:
* fix data[0] = &array() error
* fix #245

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -272,11 +272,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
             $persistence = null;
         }
 
-        foreach ($defaults as $key => $val) {
-            if ($val !== null) {
-                $this->$key = $val;
-            }
-        }
+        $this->setDefaults($defaults);
 
         if ($persistence) {
             $persistence->add($this, $defaults);
--- a/src/Model.php
+++ b/src/Model.php
@@ -263,13 +263,14 @@ class Model implements \ArrayAccess, \IteratorAggregate
      */
     public function __construct($persistence = null, $defaults = [])
     {
-        if (is_string($defaults) || $defaults === false) {
-            $defaults = [$defaults];
-        }
-
+        // persistence is optional
         if (is_array($persistence)) {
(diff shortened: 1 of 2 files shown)


romaninsh:
I don't think it actually addresses #245. Because everything works with table.

DarkSide666:
It does. See skype.

src/Model.php
```diff
@@ -272,11 +272,7 @@ public function __construct($persistence = null, $defaults = [])
             $persistence = null;
         }
 
-        foreach ($defaults as $key => $val) {
-            if ($val !== null) {
-                $this->$key = $val;
-            }
-        }
+        $this->setDefaults($defaults);
```
romaninsh: fix to ['table'=>$defaults] on line 267

PR #253, opened 2017-11-14
Feature/addfield set ui false

### LINKED ISSUE #252: calling $field->addTitle() should set $field->ui['visible'] = false.
A common case is that Grid should contain a title field, but form should contain the 'ID' field. We are excluding calculated fields from the form, but Grid still contains both fields.

This should be a very minor fix. Obviously do not change the value if it's already set to 'true' or 'false'.

DarkSide666:
fix #252

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -200,7 +200,7 @@ class Reference
      *
      * @param string $their_field
      *
-     * @return string
+     * @return string|null
      */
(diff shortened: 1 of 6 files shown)


romaninsh:
Gotta test this: https://codecov.io/gh/atk4/data/pull/253/src/src/ValidationException.php#L46

PR #255, opened 2017-11-16
add defaults for model in addTitle()

DarkSide666:


romaninsh:
few reason why I don't think this PR is appropriate:

1. it's coming after hasOne() and you can already pass parameters to the model in 2nd field.
2. addTitle() and withTitle() are short-cut methods, they are designed to be simple
3. if we are adding arguments to them, we should also change addFields() and addField()

everything becomes much more complex. we love simplicity.

Issue #256, opened 2017-11-17
sqlite-related issues.


romaninsh:
It's not possible to create and use CRUD with SQLite backend due to multiple faults:

1. DNS gets appended ';charset' value. 

As per http://php.net/manual/en/ref.pdo-sqlite.php i am using `sqlite:db.sqlite3` format, but instead it creates `db.sqlite3;charset=utf8`

2. CRUD / Form simply shows blank page!

To reproduce:

``` shell
echo 'create table if not exists user(id int not null primary key, name varchar, email varchar, password varchar, is_admin bool);' | sqlite3 db.sqlite3
```

``` php
class User extends \atk4\data\Model {
    public $table = 'user';

    function init()
    {
        parent::init();

        $this->addField('name');
        $this->addField('email');
        $this->addField('password', [
            'type'=>'password', 
            'serialize'=>'password_hash',
        ]);
    }
$db = \atk4\data\Persistence::connect('sqlite:db.sqlite3');
$app = new \atk4\ui\App();
$app->initLayout('Centered');
$app->add('Form', new \atk4\login\Model\User($db));
```

romaninsh:
OK, first issue resolved, second is no-issue. Correct syntax is:

``` php
$app->add('Form')->setModel(new User($db));
```

PR #257, opened 2017-11-17
Feature/fix 256

romaninsh:
fix for #256


romaninsh:
Because this issue is a bug, i'd merge it even if coverage is insufficient.

PR #259, opened 2017-11-23
implementation of [password]'s - storing them encrypted.

romaninsh:
The significancy of this implementation is one-way password verification.

So a new method would be introduced for model to verify if field match.  Basic usage is:

``` php
$model->addField('p', ['type'=>'password']);

$model->load(1);
var_dump($model['p']);  // Output will be NULL
```

Passwords must be protected and there is no reason why you would want to retrieve password from the database except if you want to verify it, but we're adding a new method for you:

``` php
var_Dump($model->compare('p', $old_password));  // Output will be true
```

In fact `compare()` can now be used on non-password fields too. If you wish to compare some values making sure they match. This can be further extended in field class. If you require a custom password encryption mechanic:

``` php
$model->addField('legacy_p', ['type'=>'password', 'encrypt'=>$fx1, 'compare'=>$fx2, 'decrypt'=>false]);
```

The above not only specifies your own functions for encryption and decryption, but will disable default `return null` decryption function which is used for the password type, this way you can actually retrieve your password values if you need them.

**UPDATE** Final implementation uses:

``` php
$model->addField('mypass', ['Password']);
```


DarkSide666:
Sounds good.

Except:
* We already have `type=serialize` which in Persistence class implements something similar, so I think that encrypt and decrypt should work the same as serialize works now. It should extend serialize functionality and just add md5 or sha "serialization".

Maybe something like this:
```
if ($field->type == 'password') {
    // $field->serialize = [$callable_encoding_function, $callable_decoding_function];
    $field->serialize = [function ($v) {return md5($v);}, function ($v) {return null;}];
}
```

* Not sure how `compare()` should work. Should it query database for actual value or not? Maybe actual password value should be cached in protected property of field and then just used in `compare`.
Something like this:
```
// cache actual password
PROTECTED $actual_password;
if ($field->type == 'password') {
    $field->serialize = [
        // encoding function
        function ($v) { return md5($v); },
        // decoding function
        function ($v) use ($field) { $field->actual_password = $v; return null; }
    ];
}

// comparing function
public function compare($field, $value) {
    return $value == ($field->type=='password' ? $field->actual_password : $field->get());
}
```

romaninsh:
it's probably easier to implement password with serialize. I'll try that and if I don't like it, i'll refactor. On the second note - yes you are correct, we should probably cache it.

romaninsh:
Actually, if `get()` after `set()` would immediately return null it would screw up UI forms.

src/Field/Password.php
```diff
@@ -0,0 +1,73 @@
...
+        $this->password_hash = null;
+
+        return parent::normalize($value);
+    }
+
+    /**
+     * When storing password to persistance, it will be encrypted. We will
+     * also update $this->password_hash, in case you'll want to perform
+     * verify right after.
+     */
```
DarkSide666: no @param / @return

src/Field/Password.php
```diff
@@ -0,0 +1,73 @@
...
+
+class Password extends \atk4\data\Field
+{
+    use InitializerTrait {
+        init as _init;
+    }
+
+    public $type = 'password';
+
+    protected $password_hash = null;
```
DarkSide666: no @var

src/Field/Password.php
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        $this->password_hash = password_hash($password, PASSWORD_DEFAULT);
+
+        return $this->password_hash;
+    }
+
+    /**
+     * Verify if the password user have suppplied you with is correct.
+     */
```
DarkSide666: @param, @return

src/Model.php
```diff
@@ -352,7 +352,7 @@ public function validate($intent = null)
      */
     public function addField($name, $defaults = [])
     {
-        $field = $this->factory($this->mergeSeeds($this->_default_seed_addField, $defaults), null, 'Field');
+        $field = $this->factory($this->mergeSeeds($defaults, $this->_default_seed_addField), null, '\atk4\data\Field');
```
DarkSide666: which one will take precedence here - `$defaults` or `$this->_default_seed_addField` ?

src/Model.php
```diff
@@ -623,6 +623,25 @@ public function get($field = null)
...
+    /**
+     * You can compare new value of the field with existing one without
+     * retrieving. In the trivial case it's same as ($value == $model[$name])
+     * but this method can be used for:.
+     *
+     *  - comparing values that can't be received - passwords, encrypted data
+     *  - comparing images
+     *  - if get() is expensive (e.g. retrieve object)
+     *
+     * @param string $field
```
DarkSide666: string $name not $field :P

tests/PasswordTest.php
```diff
@@ -0,0 +1,81 @@
...
+    public function testPasswordField()
+    {
+        $m = new Model(); //$db, 'job');
+
+        $m->addField('p', ['Password']);
+
+        $m['p'] = 'mypass';
+
+        // when setting password, you cannot retrieve it back
+        $this->assertEquals('mypass', $m['p']);
```
DarkSide666: so this test should fail, because `m['p']` should be `null`.

tests/PasswordTest.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $a = [];
+        $p = new Persistence_Array($a);
+        $m = new Model($p);
+
+        $m->addField('p', ['Password']);
+
+        $m['p'] = 'mypass';
+
+        $this->assertEquals('mypass', $m['p']);
```
DarkSide666: why this don't fail ?

src/Model.php
romaninsh: left has precedence.

tests/PasswordTest.php
romaninsh: only after it's saved. Otherwise we can't save it.

tests/PasswordTest.php
romaninsh: same as above - until you save password it will work like this.

src/Field/Password.php
```diff
@@ -0,0 +1,93 @@
...
+
+    public function init()
+    {
+        $this->_init();
+
+        $this->typecast = [
+            [$this, 'encrypt'],
+            function ($v, $f, $p) {
+                $this->password_hash = $v;
+                if ($p instanceof \atk4\ui\Persistence\UI) {
```
romaninsh: I'm not particularly satisfied with this implementation, after all, Agile Data shouldn't have any references to Agile UI. 

Technically Password field implementation could have been moved into a separate add-on. Perhaps we should move it into `atk4/login` ?

PR #261, opened 2017-11-30
Feature/use seed instead of class

romaninsh:
Currently Model uses _default_class_addField. This was OK in the early days but now we have a very solid implementation of factory(). This allows us to replace class names with seeds for consistency and increased flexibility.

There will also be a lot of simplifications, general removal of setDefaults(). There shouldn't be any impact other than cleaning up code, because class names are only redefined in persistences.

### DIFF
--- a/src/Field_SQL_Expression.php
+++ b/src/Field_SQL_Expression.php
@@ -30,6 +30,16 @@ class Field_SQL_Expression extends Field_SQL
     public $read_only = true;
 
     /**
+     * Specifies how to aggregate this
+     */
+    public $aggregate = null;
+
+    /**
+     * Specifies which field to use
+     */
+    public $field = null;
+
+    /**
      * Initialization.
      */
     public function init()
--- a/src/Field_SQL_Expression.php
+++ b/src/Field_SQL_Expression.php
@@ -30,12 +30,12 @@ class Field_SQL_Expression extends Field_SQL
     public $read_only = true;
 
     /**
-     * Specifies how to aggregate this
+     * Specifies how to aggregate this.
      */
     public $aggregate = null;
 
     /**
-     * Specifies which field to use
+     * Specifies which field to use.
      */
     public $field = null;
 
--- a/src/Model.php
+++ b/src/Model.php
@@ -26,35 +26,35 @@ class Model implements \ArrayAccess, \IteratorAggregate
      *
      * @var string
      */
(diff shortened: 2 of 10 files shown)


src/Field_SQL_Expression.php
```diff
@@ -29,6 +29,16 @@ class Field_SQL_Expression extends Field_SQL
      */
     public $read_only = true;
 
+    /**
+     * Specifies how to aggregate this.
+     */
+    public $aggregate = null;
```
romaninsh: im not sure if those are needed, but it's being passed through defaults...

romaninsh:
CI will fail until https://github.com/atk4/core/pull/67 is merged.......

src/Model.php
```diff
@@ -257,14 +258,14 @@ class Model implements \ArrayAccess, \IteratorAggregate
...
      *  - it's shorter
      *  - type hinting will work;
+     *  - you can specify string for a table
      *
      * @param Persistence|array $persistence
-     * @param array             $defaults
+     * @param string|array      $defaults
      */
-    public function __construct($persistence = null, $defaults = [])
+    public function __construct($persistence = null, $defaults = null)
```
DarkSide666: > Function `__construct` has a Cognitive Complexity of 7 (exceeds 5 allowed). Consider refactoring.

src/Model.php
```diff
@@ -1633,15 +1637,15 @@ public function leftJoin($foreign_table, $defaults = [])
     protected function _hasReference($c, $link, $defaults = [])
     {
         if (!is_array($defaults)) {
-            $defaults = ['model' => $defaults ?: 'Model_'.$link];
+            $defaults = ['model' => $defaults ?: 'Model\\'.$link];
```
DarkSide666: so we still have `Model` prefix or namespace hard-coded?

src/Reference.php
```diff
@@ -67,24 +68,9 @@ class Reference
      *
      * @param array $defaults
```
DarkSide666: need to change @param

src/Model.php
romaninsh: that's the rare case when you use:

```
$this->hasMany('User');
```

do you use that anywhere?

DarkSide666: Sure use it like that and a lot :)

romaninsh: reverted.

PR #265, opened 2017-12-16
Feature/add static persistence

romaninsh:
Implementation of Static Persistence. This is handy when you need a one-liner for creating a model. Simple use:

``` php
$m = new Model(Persistence_Static(['female', 'male']));

// model can be used anywhere, such as in a table
$table->setModel($m);
```

The other use-case if you have associative array like this:

``` php
$m = new Model(Persistence_Static([
    ['icon'=>'map marker', 'title'=>'Krolewskie Jadlo', 'descr'=>'An excellent polish restaurant, quick delivery and hearty, filling meals'],
    ['icon'=> 'map marker', 'title'=>'Xian Famous Foods', 'descr'=>'A taste of Shaanxi\'s delicious culinary traditions, with delights like spicy cold noodles and lamb burgers.'],
    ['icon'=> 'check', 'title'=>'Sapporo Haru', 'descr'=>'Greenpoint\'s best choice for quick and delicious sushi'],
]);
```

Static persistence will do few things:

- [x] Unlike "Array" no need to pass array by reference.
- [x] Will support either array of strings or array of hashes.
- [x] Will also support "array of arrays".
- [x] If you supply "id" column, will use that, otherwise will use key (0, 1, etc)
- [x] Will use first record from your hash to identify columns
- [x] If column is missing in your model, will add it with `addColumn()`
- [x] Will use value to determine type (string, int, date)
- [x] If column is named "title" or "name" will be used as a title column.
- [x] if no "title" or "name" column exist, will use first column as a title.

--------

 - Documentation: [docs/static.rst](docs/file.rst)
 - Demo: above

### DIFF
--- a/docs/fields.rst
+++ b/docs/fields.rst
@@ -2,13 +2,18 @@
 
 .. _Fields:
 
-======
-Fields
-======
+.. php:namespace:: atk4\data
 
-Field represents a model `property` which we do refer to as field throughout
-Agile Data, to distinguish it from object properties. Fields inside a model
-normally have a corresponding instance of Field class.
+.. php:class:: Field
+
+
+=====
+Field
+=====
+
+Field represents a model `property` that can hold informaiton about your entity.
+In Agile Data we call it a Field, to distinguish it from object properties. Fields
+inside a model normally have a corresponding instance of Field class.
 
 See :php:meth:`Model::addField()` on how fields are added. By default,
 persistence sets the property _default_seed_addField which should correspond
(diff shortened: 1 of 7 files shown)


syrm:
:+1: NO_BC_BREAK

DarkSide666:
Few things:

> If you supply "id" column, will use that, otherwise will use key (0, 1, etc)

Can it also somehow take Model->id_field into account?

> Will use value to determine type (string, int, date)

Quite risky way if it detects only by first data row.In first row I can have `null` value in `date` column for example.

> If column is named "title" or "name" will be used as a title column.

Will it respect Model->title_field?

> If no "title" or "name" column exist, will use first column as a title.

Shouldn't first column be `id` column? Or we use array keys as ids?

PR #266, opened 2017-12-16
Feature/add values property

romaninsh:
Implements $field->values property.

``` php
$model->addField('gender', ['values'=>['F'=>'Female', 'M'=>'Male']]);
```

 - Documentation: [docs/file.rst](docs/file.rst) TODO

(merge #265 first)

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -388,7 +388,7 @@ class Field
      */
     public function set($value)
     {
-        $this->owner[$this->short_name] = $value;
+        $this->owner->set($this->short_name, $value);
 
         return $this;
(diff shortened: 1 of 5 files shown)


romaninsh:
I feel that we should implement a validation for 'values' in a similar way how it's done for 'enum'.

DarkSide666 (changes requested):
Finally this property! I was waiting for this a long time :)

BUT:
- need to add proper validation in Model->set() method
- need some tests
- need documentation

DarkSide666 (approved):

PR #272, opened 2018-01-07
Feature/oracle support

DarkSide666:
* fix tests
* fix `expr()` method (at least partly)

### DIFF
--- a/src/Join_SQL.php
+++ b/src/Join_SQL.php
@@ -36,6 +36,12 @@ class Join_SQL extends Join implements \atk4\dsql\Expressionable
      */
     public function getDSQLExpression($q)
     {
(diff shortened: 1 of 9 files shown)


src/Join_SQL.php
```diff
@@ -36,6 +36,12 @@ public function getDesiredName()
      */
     public function getDSQLExpression($q)
     {
+        // If our Model has expr() method (inherited from Persistence_SQL) then use it
+        if ($this->owner->hasMethod('expr')) {
+            return $this->owner->expr('{}.{}', [$this->foreign_alias, $this->foreign_field]);
+        }
```
romaninsh: join_sql shouldn't even be called if expr is undefined. I think we should leave it here to produce error.

romaninsh:
looks good.

PR #275, opened 2018-01-11
Fix initQueryFields (274)

### LINKED ISSUE #274: After latest composer update, warning is thrown in Persistence_SQL->initQueryFields()
After the latest update yesterday, I get a lot of Warnings about a invalid argument for foreach:
`<b>Warning</b>:  Invalid argument supplied for foreach() in <b>/var/www/html/TourManagement/vendor/atk4/data/src/Persistence_SQL.php</b> on line <b>240</b><br />`

Comment says that the function initQueryFields parameter 3 is either an array (of fields) or null:
```
 /**
     * Adds model fields in Query.
     *
     * @param Model            $m
     * @param \atk4\dsql\Query $q
     * @param array|null       $fields
     */
```
If only an array or null are valid values for $fields, the following check could be changed from:
       ` if ($fields) {`
to
        `if (is_array($fields)) {`

This makes my warnings disappear.
The warning is caused by a lot of select functions like Model->tryLoad() or Model->export().

Best regards
Philipp

DarkSide666:
fix #274

### DIFF
--- a/src/Persistence_SQL.php
+++ b/src/Persistence_SQL.php
@@ -233,7 +233,7 @@ class Persistence_SQL extends Persistence
      */
     public function initQueryFields($m, $q, $fields = null)
     {
-        if ($fields) {
+        if (is_array($fields)) {
 
             // Set of fields is strictly defined for purposes of export,
             // so we will ignore even system fields.


src/Persistence_SQL.php
```diff
@@ -233,7 +233,7 @@ public function initField($q, $field)
      */
     public function initQueryFields($m, $q, $fields = null)
     {
-        if ($fields) {
+        if (is_array($fields)) {
```
romaninsh: maybe `is_array($fields) && $fields)`

DarkSide666: `[]` should be the same as `false` - no fields

PR #285, opened 2018-01-31
Model Sequence, lastInserId and other fixes

DarkSide666:
* implement model->sequence
* fix regex to work for uppercase field names too
* lastInsertID should bwe called from Connection and pass model as parameter. This is needed for sequence support because sequence is model/table specific and lastInsertId should know which model/table we have to check.

 - [ ] We probably need to add few sentences in documentation about Model->sequence and Oracle support in general ?


DarkSide666:
Tests will fail because this PR requires DSQL https://github.com/atk4/dsql/pull/133 first.

romaninsh:
ok, still looks good!

romaninsh:
Imants, i think

https://github.com/atk4/data/blob/develop/src/Reference_SQL_One.php#L44

which is currently `return $m->refLink($this->link)->action('field', [$their_field]);`
should be `return $m->refLink($this->link)->action('field', [$their_field])->reset('order');`

romaninsh:
i approve "even more proper fix" 🗡

PR #286, opened 2018-02-07
Allows static persistence to use empty array

romaninsh:
Currently the following code fails:

``` php
$table->setSource([]);
```

The issue is not with the table, but with static persistence. Currently it attempts to load first row and reverse-engineer fields from that, but if there are no rows - error is shown.

This PR fixes this problem and additionally implements `tryLoadAny` for Persistence_Array.

Here I have faced a small problem. "load($id)" would load the record and if successful, the $id will be placed in $model->id.

tryLoadAny does not know ID, it should be coming from the record, but Array specifies "id" as a key, so it's not defined by id_field. I had to add extra check for that.

I hope this does not interfere with SQL.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1238,7 +1238,9 @@ class Model implements \ArrayAccess, \IteratorAggregate
         $this->data = $this->persistence->tryLoadAny($this);
         if ($this->data) {
             if ($this->id_field) {
-                $this->id = $this->data[$this->id_field];
+                if (isset($this->data[$this->id_field])) {
+                    $this->id = $this->data[$this->id_field];
+                }
             }
 
             if ($this->hook('afterLoad') === false) {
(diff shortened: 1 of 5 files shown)


src/Model.php
```diff
@@ -1238,7 +1238,9 @@ public function tryLoadAny()
         $this->data = $this->persistence->tryLoadAny($this);
         if ($this->data) {
             if ($this->id_field) {
-                $this->id = $this->data[$this->id_field];
+                if (isset($this->data[$this->id_field])) {
```
DarkSide666: I think it's better to use `if (array_key_exists($this->id_field, $this->data))` here in case we actually have id=0 in some strange use-cases.

romaninsh: i was actually considering it, but decided against it.

Issue #288, opened 2018-02-07
add travis testing for postgresql


romaninsh:
Now that DSQL supports PostgreSQL and #287 adds "pgsql://" as a recognized connection string, we should add support into travis.ci and run everything on postgresql.

The approach should be quite similar to this:

https://github.com/atk4/dsql/commit/c8d1b403998a919ba3a11465f5f7e22bd00ca388

although there may be some hiccups.

gartner:
Lots of hiccups :-/ 
The tests are written for sqlite (I think) and some of them fail on mysql because the use of " to quote identifiers. Others fails, because they are compared to a string with " for quotes, but on mysql the resulting query is using ` ...

romaninsh:
we can roll with it, simply by ignoring non-functional tests.

PR #292, opened 2018-02-12
Travis db testing

gartner:
Add travis testing on both mysql and postgresql, see #288 

- the tests are running now, but some fail on mysql, others fail on pgsql. Will look into that. @

### DIFF
--- a/tests/Structure.php
+++ b/tests/Structure.php
@@ -31,11 +31,20 @@ class Structure extends Expression
 
     public function id($name = null)
     {
+        $driver = $this->connection->connection()->getAttribute(\PDO::ATTR_DRIVER_NAME);
+
         if (!$name) {
             $name = 'id';
         }
 
-        $val = $this->expr('integer primary key autoincrement');
+        switch ($driver) {
+            case 'pgsql':
+                $val = $this->expr('serial primary key');
+                break;
+
+            default:
+                $val = $this->expr('integer primary key autoincrement');
+        }
 
         $this->args['field'] =
             [$name => $val] + (isset($this->args['field']) ? $this->args['field'] : []);
--- a/tests/Structure.php
+++ b/tests/Structure.php
@@ -70,7 +70,7 @@ class Structure extends Expression
 
             $type = strtolower(isset($options['type']) ?
                 $options['type'] : 'varchar');
-            $type = preg_replace('/[^a-z0-9]+/', '', $type);
+            $type = preg_replace('/[^()a-z0-9]+/', '', $type);
 
             $len = isset($options['len']) ?
                 $options['len'] :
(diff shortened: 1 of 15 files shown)


romaninsh:
This looks well. For those tests that verify the specific syntax with MySQL style escaping, simply add ignore for PostgreSQL.

tests/Structure.php
```diff
@@ -14,6 +14,11 @@ class Structure extends Expression
         'drop'   => 'drop table if exists {table}',
     ];
 
+    public function setEscapeChar($char)
+    {
+        $this->escape_char = $char;
+    }
+
```
romaninsh: this class is actually defunct and should me removed in favour of atk4/schema. I'll add a separate ticket for that.

gartner: I saw the notes about Structure.php somewhere, but would just tackle one thing at a time. So I went with just getting it running.

romaninsh: yes, that's good. I'm on vacation now but if i find some free time i'll add ignores to make your PR work and will merge.

romaninsh:
I have added some skips for generated sql code into https://github.com/atk4/data/tree/pr/292. Still there are other errors.

gartner:
I am currently looking into these errors. For postgresql, there are lots of problems with duplicate primary keys...

gartner:
Some of the errors that appear when testing against postgresql, are there because of the difference in autu-numbering primary keys. MySQLs auto_increment behaves a little like using "SELECT max() + 1 as id', so if you manually sets an id for a row and insert it, the next row you try to insert without a set id, will be given the number that comes next.
Postgresql uses sequences, which are only incremented if they are actualy used. So inserting a row where you manually set the id to 1, and then the next row you have not set the id yourself, it will use the sequence. Which will now return1, as it is first time it is used. And then the test fails, id 1 is already used.

What is the most correct way to fix this? Should I just mark them as expected to fail when database is postgresql?

DarkSide666:
Why not use new `$model->sequence` feature for postgresql?

gartner:
First of, I haven't implemented the sequence-feature for postgresql. But I cannot image that to be a big problem, when I can look at how its done for Oracle.

But also, I am under the impression that atk4/dsql is to abstract away the features/quirks/differences on the backend databases. Therefore, in my mind it would be wrong to make different setups for this many of the tests, since it is not really what is tested. 

The sequence-feature could be a solution, but then that feature should also be found for mysql and sqlite, i think.

romaninsh:
sequence is for explicit use of sequences. In this case it's probably good just to skip tests which are not bogus.

romaninsh:
Tests seem to work now fine, so I'm going to merge this.

romaninsh:
Another TODO: merge coverage reports, otherwise it seems that we're decreasing coverage rather than increasing.

PR #296, opened 2018-02-26
use ValidationException

DarkSide666:
We should use `ValidationException` instead of `Exception` when we talk about validating field values.


DarkSide666:
Type of value is type checking and that's maybe not validation.
But checking for empty or null values definitely is validation and should throw appropriate exception to be able to catch it in UI.

PR #313, opened 2018-03-23
Implement export() with field values used as array keys

### LINKED ISSUE #311: Model->export: using ID as first level array key?
Whenever I use the export function of Model class, I want to access the data by the ID. So far, the keys of the first level array are not used, its just 0,1.2.3 and so on.
I wrote a hack which uses the ID as key of the first level of the array instead:

   ```
 /*
     * returns all the defined fields, the key of each array element is its id
     *
     * @param string|array Names of fields to load
     *
     * @return array
     */
    public function exportById($fields) {
        if(is_string($fields)) {
            $fields = [$fields];
        }
        if(!is_array($fields)) {
            throw new \Exception('exportById needs an array with field names as parameter');
        }
        if(!$this->id_field) {
            throw new \Exception('This model does not have an ID field defined');
        }
        $return = [];
        array_unshift($fields, $this->id_field);
        $res = $this->export($fields);

        if(count($res) > 0) {
            foreach($res as $r) {
                $return[$r[$this->id_field]] = $r;
            }
        }
        return $return;
    }
```

This returns something like:

```
array(7) {
  [1]=>
  array(7) {
    ["id"]=> string(1) "1"
    ["name"]=>
    string(5) "Lager"
  }

  [4]=>
  array(7) {
    ["id"]=>
    string(1) "4"
    ["name"]=>
    string(9) "fdfdsfsdf"
  }
  [5]=>
  array(7) {
    ["id"]=>
    string(1) "5"
    ["name"]=>
    string(12) "sadsadasdasd"
  }
  [7]=>
  array(7) {
    ["id"]=>
    string(1) "7"
    ["name"]=>
    string(9) "sdfsdfsdf"
  }
```

So now the id can be used to access the exported data (1,4,5,7).
I know this code ain't great, but it does the job at the moment. However, I really think the data is much more valuable this way, as all data can be easily accessed if the ID is known.

Maybe an Agile Data function exportByID (or whatever name) could make sense?

### PR DESCRIPTION
**Pull request by DarkSide666**
* implement exportById(), fix #311
* cleanup

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1577,7 +1577,6 @@ class Model implements \ArrayAccess, \IteratorAggregate
                 }
 
                 $fields = array_keys($fields);
-
             } else {
 
                 // Add all model fields
@@ -1590,7 +1589,6 @@ class Model implements \ArrayAccess, \IteratorAggregate
                     }
                 }
             }
-
         }
 
         // add key_field to array if it's not there
--- a/src/Model.php
+++ b/src/Model.php
@@ -1533,50 +1533,83 @@ class Model implements \ArrayAccess, \IteratorAggregate
     /**
      * Export DataSet as array of hashes.
      *
-     * @param array|null $fields Names of fields to export
+     * @param array|null $fields    Names of fields to export
+     * @param string     $key_field Optional name of field which value we will use as array key
      *
      * @return array
      */
-    public function export($fields = null)
+    public function export($fields = null, $key_field = null)
     {
-        return $this->persistence->export($this, $fields);
-    }
-
-    /**
-     * Export DataSet as array of hashes. Uses id_field values as array keys.
-     *
-     * @param array|string $fields Names of fields to export
-     *
-     * @return array
-     */
-    public function exportById($fields)
-    {
-        if (is_string($fields)) {
-            $fields = [$fields];
+        // no key field - then just do export
+        if ($key_field === null) {
+            return $this->persistence->export($this, $fields);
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by romaninsh** [comment]
Question.

What if i want to have another field as a key, - not ID ?

**Comment by romaninsh** [comment]
``` php
$data_by_id = $model->exportDistinct();

$data_by_code = $model->exportDistinct($code);
```

This would also add "distinct" for the field you have specified.

romaninsh:
Another idea, perhaps we should add extra argument into export() and that's it?

romaninsh:
Playing devil's advocate. Trying out different ideas.

PhilippGrashoff:
imo an option to export would be best. And a proper implementation, mine is just a bad hack.

DarkSide666:
OK I will change this PR then.
* no need for exportDistinct() because it will always be distinct because array can not have 2 values with same key anyway.
* will add second key to export() method - $key_field_name. If it will be set, then export() will use that model field as a key for data array elements. If it's not set then do as it does now.

That sounds good?

romaninsh:
yes, thats sounds good. 

I still think distinct is important when column is specified.

``` php
$all_names = array_keys($clients->export(['name'], 'name'));
```

if you add distinct name, receive 100 records from db. If not - receive 1000 and deal with duplicates. Also when processing duplicates for name, don't overwrite existing values.

DarkSide666:
Romans in Skype: _forget distinct._ :)

Issue #308, opened 2018-03-21
Standard types overhaul


romaninsh:
The fields type is not sufficiently flexible as well it does not contain enough information and restrictions for the field yet. I think what "date" types started is pretty cool with the support of formats and all, but this can be extended to all the types.

After #307 is done and we can create type-specific field classes, we will be able to use individual properties more freely, so here are some of the initial ideas:

Field\Boolean:
 - contain yes/no values for the database (1/0)
 - contain yes/no values for the UI (yes/)

Field\List:
 - supporting enum, values and hard-coded lists
 - add ability to specify colors

Field\Date:
 - add fields for default UI formatting
 - support date, datetime and time types
 - support min/max dates

Field\Number:
 - specify how many decimal numbers
 - is negative permitted
 - formatting style (e.g. xx-xxxx)
 - generic prefix and postfix
 - support types 'money', 'percent', 'decimal', 'float', 'rating'
 - allow min/max

Field\Measurement:
 - support values that are expressed in units
 - duration: s, m, d, w
 - weight: g, kg
 - length: mm, cm, m, km
 - imperial units
 - set to (1cm) stored as (100).
 - automatically reformats for ui
 - types: 'duration', 'length'

# PLEASE SUGGEST MORE TYPES IN THE COMMENTS.

DarkSide666:
Field\Location - Google location (latitude, longitude).
Field\RichText - full HTML encoded text, maybe can more properties than just string type field.

Overall I like this approach, but you have to be very careful not to step to deep into UI (fields) related stuff here. Data should not care about UI at all. For that we only have one property for all fields `$ui` and UI class should take care about all visualization itself.

romaninsh:
Good suggestions, but typically any fields that rely on 3rd party libs should be in their add-ons. (due to dependencies)

FrancessFractal:
Hi! I'm totally new to the library, starting today to tinker with it to see if we want to use it, and was searching for how to do something and spotted this post... Seems closely related, so I'll just ask here and if you know how to do it you'll save me from having to hunt through the docs, and you also get an idea for a new type out of it :)

[...]

Issue #306, opened 2018-03-20
V2 Refactoring (Ideas)


romaninsh:
I'm creating this ticket to list some of the V2 refactoring ideas. 

While I probably won't be able to release V2 anytime in 2018, it may be needed at some point and the main idea would be to refactor some of the anti-patterns without loosing out on any of the functionality. 

I don't want anyone to discourage from adopting ATK Data just because at some point there could be V2. 

## What should happen before V2 is fully considered:

 - [ ] much larger user-base. I'm looking at the minimum of 1000 stars.
 - [ ] at least 3 third-party persistence implementations.
 - [ ] a working integration with top 3 PHP frameworks / apps with at least 50 stars each.
 - [ ] a working integration for RestAPI backend
 - [ ] a caching mechanism
 - [ ] refactor of https://github.com/atk4/dsql

Those conditions are an indication that there would be enough benefit from the effort it takes to create a V2.

## What things are to be considered during refactor

 - [ ] refactor all properties to camelCase
 - [ ] get rid of all classes with underscores (Persistence_SQL) in favour of namespaces
 - [ ] reduce number of lines inside Model through composition.
 - [ ] use PHP7.2 or higher as a minimum requirement
 - [ ] add type hinting for all methods and arguments
 - [ ] separate concepts of "Model" the set from "Model" the element (foreach ($model_set as $model_element)
 - [ ] make Field lighter my moving out some functionality
 - [ ] simplify Join class
 - [ ] persistence_sql - simplify, possibly moving out sql-related stuff

See also this review:

``` 
# atk4/data

+
Github issue tracking.
Github release management. 32 releases.
Good documendation and Comments.
Travic CI.
Covered with Tests.
Project consists of +10K lines of code, but half of them are tests. Pure core project consist of less than 10K lines of code.
-
Lots of null checks.
Some classes have more than +800 lines of code, and class Model is like God object
Some methods are also big, they need to be divided into smaller one.
Use of Getters and Setters.
Use of Traits in some classes.
The messages in git history are also not descriptive.
These are the example Classes where I mostly found these issues:
--Field.php
  Class is quite long. +400 lines.
  The function normilize has +130 lines.
  Use of Getters and Setters.
--Join.php
  Use of traits instead of Composition.
  Class is big, and have lots of methods.
--Model.php
  Use of traits instead of Composition.
  This class has +1800 lines of code. It is God object.
  The functions save and set have more than +120 lines of code.
--Persistence.php
  Use of traits instead of composition.
[...]

PR #314, opened 2018-03-23
Epic refactor for "Field" class

romaninsh:
Currently "Field" class is to heavy. I reckon it can be refactored without BC breaks and help with with the clean-ups originally intended for later - #306.

This Epic PR will implement #308 through individual smaller PRs:

 - [x] Cleanly get rid of Feld_SQL and add field factory.

The list may change as we go forward. Each separate PR will be discussed.

 - [x] Implement Field\Boolean #315 
 - [ ] Implement Field\Number #375 
 - [ ] Implement Field\List
 - [ ] Implement Field\Date
 - [ ] Implement Field\UUID
 - [ ] Implement Field\Location
 - [ ] Implement Field\Geometry
 - [ ] Implement Field\FullText
 - [ ] Implement containment (kinda containsMany / containsOne)
 - [ ] Implement Field\Measurement

Additionally for consideration:

 - [ ] Mechanism for field to verify persistence requirement and use different implementations
 - [ ] Take on handling of addCondition() `$m->addCondition('home_location', '<10km', $xy)`
 - [ ] Support individual features of fields, e.g. `$m->intersects('area_field', $xy);`


FrancessFractal:
Something you might add to 'for consideration' is data types made up of multiple fields. For instance, Postgres' point datatype might need two columns to be represented in MySQL. Alternatively, you could say that point needs to be broken into two fields...

romaninsh:
This is doable on a model level:

 1. declare super-field
 2. mark it non perist
 3. create 2-3 shadow fields.
 4. mark them system (hide)
 5. beforeSave copy value into shadow fields.
 6. afterLoad copy value back.

All that can be also defined inside a field class logic. I don't want to use it by default, however. Perhaps some for "Address/Location" field to store geo coordinates.

romaninsh:
Another thing to add for consideration is if fields should care about their "visibility" for the UI. Currently there is no good way for ATK UI to implement this, but I don't see how this is even relevant in "ATK Data" domain: 

https://github.com/atk4/data/blob/develop/src/Field.php#L418

https://forum.agiletoolkit.org/t/visible-hidden-editable-flags-do-nothing/465

DarkSide666:
I think it's data domain where we can set  if field is visible, editable, hidden, system. UI should take these switches into account not only in CRUD component, but all components, including single form fields too.

Of course you can always overwrite that logic in form/field, but model is perfect place for _default_ definition of visibility/editability.

PR #315, opened 2018-03-23
Implement Field\Boolean

romaninsh:
Implements types:

 - boolean

Example:

``` php
$model->addField('is_vip', ['type'=>'boolean']);

$model->addField('is_vip', ['type'=>'boolean', 'enum'=>['No', 'Yes']);
```

### DIFF
--- a/src/Field/Boolean.php
+++ b/src/Field/Boolean.php
@@ -8,7 +8,7 @@ use atk4\core\InitializerTrait;
 use atk4\data\ValidationException;
 
 /**
- * Your favourite nullable binary type
+ * Your favourite nullable binary type.
  */
 class Boolean extends \atk4\data\Field
 {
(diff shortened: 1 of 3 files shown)


src/Field/Boolean.php
```diff
@@ -0,0 +1,61 @@
...
+            return (bool) $value;
+        }
+
+        throw new ValidationException([$this->name => 'Must be a boolean value']);
+    }
+
+    // TODO: REVIEW
+    public function validate($value)
+    {
+        if ($this->required && empty($value)) {
```
DarkSide666: Do `required` for boolean means that it should always be === valueTrue?

romaninsh: correct.

romaninsh:
I think test-suite is lacking..

PR #330, opened 2018-07-03
Significantly improve flexibility of aggregates

romaninsh:
Implementation of better aggregates. New features:

 - 'aggregate'=>'count' will default 'field' to '*', no need to specify through Expression anymore
 - 'aggregate'=>'group_concat' won't replace empty string with 0
 - 'expr'=>'sum([field])' can be used instead of aggregate
 - 'args' can be passed if you are using 'expr'
 - 'aggregate' can now also be expression ('field' will be unused)

Documentation and test-suite updated.

### DIFF
--- a/docs/references.rst
+++ b/docs/references.rst
@@ -184,12 +184,59 @@ keep making your query bigger and bigger::
             ['total_gross', 'aggregate'=>'sum'],
         ]);
 
-.. important::
-    Imported fields will preserve format of the field the reference. In the example,
-    if 'Invoice_line' field total_vat has type `money` then it will also be used
-    for a sum. Aggregate fields are always declared read-only, and if you try to
-    change them, you will receive exception.
+Imported fields will preserve format of the field they reference. In the example,
+if 'Invoice_line' field total_vat has type `money` then it will also be used
+for a sum. 
 
+You can also specify a type yourself::
+
+    ->addField('paid_amount', ['aggregate'=>'sum', 'field'=>'amount', 'type'=>'money']);
+
+Aggregate fields are always declared read-only, and if you try to
+change them (`$m['paid_amount'] = 123;`), you will receive exception.
+
+Available Aggregation Functions
+-------------------------------
+
+The mathematical aggregates such as sum, avg, min, max and count will automatically
+default to 0 if no respective rows were provided. The default SQL behaviour is to
+return NULL, but this does go well with the cascading formulas::
+
+    coalesce(sum([field]), 0);
+
+For other functions, such as 'group_concat' no coalesce will be used, so the empty
+string will be the result. When you specify `'aggregate'=>'count'` field defaults
+to '*'.
+
+Aggregate Expressions
+---------------------
+
+Sometimes you want to use a more complex formula, and you may do so by specifying
+expression into 'aggregate'::
+
+    ->addField('len', ['expr' => 'sum(length([name]))']),
+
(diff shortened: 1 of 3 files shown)


romaninsh:
looks like codeclimate/coverage is broken.

docs/references.rst
```diff
@@ -184,12 +184,59 @@ keep making your query bigger and bigger::
...
+    ->addField('paid_amount', ['aggregate'=>'sum', 'field'=>'amount', 'type'=>'money']);
+
+Aggregate fields are always declared read-only, and if you try to
+change them (`$m['paid_amount'] = 123;`), you will receive exception.
+
+Available Aggregation Functions
+-------------------------------
+
+The mathematical aggregates such as sum, avg, min, max and count will automatically
+default to 0 if no respective rows were provided. The default SQL behaviour is to
```
DarkSide666: Not sure if this is good or bad to use `coalesce` for all of these.
`count` - surely should be 0 not NULL
`sum` - probably also 0 is more meaningful than NULL
`min`, `max` and `avg`- 0 will give wrong assumption that actual value is 0. Not so sure if this is ok. Sometimes 0 is ok here, but i think there can be use-cases where we would like to see NULL to differ actual value (like 0) from no value at all (like NULL).

docs/references.rst
```diff
@@ -184,12 +184,59 @@ keep making your query bigger and bigger::
...
+        ->addField('page_list', [
+            'aggregate'=>$book->refModel('Pages')->expr('group_concat([number], [])', ['-'])
+        ]);
+
+
+or 'field'::
+
+    ->addField('paid_amount', ['aggregate'=>'count', 'field'=>new \atk4\dsql\Expression('*')]);
 
+.. note:: as of 1.3.4 count's field defaults to `*` - no need to specify explicitly.
```
DarkSide666: i love that

src/Reference_Many.php
```diff
@@ -86,22 +86,46 @@ public function refLink($defaults = [])
...
+                ), 'alias'=>$field]);
+            };
+            unset($defaults['args']);
+        } elseif (is_object($defaults['aggregate'])) {
+            $cb = function () use ($defaults, $field) {
+                return $this->refLink()->action('field', [$defaults['aggregate'], 'alias'=>$field]);
+            };
+        } elseif ($defaults['aggregate'] == 'count' && !isset($defaults['field'])) {
+            $cb = function () use ($defaults, $field) {
+                return $this->refLink()->action('count', ['alias'=>$field]);
```
DarkSide666: I don't see where it sets `'field' => new Expression('*')` here.
Also no test coverage for this case (fixed).
And it doesn't use `fx0` as it does few lines below. Maybe that's not needed in case of `count`, because I think in SQL count(whatever) always returns number no matter if it's 0 or it selects `*` or some nullable field.

docs/references.rst
romaninsh: ok, i agree, i'll switch the other methods to use 'fx' not 'fx0'

src/Reference_Many.php
romaninsh: action('count') does that. i think it was in the test.. coverage broken?

DarkSide666: ok

docs/references.rst
DarkSide666: ok

PR #350, opened 2018-10-22
Implement Lookups for inserts

romaninsh:
This has been an undocumented feature (perhaps partially working), but with this PR we make it a fully-functional and fully-documented one.

Adding support for "Lookup during insert". Quite often you wish to ingest data into your model like this:

``` csv
Name,Country_code
John,GB
Janis,LV
```

To do that, you normally would need to fetch all the countries into a cache, replace "Country_code" with "country_id" then ingest the data. But in ATK Data we love to make database sweat for us. And in case of SQL, we can actually use SQL expressions:

```
insert into user (name,country_id) values ("John", select id from country where code="GB");
insert into user (name,country_id) values ("Janis", select id from country where code="LV");
```

Few things to consider:

 - [x] Ingest by specifying "title" column
 - [ ] Ingest by specifying lookup added with addField
 - [ ] Ingest even if lookup table is not unique (add limit 1)
 - [ ] Ingest even if lookup is invalid, by setting null
 - [ ] support insert() and import()
 - [ ] If related model is non-sql (or in different persistence), fail with correct exception (for now)

### Follow-ups

After that, there are some additional features for consideration and for consistency:

 - [ ] Specifying array of IDs to hasMany field should populate the relations `"John", [3,9]`
 - [ ] Specifying array of values to hasMany lookup field should populate relations `"John", ['green', 'red']`


romaninsh:
Some consideration for importing data. In situation where we add condition 'is_admin' = false.

Allow 3 modes of import.
 - records that don't match condition will have their conflicting fields ignored. (trying to set is_admin=true will reset it to false)
 - records that don't match condition will generate exception and entire import will fail.
 - records that don't match condition will be skipped. (although it would be nice to receive list of skipped records)

romaninsh:
Another consideration is to have support for storing JSON values natively in mariaDB:

https://mariadb.com/resources/blog/json-with-mariadb-10-2/

src/Reference_SQL_One.php
```diff
@@ -9,6 +9,28 @@
...
+    {
+        parent::init();
+
+        // Will try to execute last
+        $this->owner->addHook('beforeSave', $this, null, 20);
+    }
+
+    public function beforeSave($m)
+    {
+        $field = $this->link;
```
DarkSide666: Why below you somewhere use $field, somewhere $this->link ? Need consistency.

src/Field_SQL_Expression.php
```diff
@@ -34,6 +34,11 @@ class Field_SQL_Expression extends Field_SQL
      */
     public $aggregate = null;
 
+    /**
+     * Aggregation by concatination.
+     */
+    public $concat = null;
```
DarkSide666: where this property is used?

DarkSide666: ok then. somewhere it is used :)

tests/LookupSQLTest.php
```diff
@@ -0,0 +1,314 @@
...
+
+    public function testImportInternationalUsers()
+    {
+        $c = new LCountry($this->db);
+
+        // Specifying hasMany here will perform input
+        $c->insert(['name'=>'Canada', 'Users'=>['Alain', ['Duncan', 'is_vip'=>true]]]);
+
+        // Both lines will work quite similar
+        $c->insert(['Latvia', 'user_names'=>'imants,juris']);
```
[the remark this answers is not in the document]
romaninsh: ok, now addressed.

DarkSide666:
Now looks good

PR #352, opened 2018-10-22
Getters for protected our_field and their_field

PhilippGrashoff:
Makes something like this possible:
Model User
```
$this->hasMany('Email', [new Email(), 'their_field' => 'user_id');
```

```
$new_user = $user->newInstance();
//copy all emails
foreach($user->ref('Email') as $e) {
    $new_e = new Email();
    $new_e->set($e->ref('Email')->getTheirField(), $e->get('email');
    $new_e->save();
}
```

This seems related to the deep copy feature currently being inplemented.

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -179,6 +179,20 @@ class Reference
         return $this->getModel($defaults);
     }
 
+    /**
+     * Getter for protected property our_field.
+     */
+    public function getOurField() {
+        return $this->our_field;
+    }
+
+    /**
+     * Getter for protected property their_field.
+     */
+    public function getTheirField() {
+        return $this->their_field;
+    }
+    
     // {{{ Debug Methods
 
     /**


romaninsh:
We are not big fans of getters/setters without a good reason. We rather make properties public and make it as internal in the heredoc. 

In regards to deepCopy, I don't see why it would need access to 'their_field'. It's used for building conditions when traversing references. I'd have to make some use-cases before i have a better idea.

PhilippGrashoff:
Well, if theres no contras on having these properties public, tha would be the easiest thing :)

romaninsh:
@PhilippGrashoff lets keep them as they are until there is real need / issue.

PhilippGrashoff:
Ok, will work around that

funyx:
+1 for public `our_field` and `their_field`

use case:
1. app\Model extends atk4\data\Model
2. app\api\Model extends app\Model
3. app\Model\Post extends app\Model 
- fields, validation, relations, hooks
4. app\api\Model\Post extends app\api\Model
- that model should have all fields and relations used in app\Model\Post (it should use app\Model\Post as a "seed")

solution:
- app\api\Model::mirrorModel
```php
private function mirrorModel()
    {
        $m = $this->factory($this->appModelSeed, [$this->app->db]);
        $this->table = $m->table;
        $m->addMethod('initMirrorModel', function () use ($m) {
            $refMap = $this->refMap ?: [];
            $refSeed = $this->appModelSeedRefs ?: [];
            $refs = array_unique(array_merge($refSeed, array_keys($refMap)));
            foreach ($refs as $ref) {
                // check if it exists in the seed model
                if ($m->hasRef($ref)) {
                    // if is in ref map
                    if (in_array($ref, $refMap)) {
                        $this->addRef($refMap[$ref], $m->getRef($ref)->model);
                    } else {
                        $refProps = (Array) $m->getRef($ref);
                        $defaults = [];
                        $defaults[0] = $ref;
                        $defaults['model'] = $refProps['model'];
                        if ($refProps['*our_field']) {
                            $defaults['our_field'] = $refProps['*our_field'];
                        }
                        if ($refProps['*their_field']) {
                            $defaults['their_field'] = $refProps['*their_field'];
                        }
                        $this->add($this->factory(get_class($m->getRef($ref)), $defaults));
                    }
                }
            }
            foreach ($m->elements as $fieldName => $field) {
                // add fields (not controllers and not foreign key)
               /**copy all fields**/
            }
        });
        $m->initMirrorModel();
    }
```

- app\api\Model\Post
```php

namespace app\api\Model;

use \app\api\Model;

class Post extends Model
{
    // model to use as seed
    public $appModelSeed = '\app\Model\Post';

    // refs to use from "seed" model
    public $appModelSeedRefs = [
      'user_id'
    ];

    public function init()
    {
        parent::init();
        // api related logic
        // - additional fields
        // - additional relations
        // - setup api validation
        // - setup api hooks
    }
}
```

PS: there might be a better solution to "copy" a Model, but that's what i came up with..

PR #353, opened 2018-10-25
Feature/line end normalize

DarkSide666:
Normalize `type='string'` field value to one-liner (no line breaks).


DarkSide666:
Now +1.5% coverage :)

PR #366, opened 2018-12-17
Save id value after insert to id_field

### LINKED ISSUE #365: Model: $this->set($this->id_field, $this->id) after insert
Hi,

will create a pull request for this right after I published this here.
In atk4\data\Model::save(), after inserting a record to the database the id given from persistence is stored in $this->id, but its not populated to the field of the model.
So, if reload_after_save is set to false or if some other action should be done with the model inside the afterInsert hook, Model->get('id') will return null though this value is available and stored in db.
setting it by $this->set($this->id_field, $this->id) will make Model->get('id') return the correct value always.

PhilippGrashoff:
fixes #365. Now, Model->get('id') will always return the id (which is known at that point, just only set to $this->id until now), in the afterInsert hook as well as anywhere after if reload_after_save is set to false.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1463,6 +1463,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
 
                     $this->dirty = [];
                 } elseif ($this->id) {
+                    $this->set($this->id_field, $this->id);
                     $this->hook('afterInsert', [$this->id]);
 
                     if ($this->reload_after_save !== false) {


romaninsh:
in some situations ID won't be a field at all. With SQL you normally get ID field but when you work with array persistences - ID is array index.. will have to review this further.

romaninsh:
@PhilippGrashoff thanks, i've resubmitted PR in the other branch: https://github.com/atk4/data/pull/368

PR #367, opened 2018-12-17
Feature/extend setorder even more

DarkSide666:
This PR extends https://github.com/atk4/data/pull/364 even more.

* adds test cases
* adds support for `setOrder(['net'=>true, 'vat'=>false, 'gross'])` and adds test cases for them.


DarkSide666:
Added docs here https://github.com/atk4/data/commit/4117bc7bcd6c66540470cd3a4243d3d376681dfb

PR #368, opened 2018-12-18
Save id value after insert to id_field

### LINKED ISSUE #365: Model: $this->set($this->id_field, $this->id) after insert

romaninsh:
fixes #365. Now, Model->get('id') will always return the id (which is known at that point, just only set to $this->id until now), in the afterInsert hook as well as anywhere after if reload_after_save is set to false.

Romans: added test-case

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -252,7 +252,7 @@ class Field
             // other field types empty value is the same as no-value, nothing or null
             if ($f->type && $f->type != 'string' && $value === '') {
                 if ($this->required && empty($value)) {
-                    throw new ValidationException([$this->name => 'Must not be a empty']);
+                    throw new ValidationException([$this->name => 'Must not be empty']);
                 }
 
                 return;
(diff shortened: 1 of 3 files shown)


PhilippGrashoff:
Thanks for the test case. Was in a hurry yesterday as we went on holidays today, so didnt have time for that. For future PRs I will provide test cases.

DarkSide666:
See my comment here: https://github.com/atk4/data/issues/365#issuecomment-448731314

PR #372, opened 2019-01-04
Fix Join->hasOne/hasMany

### LINKED ISSUE #369: hasMany on Join in not implemented yet
https://github.com/atk4/data/blob/749e43708f03e28e2bf251baf8b8a0142eb9fea4/src/Join.php#L352

Join does not have table property 
and Join is not extending any other class

DarkSide666:
* add test for Join->hasOne/hasMany (also improves coverage a bit)
* fix Join->hasOne/hasMany methods
* fully fix #369 issue

### DIFF
--- a/src/Join.php
+++ b/src/Join.php
@@ -321,38 +321,42 @@ class Join
     /**
      * Creates reference based on a field from the join.
      *
-     * @param Model $model
-     * @param array $defaults
+     * @param string $link
+     * @param array  $defaults
      *
      * @return Reference_One
      */
-    public function hasOne($model, $defaults = [])
+    public function hasOne($link, $defaults = [])
     {
         if (!is_array($defaults)) {
-            $defaults = ['model' => $defaults];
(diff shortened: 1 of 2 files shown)


src/Join.php
```diff
@@ -321,38 +321,42 @@ public function weakJoin($defaults = [])
...
+    public function hasMany($link, $defaults = [])
     {
+        if (!is_array($defaults)) {
+            $defaults = ['model' => $defaults ?: 'Model_'.$link];
+        }
+
         $defaults = array_merge([
             'our_field'   => $this->id_field,
-            'their_field' => $this->table.'_'.$this->id_field,
+            'their_field' => $this->owner->table.'_'.$this->id_field,
```
romaninsh: how was this working before? what about recursive joins. ?

DarkSide666: Most likely it was not working before, because $this (join) ->table property didn't exist, so ... it should throw exception anyway.
Can you add test case for recursive joins? What do you mean? Something like `$this->join($this->table, 'parent_id')` ?

PR #375, opened 2019-01-07
Add support for more fields

romaninsh:
implement types using DBAL first - https://github.com/atk4/data/pull/727

Implements separate classes for Number, Percent and Rating. Also includes Money for a basic implementation.


src/Field/Integer.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+class Integer extends Numeric
+{
+    /** @var string Field type for backward compatibility. */
+    public $type = 'integer';
+
+    /**
+     * Specify how many decimal numbers should be saved.
+     */
+    public $decimalNumbers = 0;
```
romaninsh: we commonly use $decimal_number, no caps.. i think.

src/Field/Integer.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Enable number rounding. If true will round number, otherwise will round it down (trim).
+     */
+    public $enableRounding = false;
```
romaninsh: already defined in a parent.

src/Persistence/Array_.php
```diff
@@ -56,7 +56,10 @@ public function add($m, $defaults = [])
 
         if ($f = $m->hasField($m->id_field)) {
             if (!$f->type) {
-                $f->type = 'integer';
+                //$f->type = 'integer';
+                // have to re-add field to use proper Field class
+                $m->removeField($m->id_field);
+                $m->addField($m->id_field, ['type'=>'integer']);
```
romaninsh, line 64: probably could be decided earlier, before field is added.

romaninsh (commented):
WIP..

src/Persistence/Array_.php
DarkSide666: Not really because ID field is added in Model automatically even before Model is added into Persistence.

src/Field/Integer.php
DarkSide666: addressed

src/Field/Integer.php
DarkSide666: renamed to enable_rounding and it's defined here because it changes default value (to false) for integer.

src/Field.php
```diff
@@ -269,19 +275,9 @@ public function normalize($value)
...
-            break;
+            throw new Exception(['Use Field\ShortText for type=string', 'this'=>$this]);
         case 'text':
-            // normalize line-ends to LF and trim
-            $value = trim(str_replace(["\r\n", "\r"], "\n", $value));
-            if ($this->required && empty($value)) {
-                throw new ValidationException([$this->name => 'Must not be empty']);
-            }
-            break;
+            throw new Exception(['Use Field\Text for type=text', 'this'=>$this]);
```
romaninsh: didnt' we agree on the meeting to keep backwards compatibility ?

if we accept that, it will break code in a very "python" way (no steady update path)

src/Field/ShortText.php
```diff
@@ -0,0 +1,42 @@
...
+
+use atk4\data\Field;
+use atk4\data\ValidationException;
+
+/**
+ * Basic string field type. Think of it as field type "string" in past.
+ *
+ * Most of the time this is most basic type you will use.
+ */
+class ShortText extends Text
```
romaninsh: I suggest we have two classes:

Line (may be a short or a long line, but enter is not permitted)
Text (may be a short or a long text, will contain some paragraphs)

src/Field.php
```diff
@@ -235,195 +217,79 @@ public function __construct($defaults = [])
...
-                return;
-            }
+        $f = $this;
 
-            // validate scalar values
-            if (in_array($f->type, ['string', 'text', 'integer', 'money', 'float']) && !is_scalar($value)) {
-                throw new ValidationException([$this->name => 'Must use scalar value']);
+        // only string type fields can use empty string as legit value, for all
+        // other field types empty value is the same as no-value, nothing or null
+        if ($f->type && $f->type != 'string' && $value === '') {
```
romaninsh: probably incorrect.......

src/Field.php
```diff
@@ -235,195 +217,79 @@ public function __construct($defaults = [])
      */
     public function normalize($value)
     {
-        try {
-            if (!$this->owner->strict_types) {
-                return $value;
+        if ($value === null) {
+            if ($this->required) {
```
romaninsh, line 276: rename into not_null

src/Field.php
```diff
@@ -235,195 +217,79 @@ public function __construct($defaults = [])
...
-                break;
-            }
+            return;
+        }
 
-            return $value;
-        } catch (Exception $e) {
-            $e->addMoreInfo('field', $this);
+        // validate scalar values
+        if (in_array($f->type, ['string', 'text', 'integer', 'money', 'float']) && !is_scalar($value)) {
```
romaninsh: we don't care about type, remove.

src/Field.php
```diff
@@ -235,195 +217,79 @@ public function __construct($defaults = [])
...
+            throw new ValidationException([$this->name => 'Must use scalar value']);
+        }
 
-            throw $e;
+        // normalize
+        // @TODO remove this block in future - it's useless
+        switch ($f->type) {
```
romaninsh: switch not needed. remove.

src/Field.php
```diff
@@ -475,49 +341,66 @@ public function useAlias()
...
+    }
+
+    /**
+     * Returns true if field allows NULL values.
+     *
+     * @return bool
+     */
+    public function canBeNull() : bool
+    {
+        return $this->mandatory === false;
```
romaninsh, line 446: return !$this->not_null

src/Field.php
```diff
@@ -475,49 +341,66 @@ public function useAlias()
...
+    }
+
+    /**
+     * Returns true if field allows EMPTY values like empty string and NULL.
+     *
+     * @return bool
+     */
+    public function canBeEmpty() : bool
+    {
+        return $this->mandatory === false && $this->required === false;
```
romaninsh, line 456: return !$this->['ui']['required'] ?? true;

src/Field/Array_.php
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @throws ValidationException
+     *
+     * @return mixed
+     */
+    public function normalize($value)
+    {
+        if ($value === null || $value === '') {
+            if ($this->required) {
+                throw new ValidationException([$this->name => 'Must not be null or empty']);
```
romaninsh, line 31: shouldn't we call parent?

src/Field.php
```diff
@@ -222,6 +202,8 @@ public function __construct($defaults = [])
     }
```
romaninsh, line 248: set($this->default_value)

src/Field/Boolean.php
```diff
@@ -55,36 +58,42 @@ public function init()
...
-        if ($value === $this->valueFalse) {
-            return false;
+        if (!is_bool($value)) {
+            throw new ValidationException([$this->name => 'Must be a boolean value']);
         }
 
-        if (is_numeric($value)) {
-            return (bool) $value;
+        // if value required, then only valueTrue is allowed
+        if ($this->required && $value !== true) {
```
romaninsh: technically should be now job of checkbox, but we can leave it for now..

src/Field/DateTime.php
```diff
@@ -0,0 +1,105 @@
...
+    public $type = 'datetime';
+
+    /**
+     * Array with Persistence settings like format, timezone etc.
+     * It's job of Persistence to take these settings into account if needed.
+     *
+     * @var array
+     */
+    public $persistence = [
+        'format'   => null, // for date it can be 'Y-m-d', for datetime - 'Y-m-d H:i:s' etc.
```
romaninsh, line 25: why not use Y-m-d as default ?

src/Field/DateTime.php
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @throws ValidationException
+     *
+     * @return mixed
+     */
+    public function normalize($value)
+    {
+        if ($value === null) {
+            if ($this->mandatory || $this->required) {
+                throw new ValidationException([$this->name => 'Must not be null']);
```
romaninsh, line 65: not_null

src/Field/Money.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+class Money extends Numeric
+{
+    /** @var string Field type for backward compatibility. */
+    public $type = 'money';
+
+    /**
+     * @var int Specify how many decimal numbers should be saved.
+     */
+    public $decimal_numbers = 4;
```
romaninsh: normally that's 2

src/Field/Numeric.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Round up to the nearest number.
+     *
+     * @param float $n Number
+     * @param int   $p Precision
+     *
+     * @return float
+     */
+    protected function round_up(float $n, int $p) : float
```
romaninsh: roundUp

src/Field/Numeric.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Round down to the nearest number.
+     *
+     * @param float $n Number
+     * @param int   $p Precision
+     *
+     * @return float
+     */
+    protected function round_down(float $n, int $p) : float
```
romaninsh: roundDown

src/Field/Object_.php
```diff
@@ -0,0 +1,67 @@
...
+     * @param mixed $value
+     *
+     * @throws ValidationException
+     *
+     * @return mixed
+     */
+    public function normalize($value)
+    {
+        if ($value === null || $value === '') {
+            if ($this->required) {
```
romaninsh, line 30: parent call? not_null..

src/Field/Object_.php
```diff
@@ -0,0 +1,67 @@
...
+                throw new ValidationException([$this->name => 'Must not be null or empty']);
```
romaninsh, line 31: how can object be empty ?

src/Field/Object_.php
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            return;
+        }
+
+        if (is_string($value) && $this->owner && $this->owner->persistence) {
+            $value = $this->owner->persistence->jsonDecode($this, $value, false);
```
romaninsh: is that needed ?

src/Field/Object_.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Casts field value to string.
+     *
+     * @param mixed $value Optional value
+     *
+     * @return string
+     */
+    public function toString($value = null) : ?string
```
romaninsh: use native serialization

src/Field/_Percent.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+namespace atk4\data\Field;
+
+class _Percent extends Number
+{
+    /**
+     * @var int IMPORTANT: set to 2+precision, since percentage is stored as a 0 .. 1
+     */
+    public $decimalNumbers = 2;
```
romaninsh: decimal_numbers

src/Model.php
```diff
@@ -482,8 +482,19 @@ public function fieldFactory($seed = [])
         return $field;
     }
 
+    /** @var array [type => classname] */
     protected $typeToFieldSeed = [
-        'boolean' => ['Boolean'],
+        'boolean'  => ['Boolean'],
+        'float'    => ['Numeric'],
```
romaninsh: 'float' => ['LegacyField', 'type'=>'float'] 

etc

src/Persistence/Array_.php
romaninsh: ok, lets have id field seed then, which is by default id_field_seed = ['Integer'] but could also be something else.

DarkSide666: Added id field seed property in Model class and removed $type property dependency from Array Persistence class.

src/Field/ShortText.php
DarkSide666: Addressed.
Simply renaming ShortText class to Line does this :)
No more need for _short_ vs _long_. Limit can be defined by max_length property in both of these classes.

src/Field/Money.php
DarkSide666: it's 4 currently so I thought we should have the same (for BC), but normally it should be 2 in most cases.

src/Field/Numeric.php
DarkSide666: addressed

src/Field/Numeric.php
DarkSide666: addressed

src/Field/Money.php
DarkSide666: addressed

src/Field/Object_.php
DarkSide666: addressed

src/Field/_Percent.php
DarkSide666: addressed

src/Field/Object_.php
DarkSide666: why? jsonEncode in persistence class is native, but _better_ method because it treats exception cases better.

src/Field/Array_.php
DarkSide666: I am avoiding calling parent normalize because was trying to get rid of it at first and then see if a lot of field classes will actually duplicate same code over and over again, then move it back to parent class normalize method.

src/Field.php
DarkSide666: what exactly?

src/Field.php
DarkSide666: It should be BC so far. Derived field classes don't call parent::normalize() so this switch is not used at all. Theoretically all this normalize method in Field class can be removed, but not 100% yet. That's why I left it there for cases when Field->type === null (some custom fields).

src/Field.php
abbadon1334: @romaninsh '$this->['ui']['required']' <--- i lost some meeting but why check for UI required in place of required ?

src/Field/Object_.php
abbadon1334: not_null will be another extra property we already have mandatory/system/required that are still hard to understand at first, i suggest to hardcode in a parent call or in a specific protected method like : 

```php
protected function notNull($value) : void {
    if ((null === $value || $value === '') && $this->required) {
        throw new ValidationException([$this->name => 'Must not be null or empty']);
    }
}
```

abbadon1334 (changes requested):
LGTM, i agree to hardcode in some way the notNull exception

mvorisek:
A lot of changes are already integrated, different fields itself should be fixed in #727, closing this one.

PR #377, opened 2019-01-09
implement Array persistence traversal, conditions and actions

romaninsh:
Array persistence currently does not support even basic conditions or actions. I have implemented those and now array persistence models can also be traversed.

``` php
$user = new User($array_persistence);
$user->addCondition('name', 'foo');
echo $user->action('count')->getOne();  // 123
```

 - [ ] TODO - look for inconsistency in documentation
 - [x] TODO - check if Grid now works ok with paginator

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ phpunit-mysql-local.xml
 /vendor
 .DS_Store
 /.idea
+atk-test*.csv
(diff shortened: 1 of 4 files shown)


DarkSide666:
Checking Grid/Paginator:
- [ ] Model->limit is not working

PR #384, opened 2019-01-31
Feature/add action in aftersave hook

### LINKED ISSUE #383: Issue with afterSave, AfterInsert and AfterUpdate Hooks together with hasOne()->addField()
Hi There,

lets hope I can describe this so you understand. I will stick to my real life case.
1) After my model is saved, I want to perform some action. I want to always do this on insert, on update only if certain fields are changed. As I didnt find a way to determine in afterSave hook if its an insert or update operation, the only viable option is to call the action from afterInsert and from afterUpdate hook,

2) The Model has some fields from a hasOne relationship. Like Model->hasOne('tour_type_id')->addFields(['tour_type_name' => 'name', 'activity_id' => 'activity_id']);

3) The action I perform uses one of these fields set by hasOne()->addFields(), and here is the issue:
When I call the action from afterInsert hook, these fields are not populated yet as the reload after save is executed after the afterInsert hook. So my action fails when called from afterInsert hook.

So there is 2 sensible solutions for this situation of which I like the first way more
1) Find a way to determine in afterSave hook if it was an update/insert operation
2) change the order in atk4\data\Model that the afterInsert hook is executed after the reload after save. This could break other stuff as it means a change of logic.

DarkSide666:
Adds `is_update` parameter in `afterSave` hook so we can now know was it insert or update.
fix #383

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -712,7 +712,7 @@ section. Add this into your Invoice Model::
 Next both payment and lines need to be added after invoice is actually created,
 so::
 
-    $this->addHook('afterSave', function($m){
+    $this->addHook('afterSave', function($m, $is_update){
         if(isset($m['payment'])) {
             $m->ref('Payment')->insert($m['payment']);
         }
--- a/docs/hooks.rst
+++ b/docs/hooks.rst
@@ -122,7 +122,7 @@ A good place to hook is beforeSave as it will be fired when adding new records
 or modifying existing ones:
 
  - beforeSave($m) (saving existing or new records. Not executed if model is not dirty)
- - afterSave($m) (same as above)
+ - afterSave($m, $is_update) (same as above, $is_update is boolean true if it was update and false otherwise)
 
 You might consider "save" to be a higher level hook, as beforeSave is called
 pretty early on during saving the record and afterSave is called at the very end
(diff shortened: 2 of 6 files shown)


PhilippGrashoff:
Hi,

regarding the added docs: in afterUpdate Hook, the model is already reloaded. its done by $to_persistence->update(). So in afterUpdate the model is reloaded, in afterInsert not.

PhilippGrashoff:
And thanks for this PR, will make coding easier :)

DarkSide666:
will fix that in docs. thanks for noticing!

PhilippGrashoff:
Would be a future thing to have afterUpdate and afterInsert hooks the same way :) Now, one has the reloaded model, the other not.

DarkSide666:
Well, that will break backward compatibility because now afterUpdate hook gets data array as parameter, but afterInsert hook gets new id as parameter.

PhilippGrashoff:
Yeah I know, this could be something for V2, to have all hooks behave the same.

DarkSide666:
Rule of thumb I am trying to stick to is to not use before/after Insert/Update hooks but use only before/after Save hooks :)

PhilippGrashoff:
Which is now always possible thanks to this PR ;)

PR #389, opened 2019-02-23
Fix normalization of type Time

abbadon1334:
i don't know if the new refactoring of fields #314 worth correct this bug, but in Field->normalize the type **time** is normalized like this : 
```php
if ($f->type == 'time') {
    // remove date portion from date type value
    $value->setDate(0, 1, 1);
}
```
value will be normalized like this : 0000-01-01 11:45:00
but value returned by model is : 1970-01-01 11:45:00

if mysql field is of time it will return 1970-01-01, this will trigger a save because is dirty and consequently a useless audit insert.

can be set to 1970 to uniform all drivers? due to the fact that date value are useless in Time definition

To test the issue try the code below : 
```php
class ModelTestTime extends \atk4\data\Model {
    
    function init()
    {
        parent::init();
        
        $this->addField('signin_open_time', [
            'type'    => 'time',
            'default' => new \DateTime(),
        ]);
    }
}

        $model = new ModelTestTime();
        $model->tryLoad(1);
        
        \Kint\Kint::$enabled_mode = 'r';
        $this->add($loader = new Loader());
    
        $loader->set(function($p) use ($model,$loader) {
    
            $sent_value = $this->app->stickyGet('sent_value');
    
            $p->add(new Text())
              ->addParagraph('URI: ' . $_SERVER['REQUEST_URI'])
              ->addParagraph('Stored value : ' . $model->get('signin_open_time')
                                                       ->format('Y-m-d H:i:s'))
              ->addParagraph('Saved value : ' . (empty($sent_value) || is_null($sent_value) ? ' not sent yet  '
                                 : $sent_value));
    
    
            $p->add($form = new Form());
            $form->setModel($model, ['signin_open_time']);
            $form->onSubmit(function ($f) use ($loader) {
        
                $f->model->save();
        
                return new \atk4\ui\jsReload($loader, [
                    'sent_value' => $f->model->get('signin_open_time')
                                             ->format('Y-m-d H:i:s')
                ]);
            });
        });
```


PhilippGrashoff:
strange I never noticed that yet, I work a lot with time fields and checks if its value is dirty...

abbadon1334:
> strange I never noticed that yet, I work a lot with time fields and checks if its value is dirty...

can be an issue with mysql config and NO_ZERO_DATE but i don't think so, did you test the code?

it will do it unnoticed, because it will not change anything in stored data, but it will trigger an update in model and an insert in audit.

i saw it while i was working on catching errors PR, because the sum of queries was strange when i save the form without changes

DarkSide666:
Yes there is such issue with time type. Not sure, maybe on Linux you can set date as 0000-01-01 or maybe it's some kind of mysql setting, but still...

Overall I would like to say that I have never really understood this field type `time` as such. Time is actually not related to date or DateTime object. I think that `time` should actually be stored as integer `(H*24+M)*60+S` or as text `H:M:D` or any other format, but not DateTime.

romaninsh:
that "demo" you mentioned has too much extra code. It would be nice to have it as a test-case.

PhilippGrashoff:
in my linux test env, 1970-01-01 works fine.

abbadon1334:
The error comes from the database when field type and database table are set to type = time

```php

class testModel extends \atk4\data\Model {
    
    public $table = 'test';
    
    public function init()
    {
        parent::init();
    
        $this->addField('t_dt', ['type' => 'time']);
    }
}

$p = new \atk4\data\Persistence_SQL('sqlite:/tmp/foo1.db');

$m = new testModel($p);
$m['t_dt'] = '18:00:00';

echo '=== beforeSave' . PHP_EOL;

echo "is Dirty : " . ($m->isDirty(['t_dt']) ? 'true' : 'false') . PHP_EOL;
echo '$m["t_dt"] returns => ' . $m["t_dt"]->format('Y-m-d H:i:s'). PHP_EOL;
echo '$m->data["t_dt"] returns => ' . $m->data["t_dt"]->format('Y-m-d H:i:s') . PHP_EOL;
echo '$m->dirty["t_dt"] returns => ' . (!isset($m->dirty["t_dt"]) ? 'null' : $m->dirty["t_dt"]->format('Y-m-d H:i:s')) . PHP_EOL;

$m->save();

echo '=== afterSave' . PHP_EOL;

echo "is Dirty : " . ($m->isDirty(['t_dt']) ? 'true' : 'false') . PHP_EOL;
echo '$m["t_dt"] returns => ' . $m["t_dt"]->format('Y-m-d H:i:s'). PHP_EOL;
echo '$m->data["t_dt"] returns => ' . $m->data["t_dt"]->format('Y-m-d H:i:s') . PHP_EOL;
echo '$m->dirty["t_dt"] returns => ' . (!isset($m->dirty["t_dt"]) ? 'null' : $m->dirty["t_dt"]->format('Y-m-d H:i:s')) . PHP_EOL;

$m->tryLoad(1);

echo '=== afterLoad' . PHP_EOL;

echo "is Dirty : " . ($m->isDirty(['t_dt']) ? 'true' : 'false') . PHP_EOL;
echo '$m["t_dt"] returns => ' . $m["t_dt"]->format('Y-m-d H:i:s'). PHP_EOL;
echo '$m->data["t_dt"] returns => ' . $m->data["t_dt"]->format('Y-m-d H:i:s') . PHP_EOL;
echo '$m->dirty["t_dt"] returns => ' . (!isset($m->dirty["t_dt"]) ? 'null' : $m->dirty["t_dt"]->format('Y-m-d H:i:s')) . PHP_EOL;

$m["t_dt"] = "18:00:00";

echo '=== afterChange' . PHP_EOL;

echo "is Dirty : " . ($m->isDirty(['t_dt']) ? 'true' : 'false') . PHP_EOL;
echo '$m["t_dt"] returns => ' . $m["t_dt"]->format('Y-m-d H:i:s'). PHP_EOL;
echo '$m->data["t_dt"] returns => ' . $m->data["t_dt"]->format('Y-m-d H:i:s') . PHP_EOL;
echo '$m->dirty["t_dt"] returns => ' . $m->dirty["t_dt"]->format('Y-m-d H:i:s') . PHP_EOL;

echo (new \DateTime())->format('a');
```
this is my output 

=== beforeSave
is Dirty : true
$m["t_dt"] returns => 0000-01-01 18:00:00
$m->data["t_dt"] returns => 0000-01-01 18:00:00
$m->dirty["t_dt"] returns => null
=== afterSave
is Dirty : false
$m["t_dt"] returns => 1970-01-01 18:00:00
$m->data["t_dt"] returns => 1970-01-01 18:00:00
$m->dirty["t_dt"] returns => null
=== afterLoad
is Dirty : false
$m["t_dt"] returns => 1970-01-01 18:00:00
$m->data["t_dt"] returns => 1970-01-01 18:00:00
$m->dirty["t_dt"] returns => null
=== afterChange
is Dirty : true
$m["t_dt"] returns => 0000-01-01 18:00:00
$m->data["t_dt"] returns => 0000-01-01 18:00:00
$m->dirty["t_dt"] returns => 1970-01-01 18:00:00

PhilippGrashoff:
Agree with Imants that having time fields as DateTime object seems to have no benefit. Its a shame PHP does not have Time and Date objects but DateTime only. I work a lot with dates and times (one of my project is a calendar) and I ran into trouble so often already comparing dates, times and so on where the other, non used part (time part of date, date part of time) made trouble.

So maybe having time fields not as DateTime object but as HH:ii:ss string is worth thinking about? I'd prefer a string to an integer, as it would saving casting 2 times:
1) for saving into db
2) for output to a user.

abbadon1334:
create a new PR #394 i hope this time is correct

PR #390, opened 2019-02-25
Add Persistence which use \PhpOffice\PhpSpreadsheet

abbadon1334:
All is in the doc but still working on.
Need some improvements for sure writing data on every change is bad.
i think the best performance can be accomplished using csv has read and write format, after that call  saveSheetAs($filename) that permit to save the actual content of the Persistence in another format and there use xls or pdf that are heavy calculation methods.

I tried testing many possible mess in PHPUnit + the PHPUnit tests of Persistence_CSV, if someone have an idea to add more.

i write docs based on Persistence_CSV, but i don't know why is keep unformatting code part in rst.


DarkSide666:
This would be very good to have.
One model/table = one Excel sheet. That way you can theoretically use Excel as quite good database. Of course for not huge amounts of data.
That would be especially useful to be able to import/export data models to/from Excel spreadsheets not just plain CSV file.

romaninsh:
I'd vote to have this in a separate repository due to requirements.

romaninsh:
I'd love to see support for ranges too (when linking model). Here is my potential use case:

 1. spreadsheet is provided with test-data.
 2. range is defined where test-data is located
 3. persistence is associated with model 
 4. copying data from external source (aggregation report) into spreadsheet.

After that spreadsheet is saved for the user. We now have an ability to generate excel reports with complex sheet logic, various tables per sheet and ability to "insert".

Additionally we want to keep formulas working (e.g. totals).

romaninsh:
[nominal-activity2.xlsx](https://github.com/atk4/data/files/2978147/nominal-activity2.xlsx)

abbadon1334:
> One model/table = one Excel sheet. That way you can theoretically use Excel as quite good database. Of course for not huge amounts of data.
> That would be especially useful to be able to import/export data models to/from Excel spreadsheets not just plain CSV file.
Good idea! but i think that must be a separate feature like ex:Persistence_PhpSpreadsheetAsDB

> I'd vote to have this in a separate repository due to requirements.
Agree, i already said this in chat, because phpSpreadsheet requires a lot of stuff that is not needed by atk4/data

> I'd love to see support for ranges too (when linking model). Here is my potential use case:
>     1. spreadsheet is provided with test-data.
>     2. range is defined where test-data is located 
>     3. persistence is associated with model 
>     4. copying data from external source (aggregation report) into spreadsheet. 
> 
> After that spreadsheet is saved for the user. We now have an ability to generate excel reports with complex sheet logic, various tables per sheet and ability to "insert".
Good point, what you think about this : 

1. define ranges on open file like in html templates with {$field_name}
2. define a key for defined range
3. store the row position of the defined range respect of the others defined ranges
4. on every insert increase row position of every range

> Additionally we want to keep formulas working (e.g. totals).
Ok

i hope this week i can go back coding, too much starting projects and issue still to be solved out of atk

DarkSide666:
@romaninsh I think what you're talking about here goes outside of _Persistence_ scope. It's mot related to _report_ scope.

As I understood Excel as Persistence is:
* each sheet = like database table
* each column = database table field
* each column first top cell is database table field name

With this setup I should be able to:
* load(), save(), insert(), delete() - basically do all CRUD operations.
* maybe some more functionality required by Persistence class, like _action()_etc.

romaninsh:
Yes, I was thinking about a more high-level features. By logic was - if we have PHPSpreadsheet as an external add-on, it might as well have lots of interesting features inside it that make use of PHPSpreadsheet.

@DarkSide666 what you propose is a very good "basic functionality" for atk4/spreadsheet. I can set that repository up. 👍 ?

abbadon1334:
the level is basic letterally a copy of CSV, but can already save a model from a filtered grid to multiple formats in 3-4 lines of code in place of using external solutions.

We will speak together about new features.

@romaninsh if you wants to make it as an addon which i agree, because it has some unnecessary dependency, just prepare it.

there is only one problem, i think to be used like a real persistence is slow but not slow, sooooo sloooowww.

romaninsh:
Yes please create add-on. Even if slow, this might be good for us in Saasty, we only have CSV import, but would like to import from all spreadsheets.

PR #391, opened 2019-02-25
Allow non-scalar value on serialize field

PhilippGrashoff:
If a field has a serialization like "json", "serialize" set, it may have non-scalar values. At least I store serialized values in fields of 'type' => 'text'.
This PR adds an additional condition in normalize()'s check for scalar values. If $this->serialize is set, value does not need to be scalar as its made scalar by serialization on save.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -261,7 +261,7 @@ class Field implements Expressionable
             }
 
             // validate scalar values
-            if (in_array($f->type, ['string', 'text', 'integer', 'money', 'float']) && !is_scalar($value)) {
+            if (in_array($f->type, ['string', 'text', 'integer', 'money', 'float']) && !$this->serialize && !is_scalar($value)) {
                 throw new ValidationException([$this->name => 'Must use scalar value']);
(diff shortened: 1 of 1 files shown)


PhilippGrashoff:
```

class Stock extends \atk4\data\Model {
    public $table = 'stock_cache';

    public function init() {
        parent::init();
        $this->addFields([
            ['tour_type_id',   'type' => 'integer'],
            ['last_updated',   'type' => 'datetime'],
            ['last_exported',  'type' => 'datetime'],
            ['cache',          'type' => 'text',       'serialize' => 'json'],
        ]);
    }


....
        $this->set('cache', $cached_dates);
        //$cached_dates is an array. This throws a validationException as a non-scalar value is used

....
```
So the solution would be to use another field type for serialized fields than text?

PhilippGrashoff:
Ah, I see in docs that for serialized fields no type is set at all... ok, will delete this PR, thought a field type must be set... and set it to "text" as its a text field in DB.

PR #394, opened 2019-03-06
Fix #389 Field Time after save wrong return value

abbadon1334:
#389 wrong PR format

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -348,7 +348,8 @@ class Field implements Expressionable
                 }
                 if ($f->type == 'time') {
                     // remove date portion from date type value
-                    $value->setDate(0, 1, 1);
+                    // need 1970 in place of 0 - DB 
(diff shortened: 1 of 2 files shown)


PhilippGrashoff:
Think we should add some test for it to prevent that issue happening again, something like:
```
   public function testDirtyWorksAsWanted() {
        $tour = new \EOO\Data\Tour(self::$app->db);
        $tour->set('start_time', '18:00');
        $this->assertTrue($tour->isDirty('start_time'));
        $tour->save();
        $this->assertFalse($tour->isDirty('start_time'));
        $tour->tryLoad($tour->get('id'));
        $this->assertFalse($tour->isDirty('start_time'));
        $tour->set('start_time', '18:00');
        $this->assertFalse($tour->isDirty('start_time'));
    }
```

PR #397, opened 2019-03-16
Fix logic error in example

PhilippGrashoff:
Checking if field value is unique: The condition has to be that the id of the new Instance IS NOT the id of the original record.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -486,7 +486,7 @@ Start by creating a beforeSave handler for Order::
 
             if (
                 $m->newInstance()
-                    ->addCondition('client_id', $m['client_id'])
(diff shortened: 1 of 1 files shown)


PhilippGrashoff:
Ah forget it I took client_id for id_field...

DarkSide666:
but `    ->addCondition($m->id_field, '!=', $m->id)   // has another order` should be there.
can you change your PR and add that line instead and i will merge in.

DarkSide666:
@romaninsh why not? I think we just need to change that one line there from client_id check to id and that's all. Don't you think so?

romaninsh:
i think the text explaining example was a bit too brief. Was unclear to me what i wrote myself 😄. probably need to better explain case and how it's solved.

PR #398, opened 2019-03-20
Dark side666 patch 1

romaninsh:
Old branch, but looks still relevant.


romaninsh:
Looks like you forgot to create PR, @DarkSide666

PR #399, opened 2019-03-20
Implements User Actions as per wiki

romaninsh:
https://github.com/atk4/data/wiki/ADP1903

Adds support for "User Actions" into model. You may declare user actions when initializing model. The UI, API or 3rd party will correctly identify actions and display them as buttons.

``` php
$this->addAction('reset_password');
```

This will allow user to click **Reset Password** button inside CRUD.

Moved TODO into : https://github.com/atk4/data/projects/5#card-19128771


funyx:
i think this feature should extend the Persistence::action implementation probably via new \atk4\data\Persistence\Action class which will then be extended into per persistence type Actions e.g. :
- \atk4\data\Persistence\SQL will have many action 'containers' in the \atk4\data\Persistence\Action\SQL one of which will be 

```
namespace atk4\data\Persistence\Action\SQL;

use \atk4\data\Persistence;

class Select extends Persistence\Action\SQL
{
    // chmod :
    // c - create
    // r - read
    // u - update
    // d - delete
    // full access -> protected $chmod = 'crud';

    protected $chmod = 'r';

    public function init()
    {
        parent::init();
        $this->dsql->*****
    }
}
```

```
class SQL extends Persistence
{
     public function init()
     {
        ***
        if($this->action_select_seed) {
            $select = $this->factory(..
            $this->addAction($select,$e_defaults=null);
        }
        ***
     }
     ***
     public function addModel(***)
     {
          ***
          // filter actions based on $m->chmod
          if($m->chmod == 'crud'){ // new Model->chmod = 'crud' by default
               // add all
          }
          ***
     }
     ***
     public function action(***)
     {
         if($this->getAction($actionName) && $this->getAction($actionName)->disabled){
              return throw new Exception(***
         }
     }
     public function getAction($actionName)
     {
         $this->getElement($actionName);
     }
     public function addAction($seed,$defaults=null)
     {
         $this->add($seed,$defaults);
     }

     public function removeAction($actionName)
     {
         unset($this->elements[$actionName]);
     }
}
```

....the magic should happen in the Persistence instance(s)...

romaninsh:
@funyx unfortunatelly those are different type of actions.

- UserAction - implements a high-level action user can perform (by clicking button) in UI or through API with a generic implementation only.
 - Actions - object returned by $model->action() which drops you to a lower level of persistence implementation (data->dsql)

Having both named actions are confusing, but I can't think of a better way.

src/Model.php
```diff
@@ -871,6 +879,118 @@ public function offsetUnset($name)
...
+     * getting.
+     *
+     * @param string $name
+     *
+     * @return UserAction\Action|false
+     */
+    public function hasAction($name)
+    {
+        $actions = array_filter(array_keys($this->elements), function ($var) use ($name) {
+            return stripos($var, 'action:'.$name) === 0;
```
DarkSide666: it would be good to refactor this and step out of hasAction method as soon as first matching action is found.

src/UserAction/Action.php
```diff
@@ -0,0 +1,115 @@
...
+
+    public function init()
+    {
+        if ($this->caption === null) {
+            $this->caption = substr($this->short_name, 7); // remove action: in front
+        }
+        $this->init_();
+    }
+
+    public function execute(...$args)
```
DarkSide666: No description

src/UserAction/Action.php
```diff
@@ -0,0 +1,115 @@
...
+        } elseif (is_string($this->callback)) {
+            return call_user_func_array([$this->owner, $this->callback], $args);
+        } else {
+            array_unshift($args, $this->owner);
+
+            return call_user_func_array($this->callback, $args);
+        }
+    }
+
+    public function preview(...$args)
```
DarkSide666: No description

src/UserAction/Action.php
```diff
@@ -0,0 +1,115 @@
...
+    public function execute(...$args)
+    {
+        // todo - assert owner model loaded
+
+        // todo - start transaction, if atomic
+
+        // todo - pass model as first argument ?
+
+        if (is_null($this->callback)) {
+            $callback = $this->callback ?: [$this->owner, str_replace('action:', '', $this->short_name)];
```
DarkSide666: if this->callback is null, then this 2nd condition will never be true

src/UserAction/Action.php
```diff
@@ -0,0 +1,115 @@
...
+            array_unshift($args, $this->owner);
+
+            return call_user_func_array($this->callback, $args);
+        }
+    }
+
+    public function preview(...$args)
+    {
+        if (is_null($this->preview)) {
+            $preview = $this->preview ?: [$this->owner, str_replace('action:', '', $this->short_name)];
```
DarkSide666: if this->preview is null, then this 2nd condition will never be true

src/UserAction/Action.php
```diff
@@ -0,0 +1,115 @@
...
+
+        // todo - pass model as first argument ?
+
+        if (is_null($this->callback)) {
+            $callback = $this->callback ?: [$this->owner, str_replace('action:', '', $this->short_name)];
+
+            return call_user_func_array([$this->owner, $callback], $args);
+        } elseif (is_string($this->callback)) {
+            return call_user_func_array([$this->owner, $this->callback], $args);
+        } else {
```
DarkSide666: better `elseif (is_callable($this->callback)` here

src/UserAction/Action.php
```diff
@@ -0,0 +1,115 @@
...
+
+    public function preview(...$args)
+    {
+        if (is_null($this->preview)) {
+            $preview = $this->preview ?: [$this->owner, str_replace('action:', '', $this->short_name)];
+
+            return call_user_func_array([$this->owner, $preview], $args);
+        } elseif (is_string($this->preview)) {
+            return call_user_func_array([$this->owner, $this->preview], $args);
+        } else {
```
DarkSide666: better `elseif (is_callable($this->preview)` here

tests/UserActionTest.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        parent::init();
+
+        $this->addField('name');
+
+        // this action can be invoked from UI
+        $this->addAction('send_reminder');
+
+        // this action will be system action, so it will not be invokable from UI
+        $a = $this->addAction('backup_clients', null, [], ['scope' => UserAction\Action::ALL_RECORDS, 'system' => true]);
```
DarkSide666: no need for 2nd and 3rd parameters here

src/Model.php
romaninsh: yeh, i was contemplating about having alternative to "elements" for all those.. Not sure how though.

src/UserAction/Generic.php
```diff
@@ -0,0 +1,134 @@
...
+class Generic
+{
+    use DIContainerTrait;
+    use TrackableTrait;
+    use InitializerTrait {
+        init as init_;
+    }
+
+    /** Defining scope of the action */
+    const NO_RECORDS = 'none'; // e.g. add
```
abbadon1334: why in place of using const and add methods that are not used, use some interface to classify the usage ?
I mean in place of a "sort of abstract" class Generic, add something like this :
- Action/iPrevieweable
- Action/iSingleRecord
- Action/iMultipleRecord
- Action/iAllRecord
After that we can make an abstract class : 
- AbstractAction
and a final : 
GenericAction
to force the extension of Abstract and not the Generic

i don't want to add only complexity, but i think that we need consistency and check over Type and Contracts is better than check over const.

romaninsh: I thought about this, and probably can be done in the future, for now i'd like to keep options open.

PR #401, opened 2019-03-26
Allow showing field on ui even if it is set as never_persist 

gowrav-vishwakarma:
if never_persist is set, and ui->editable is set true, even then Field is shown as non-editable. This correction is done to fix the issue.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -440,8 +440,12 @@ class Field implements Expressionable
     public function isEditable()
     {
         return $this->read_only || $this->never_persist
-            ? false
-            : (isset($this->ui['editable']) ? $this->ui['editable'] : !$this->system);
+            ? 
+            isset($this->ui['editable'])
+                ? 
+                $this->ui['editable']
+                : !$this->system
+            : false;
     }
 
     /**
--- a/src/Field.php
+++ b/src/Field.php
@@ -439,13 +439,9 @@ class Field implements Expressionable
      */
     public function isEditable()
     {
-        return $this->read_only || $this->never_persist
-            ? 
-            isset($this->ui['editable'])
-                ? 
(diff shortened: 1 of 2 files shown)


gowrav-vishwakarma:
Use 
`$model->addField('description',['never_persist'=>true,'ui'=>['editable'=>true]]);`

gowrav-vishwakarma:
Does this pull request sound okay?

As without it, there are two problems

1. No use of never_persist fields as they are disabled in forms no matter if you set them as ui->editbale->true
2. You don't get these never_persist fields in various hooks, so it makes them of no use.

src/Model.php
```diff
@@ -1442,54 +1442,61 @@ public function save($data = [], Persistence $to_persistence = null)
...
                     }
                 }
 
-                // No save needed, nothing was changed
-                if (!$data && !$dirty_join) {
-                    return $this;
-                }
 
-                if ($this->hook('beforeUpdate', [&$data]) === false) {
+                if ($this->hook('beforeUpdate', [&$data_all]) === false) {
```
romaninsh: $val = $m['field_never_persisted'];

src/Field.php
```diff
@@ -439,7 +439,7 @@ public function useAlias()
      */
     public function isEditable()
     {
-        return $this->read_only || $this->never_persist
+        return !$this->ui['editable'] && ($this->read_only || $this->never_persist)
```
DarkSide666: Probably this condition could be simplified to not check ui/editable multiple times:
```
return isset($this->ui['editable]) ? $this->ui['editable']
    : !( $this->read_only || $this->never_persist || $this->system );
```

gowrav-vishwakarma:
Can be better way to do this

PR #405, opened 2019-04-08
Refactor namespaces

DarkSide666:
* Refactor namespace of Persistence, Reference, Join classes.
* Adds Model->addCalculatedField() method, $_default_seed_addRef property for easier extending.
* Split Reference->getModel() into multiple methods for easier extending (protected methods addToPersistence() and getDefaultPersistence()).
* Adds comments, more strict method parameters etc.

This PR is extracted out of #400. This doesn't contain anything related to `containsOne` or `containsMany` implementation.
I have made proxy classes so this PR should keep backward compatibility. These proxy classes should be removed in next major version.


tests/RandomTest.php
```diff
@@ -4,7 +4,7 @@
 
 use atk4\data\Model;
 use atk4\data\Persistence;
-use atk4\data\Persistence_SQL;
+use atk4\data\Persistence\SQL as Persistence_SQL;
```
romaninsh: we should leave class Persistence_SQL extends Persistence\SQL for compatibility, so this shouldn't be necessary.

DarkSide666: We leave these old classes for compatibility for now, but should remove them at some point (in next major release anyway), so it's better to prepare our tests already for that.

PR #413, opened 2019-05-02
add condition if ref(hasOne) model not loaded

funyx:
I think the default behaviour of hasOne and hasMany should be as close as possible 
```
$p->load(2);
$p->ref('Stadium')->import([['name'=>'Nou camp nou']]);
```
..it should not matter if the reference is hasOne or hasMany..


romaninsh:
We've considered adding condition on model returned by hasOne(), but not only that would be a BC now, but might not be correct concept.

funyx:
@romaninsh that condition is added only in case the Model::id_field is equal to the Model::Reference_hasOne::our_field and the Model::Reference_hasOne::model is not loaded. I don't think it's a BC, nor an incorrect concept..

The main purpose of using model::ref is to give you a (referred) model instance that you can use to manipulate data within it, without writing additional code to set its state properly (setting the condition manually in this case), atm to implement the example above i have to do something like :
```
$p->load(2);
$p->hasOne('Stadium', [
    new Stadium(),
    'their_field' => 'player_id',
    'our_field' => 'id'
]);

// in case we have a player record but no stadium record for the player
// the stadium model has id=null and coditions=[]
// and to add a stadium i need to add the stadium.player_id value
// else PDO throws exception, because the foreign key constraint 
// (stadium.player_id -> player.id) is null when trying to insert

$s = $p->ref('Stadium'); // stadium model clone

if (is_a($p->getRef('Stadium'), atk4\data\Reference_One::class) &&
    !$s->loaded()) {

    // get their_field and set it with our_field's value 
    $sRef = (Array) $p->getRef('Stadium'); // because $our_field and $their_field are protected
    $sRef_their_field = $sRef["\0*\0".'their_field'];
    $sRef_our_field = $sRef["\0*\0".'our_field'];

    // set the condition manually with our_field's value
    $s->addCondition($sRef_their_field, $p[$sRef_our_field]);
}

// do stuff on referred model (in correct state)
$s->import([['name'=>'Nou camp nou']]);
```

src/Reference/HasOne_SQL.php
```diff
@@ -148,13 +148,29 @@ public function ref($defaults = []) : Model
     {
         $m = parent::ref($defaults);
 
+        if (!isset($this->owner->persistence) || !($this->owner->persistence instanceof Persistence_SQL)) {
```
DarkSide666: `Persistence\SQL` here

DarkSide666:
Provided test cases looks quite legit use-case. Just have to make sure it's no BC breaker.

src/Reference/HasOne_SQL.php
```diff
@@ -8,7 +8,7 @@
 use atk4\data\Field;
 use atk4\data\Field_SQL_Expression;
 use atk4\data\Model;
-use atk4\data\Persistence\SQL as Persistence_SQL;
+use atk4\data\Persistence\SQL;
```
DarkSide666: i think you should have `use atk4\data\Persistence;` here isn't it?

romaninsh: actually, that wouldn't be required too, since we are in the data namespace :D

romaninsh:
Tests fail.............

src/Reference/HasOne_SQL.php
funyx: it should :
```
use atk4\data\Persistence
```

PR #414, opened 2019-05-03
should solve bug #698

gowrav-vishwakarma:
in field ui=>editable=>true should override never_persists's readonly settings


romaninsh:
I agree that "read-only" should override never-persist settings like this:

 read-only is null - determine based on never-persist.
 read-only is true - display as readonly
 read-only is false - display normally.

@DarkSide666 In some cases you set never-persist but you manually handle field in beforeSave event, so it's a normal condition.

romaninsh:
@gowrav-vishwakarma , loose the "@" and we can merge.

DarkSide666:
> I agree that "read-only" should override never-persist settings like this:
> 
> read-only is null - determine based on never-persist.
> read-only is true - display as readonly
> read-only is false - display normally.
> 
> @DarkSide666 In some cases you set never-persist but you manually handle field in beforeSave event, so it's a normal condition.

Well currently in https://github.com/atk4/data/pull/414/files#diff-664bea79935ca266367c5a7a405b0909R442 read_only property is treated the same as never_persist, so that's not correct. read_only should take precedence over ui/editable if i understood @romaninsh correct.

gowrav-vishwakarma:
As per me,

Editable is itself a property and final authority to decide if a field is
editable directly. So it should be on top, then read only followed by never
persists.

DarkSide666:
Let's leave this for @romaninsh to decide that :)

romaninsh:
@gowrav-vishwakarma 's idea is fine, just need to fix implementation!

romaninsh:
@gowrav-vishwakarma - it's not only "remove" it but also implement a correct checking, now we've got warnings!

PR #431, opened 2019-05-22
Nesty BUG. was still returning wrong

gowrav-vishwakarma:
Now, working fine

a small precedence bug.


PhilippGrashoff:
As this is already the second of these I am aware of, maybe we should go with the official PHP advise and do not nest ` ? : ` ?

DarkSide666:
Just always have to use parenthesis then it's OK.

PR #440, opened 2019-06-21
Add support, specifying which fields can be dirty during action call.

romaninsh:
Currently it is possible to mess up with the action if you modify the record first:

``` php
$user->load(3);
$user['is_admin']=true;
$user->getAction('send_password_reminder')->execute();
```

This PR implements a check for dirty fields in the model. By default dirty fields are not allowed at all, but you can define them as allowed:

``` php
$user->addAction('signup', ['fields'=>['name', 'password', 'email'], 'scope'=>'none']);
```

The implementation of the signup method would be:

``` php
function signup() {
    $this->save();
    $this->sendWelcomeEmail();
}
```

And the invocation:

``` php
$user = new User($db);
$user['name']='John';
$user['password'] = 'pass';
$user['email'] = 'ema@com';
// $user['is_admin']=true; - uncommenting this will cause exception!
// $user->load(2);    - this also would cause exception
$user->getAction('signup')->execute();
```

ATK UI should also explore $fields - and if it is a non-empty array, then those fields can be modified by a Form prior to calling the action.


src/UserAction/Generic.php
```diff
@@ -70,28 +77,78 @@ public function init()
...
 
-        if ($this->callback === null) {
-            $cb = [$this->owner, substr($this->short_name, strlen('action:'))];
-        } elseif (is_string($this->callback)) {
-            $cb = [$this->owner, $this->callback];
-        } else {
-            array_unshift($args, $this->owner);
-            $cb = $this->callback;
+        try {
+            if (!$this->enabled) {
```
DarkSide666: This doesn't support `enabled` property as callable. See how it's described above (line 55) https://github.com/atk4/data/pull/440/files#diff-478fc934e86fa5d120af35d752fe76feR55

tests/UserActionTest.php
```diff
@@ -99,16 +99,17 @@ public function testBasic()
...
+        $client->load(1);
+        $this->assertEquals('John', $client->getAction('say_name')->execute());
 
-        $client->getAction('say_a')->preview = function ($m, $arg) {
-            return ($m instanceof ACClient) ? 'will say '.$arg : 'will fail';
+        $client->getAction('say_name')->preview = function ($m, $arg) {
+            return ($m instanceof ACClient) ? 'will say '.$m['name'] : 'will fail';
         };
-        $this->assertEquals('will say x', $client->getAction('say_a')->preview('x'));
+        $this->assertEquals('will say John', $client->getAction('say_name')->preview('x'));
```
DarkSide666: What is `'x'` here?

src/UserAction/Generic.php
romaninsh: well spotted! Implemented too.

tests/UserActionTest.php
romaninsh: just a random string

PR #442, opened 2019-06-21
EXPEREMENTAL - use Actions internally

romaninsh:
Define actions for Add, Delete, Edit and Validate (system). 

Smart CRUD should be able to automatically discover those actions and show to the user. This would allow us to fully refactor CRUD to rely only on actions and don't hard-code anything!

Implications of this change are really awesome, you would be able to manipulate those native model actions like any other action defined by yourself or third party.


src/Model.php
```diff
@@ -371,7 +373,39 @@ public function init()
...
+            'callback'=> 'save',
+            'ui'      => ['icon'=>'edit'],
+        ]);
+        $this->addAction('delete', [
+            'scope'=> UserAction\Generic::SINGLE_RECORD,
+            'ui'   => ['icon'=>'trash', 'danger', 'confirm'=>'Are you sure?'],
+        ]);
+
+        $this->addAction('validate', [
+            //'scope'=> any!
```
DarkSide666: Probably should be UserAction\Generic::SINGLE_RECORD scope, but also could be UserAction\Generic::NO_RECORDS. UserAction\Generic do not support multiple scopes for one action. Maybe it should?

src/Model.php
```diff
@@ -371,7 +373,39 @@ public function init()
...
+            return; // don't declare actions for model without id_field
         }
+
+        if ($this->read_only) {
+            return; // don't declare action for read-only model
+        }
+
+        // Declare our basic CRUD actions for the model.
+        $this->addAction('add', [
+            'fields'  => true,
```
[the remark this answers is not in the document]
romaninsh: Or false. I think. Will amend.

src/Model.php
romaninsh: Yes I was considering. Will discuss tomorrow

PR #443, opened 2019-07-08
Add email field validation

acicovic:
Adds email field type for validation purposes.

Do you think it should be documented in https://agile-data.readthedocs.io/en/latest/types.html?

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -30,7 +30,7 @@ class Field implements Expressionable
      * Field type.
      *
      * Values are: 'array', 'boolean', 'date', 'datetime', 'email', 'float',
-     *             'integer', 'money', 'object', 'string', 'time', 'text'.
+     *             'integer', 'money', 'object', 'string', 'text', 'time'.
      * Can also be set to unspecified type for your own custom handling.
      *
      * @var string
--- a/src/Field.php
+++ b/src/Field.php
@@ -29,8 +29,8 @@ class Field implements Expressionable
     /**
      * Field type.
      *
-     * Values are: 'string', 'text', 'boolean', 'integer', 'money', 'float',
-     *             'date', 'datetime', 'time', 'array', 'object'.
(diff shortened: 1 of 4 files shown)


src/Field.php
```diff
@@ -322,6 +322,16 @@ public function normalize($value)
                     throw new ValidationException([$this->name => 'Must not be a zero']);
                 }
                 break;
+            case 'email':
+                // remove all line-ends and trim
+                $value = trim(str_replace(["\r", "\n"], '', $value));
+                if ($this->required && empty($value)) {
```
DarkSide666: This check here probably is not required because it's already validated somewhere else. See line 256 above.
Also see line 264 above, need to add 'email' there too.

DarkSide666: Also adding it in docs would be nice.

acicovic: 1. The check you see is also done for 'string' type (lines 275-280) and there's also a similar validation in 'text' type (282-287) so it may need to be included. The 256 line check just checks for required and empty but without trimming whitespace or removing possible newlines.
2. I've added email to scalar validation.

acicovic: I've added documentation.

In the 'additional documentation' commit I've also arranged field types alphabetically. In that commit, the email addition is in line 115. I think that certain things in the docs should be listed alphabetically because it's more intuitive and easier to spot what you're searching for.

romaninsh:
Shouldn't this be more consistent with https://github.com/atk4/data/pull/375 ? 

e.g. adding Field/Email.php class and custom decorators for UI?

acicovic:
@romaninsh I don't know, it's only a validation. For the time being it doesn't feature any properties or decorators. I'm not sure how to implement it the way you refer to but I don't mind if someone ports it or tells me how I should do it.

romaninsh:
See also: https://github.com/atk4/data/issues/308

romaninsh:
@acicovic will post a new PR..

PR #445, opened 2019-07-10
Model::addField() to use $model->fields property rather than elements (and new Trait)

romaninsh:
Implements getFields() and refactors fields in general to use MultiContainerTrait..

Depends on #444 (next is to refactor "references" into separate array)

## Note on type->seed conversion

Previously there was a check - type=boolean -> use Bolean class. I want this to be similar to factories in Table, Form etc. However, i think Persistence should be able to influence which classes are used for types.

Model fields can also be used without persistence and as a result there are some code duplication right now (fieldFactory). 

Please suggest how to resolve that.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -6,13 +6,13 @@ namespace atk4\data;
 
 use ArrayAccess;
 use atk4\core\AppScopeTrait;
+use atk4\core\CollectionTrait;
 use atk4\core\ContainerTrait;
 use atk4\core\DIContainerTrait;
 use atk4\core\DynamicMethodTrait;
 use atk4\core\FactoryTrait;
 use atk4\core\HookTrait;
 use atk4\core\InitializerTrait;
-use atk4\core\CollectionTrait;
 use atk4\core\NameTrait;
 use atk4\dsql\Query;
 use IteratorAggregate;
@@ -1627,8 +1627,9 @@ class Model implements ArrayAccess, IteratorAggregate
      * @param mixed $field_name
      * @param mixed $value
      *
-     * @return $this
      * @throws \atk4\core\Exception
+     *
+     * @return $this
      */
     public function loadBy(string $field_name, $value)
     {
@@ -1664,10 +1665,11 @@ class Model implements ArrayAccess, IteratorAggregate
      * Will not throw exception if record doesn't exist.
      *
      * @param string $field_name
-     * @param mixed $value
+     * @param mixed  $value
+     *
+     * @throws \atk4\core\Exception
      *
      * @return $this
-     * @throws \atk4\core\Exception
      */
     public function tryLoadBy(string $field_name, $value)
     {
--- a/src/Model.php
+++ b/src/Model.php
@@ -45,7 +45,7 @@ class Model implements ArrayAccess, IteratorAggregate
      *
      * @var string|array
      */
-    public $_default_seed_addField = '\atk4\data\Field';
+    public $_default_seed_addField = Field::class;
 
     /**
      * The class used by addField() method.
@@ -431,29 +431,23 @@ class Model implements ArrayAccess, IteratorAggregate
     /**
      * Adds new field into model.
      *
-     * @param string $field
-     * @param array  $defaults
+     * @param string $name
+     * @param array  $seed
      *
      * @throws \atk4\core\Exception
      *
      * @return Field
      */
-    public function addField($field, $defaults = [])
+    public function addField($name, $seed = [])
     {
-        // compatibility: for field types
-        $class = $this->_default_seed_addField;
-        if (is_array($defaults) && isset($defaults['type'])) {
-            switch (strtolower($defaults['type'])) {
-                case 'boolean':
-                    $class = 'Boolean';
-                    break;
-            }
+
+        if ($this->persistence) {
+            $field = $this->persistence->fieldFactory($seed);
+        } else {
+            $field = $this->factory($seed);
         }
 
-        $field_object = $this->factory($this->mergeSeeds($defaults, $class), null, '\atk4\data\Field');
-        $this->add($field_object, $field);
-
-        return $field_object;
+        return $this->_addIntoCollection($name, $field, 'fields');
     }
 
     /**
@@ -463,6 +457,7 @@ class Model implements ArrayAccess, IteratorAggregate
      * @param array $defaults
      *
      * @return $this
+     * @throws \atk4\core\Exception
      */
     public function addFields($fields = [], $defaults = [])
     {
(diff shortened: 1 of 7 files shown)


src/Model.php
```diff
@@ -404,40 +444,57 @@ public function validate($intent = null)
...
-        $field_object = $this->factory($this->mergeSeeds($defaults, $class), null, '\atk4\data\Field');
-        $this->add($field_object, $field);
-
-        return $field_object;
+        return $this->_addIntoCollection($name, $field, 'fields');
     }
 
+    protected $typeToFieldSeed = [
+        'boolean' => ['Boolean'],
+    ];
```
DarkSide666: I think this is good.

For other basic field types in future could be like:
```
'integer' => ['Number', 'unsigned'=>false, 'decimal'=>0, 'size'=>12],
'date' => ['DateTime', 'time'=>false, 'format'=>'Y-m-d'],
```
Basically it converts _basic_ field types to extended Field classes and sets some properties of these objects.

src/Model.php
```diff
@@ -459,28 +516,31 @@ public function addFields($fields = [], $defaults = [])
         return $this;
     }
 
+    /**
+     * Remove field that was added previously.
+     *
+     * @param $name
```
DarkSide666: `@param string $name`

src/Model.php
```diff
@@ -459,28 +516,31 @@ public function addFields($fields = [], $defaults = [])
...
+     *
+     * @throws \atk4\core\Exception
+     */
+    public function removeField($name)
+    {
+        return $this->_removeFromCollection($name, 'fields');
```
DarkSide666: no need for `return` here because it will not return anything anyway.
or return $this to be able to chain these methods (and write that in method comments above too `@return $this`)

src/Model.php
```diff
@@ -489,26 +549,25 @@ public function hasField($name)
      *
      * @param string|Field $name
```
DarkSide666: only string not string|Field

src/Model.php
```diff
@@ -600,6 +661,32 @@ public function isDirty($fields = [])
         return false;
     }
 
+    /**
+     * @param string|null $filter
```
DarkSide666: No method description

src/Model.php
```diff
@@ -600,6 +661,32 @@ public function isDirty($fields = [])
...
+     *
+     * @return array
+     */
+    public function getFields(string $filter = null)
+    {
+        if (!$filter) {
+            return $this->fields;
```
DarkSide666: Not sure about this, but if we don't set $filter, then it will return fields also outside of only_fields scope.
Is that OK?

src/Model.php
```diff
@@ -600,6 +661,32 @@ public function isDirty($fields = [])
...
+        return array_filter($this->fields, function ($field, $name) use ($filter) {
+
+            // do not return fields outside of "only_fields" scope
+            if ($this->only_fields && !in_array($name, $this->only_fields)) {
+                return false;
+            }
+
+            switch ($filter) {
+                case 'system': return $field->system;
+                case 'editable': return $field->ui['editable'] ?? !$field->system;
```
DarkSide666: Should probably use $field->isEditable() and $field->isVisible() methods here because:
1. DRY principle
2. editable will then take read_only and never_persist properties into account too.

src/Model.php
```diff
@@ -1565,15 +1628,15 @@ public function tryLoadAny()
      *
      * @return $this
      */
-    public function loadBy($field, $value)
+    public function loadBy(string $field_name, $value)
```
DarkSide666: should also fix $field_name type in method @param comment

src/Persistence.php
```diff
@@ -76,12 +76,41 @@ public function disconnect()
...
+    /**
+     * Given a field seed, return a feld object.
+     *
+     * @param string $type
+     *
+     * @throws \atk4\core\Exception
+     *
+     * @return Field
+     */
+    public function fieldFactory($seed = [])
```
DarkSide666: I still don't think this is needed in Persistence. See my other comment above.

DarkSide666 (changes requested):

src/Model.php
```diff
@@ -89,6 +105,16 @@ class Model implements \ArrayAccess, \IteratorAggregate
      */
     public $_default_seed_action = UserAction\Generic::class;
 
+    /**
+     * @var array Collection containing Field Objects - using key as the field system name
+     */
+    public $fields = [];
```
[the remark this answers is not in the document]
romaninsh: agreed. made them protected.

src/Model.php
```diff
@@ -404,40 +444,57 @@ public function validate($intent = null)
         return $errors;
     }
 
+    public function add($obj, $args = [])
+    {
+        $obj = $this->_add($obj, $args);
+        if ($obj instanceof Field) {
```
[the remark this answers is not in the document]
romaninsh: this is primarily a legacy fix, helping to weed out code..

actions are new and there should be no direct adding.

ideally i'd like to get rid of this method later.

src/Model.php
```diff
@@ -404,40 +444,57 @@ public function validate($intent = null)
         return $errors;
     }
 
+    public function add($obj, $args = [])
```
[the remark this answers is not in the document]
romaninsh: noted.

src/Model.php
romaninsh: yes, that seems to be the best, since we recommend to replace foreach($this->elements) with getFields() and that would break it for only_fields then. (2 tests actually break if i apply only_fields scope).

we could add new filter 'accessible', which would respect only_fields, if needed.

src/Model.php
romaninsh: addressed.

src/Persistence.php
romaninsh: removed.

src/Model.php
romaninsh: ok, changed, also dropped check for system in this case. Test continue to work, but i haven't done any tests for read-only and never_persist.

src/Model.php
DarkSide666: Fine.

src/Model.php
```diff
@@ -469,20 +453,34 @@ public function addField($name, $seed = [])
...
+            $field = $this->fieldFactory($seed);
         }
 
         return $this->_addIntoCollection($name, $field, 'fields');
     }
 
+    /**
+     * Given a field seed, return a feld object.
+     *
+     * @param string $type
```
DarkSide666: wrong @param :)

DarkSide666 (approved):

PR #447, opened 2019-07-26
Reference get caption

abbadon1334:
I have this code : 
```php
//Model Measure:
        $this->addField('code', ['type' => 'string', 'caption' => 'U.M.']);

//Model Article : 
        $this->hasOne('spec_measure', Measure::class)
                ->addField('spec_measure_code','code');

//Model ComponentOfArticleBOM
        $this->hasOne('article_id', Article::class)
                ->addField('spec_measure_code');
```
if i call `ComponentOfArticleBOM->getField('spec_measure_code')->caption` i get : "Spec Measure Code"

Everytime we add a field from a relation we have to redefine the caption it. Caption is lost, and if we change it we have to find all occurencies and change by hand.

With this PR i fix it and if i call : `ComponentOfArticleBOM->getField('spec_measure_code')->caption` i get : "U.M."

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -22,7 +22,7 @@ class Reference
 
     /**
      * Owner Model of the reference.
-     * override the hint type definition already present in TrackableTrait
+     * override the hint type definition already present in TrackableTrait.
      *
      * @var Model
      */
@@ -163,6 +163,9 @@ class Reference
      *
      * @param Model $model
      * @param array $defaults
+     *
+     * @throws Exception
+     * @throws \atk4\core\Exception
      *
      * @return Model
      */
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -106,6 +106,8 @@ class Reference
      *
      * @param array $defaults Properties
      *
+     * @throws \atk4\core\Exception
+     *
      * @return Model
      */
     public function getModel($defaults = []) : Model
@@ -126,7 +128,7 @@ class Reference
         // if model is Closure, then call it and return model
         if (is_object($this->model) && $this->model instanceof \Closure) {
             $c = $this->model;
-            $c = $c($this->owner, $this, $defaults);
+            $c = ($c)($this->owner, $this, $defaults);
 
             return $this->addToPersistence($c, $defaults);
         }
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -21,6 +21,14 @@ class Reference
     use \atk4\core\FactoryTrait;
 
     /**
+     * Owner Model of the reference.
+     * override the hint type definition already present in TrackableTrait
+     *
+     * @var Model
+     */
+    public $owner;
+
+    /**
      * Use this alias for related entity by default. This can help you
      * if you create sub-queries or joins to separate this from main
      * table. The table_alias will be uniquely generated.
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -127,8 +127,7 @@ class Reference
 
         // if model is Closure, then call it and return model
         if (is_object($this->model) && $this->model instanceof \Closure) {
-            $c = $this->model;
-            $c = ($c)($this->owner, $this, $defaults);
+            $c = ($this->model)($this->owner, $this, $defaults);
 
             return $this->addToPersistence($c, $defaults);
         }
@@ -205,6 +204,8 @@ class Reference
      *
      * @param array $defaults Properties
      *
+     * @throws \atk4\core\Exception
+     *
      * @return Model
      */
     public function ref($defaults = []) : Model
@@ -218,6 +219,8 @@ class Reference
      * or scope.
      *
      * @param array $defaults Properties
+     *
+     * @throws \atk4\core\Exception
      *
      * @return Model
      */
--- a/src/Reference/HasOne.php
+++ b/src/Reference/HasOne.php
@@ -4,6 +4,7 @@
 
 namespace atk4\data\Reference;
 
+use atk4\core\Exception;
 use atk4\data\Field;
 use atk4\data\Join;
 use atk4\data\Model;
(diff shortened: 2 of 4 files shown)


src/Reference.php
```diff
@@ -118,7 +128,7 @@ public function getModel($defaults = []) : Model
         // if model is Closure, then call it and return model
         if (is_object($this->model) && $this->model instanceof \Closure) {
             $c = $this->model;
-            $c = $c($this->owner, $this, $defaults);
+            $c = ($c)($this->owner, $this, $defaults);
```
DarkSide666: what this change does?

src/Reference/HasOne.php
```diff
@@ -228,18 +236,14 @@ public function ref($defaults = []) : Model
...
-            if ($this->owner[$this->our_field]) {
-                $m->tryLoad($this->owner[$this->our_field]);
-            }
-
-            return
-                $m->addHook('afterSave', function ($m) {
-                    $this->owner[$this->our_field] = $m->id;
-                });
+        }
+        if ($this->owner[$this->our_field]) {
```
DarkSide666: What if `$this->our_field` is not defined at all (equals `null`)? I think then it will throw exception here.
I know it was like that before too, but probably should be `if ($this->our_field !== null && $this->owner[$this->our_field])` here to be 100% sure it's set.

src/Reference/HasOne.php
```diff
@@ -228,18 +236,14 @@ public function ref($defaults = []) : Model
...
+            $m->tryLoad($this->owner[$this->our_field]);
         }
 
-        // can not load referenced model or set conditions on it, so we just return it
-        return $m;
+        return
```
DarkSide666: not completely sure about this change here. what if $this->our_field is not defined at all? then this will fail.
I understand why you split old if/else block into two if blocks, but this change here changes more than that.

DarkSide666 (changes requested):

src/Reference.php
```diff
@@ -20,6 +20,14 @@ class Reference
     use \atk4\core\DIContainerTrait;
     use \atk4\core\FactoryTrait;
 
+    /**
+     * Owner Model of the reference.
+     * override the hint type definition already present in TrackableTrait.
+     *
+     * @var Model
+     */
+    public $owner;
```
romaninsh: Defined in https://agile-core.readthedocs.io/en/develop/container.html?highlight=owner#trackable-trait

src/Reference.php
romaninsh: I think it executes closure with arguments. Interesting syntax

romaninsh: probably same thing, i don't mind either way.

src/Reference.php
abbadon1334: i know but intellisense get defintion of owner model as object in place of Model and don't work.

Reference has always a ->owner = Model, no? why not type hint it and let intellisense autocomplete?

src/Reference/HasOne.php
abbadon1334: looking at the previous version, code will never arrive here : https://github.com/atk4/data/blob/0036a94d140c3474c696bb622a861325802abb27/src/Reference/HasOne.php#L243

just because it will exit here : 
https://github.com/atk4/data/blob/0036a94d140c3474c696bb622a861325802abb27/src/Reference/HasOne.php#L227
or here : 
https://github.com/atk4/data/blob/0036a94d140c3474c696bb622a861325802abb27/src/Reference/HasOne.php#L236

src/Reference.php
abbadon1334: sorry the idea was to remove all "yellow" and "red" syntax notice.
is only a sort of much correct syntax.

```php
$c = ($this->model)($this->owner, $this, $defaults);
return $this->addToPersistence($c, $defaults);
```
no changes, at functional level just syntax.

src/Reference/HasOne.php
abbadon1334: We can put a conditional, but there was no need. probably because reference is never called directly, but from Model and is always associated with a field_name which in relations became our_field, i'm wrong?

DarkSide666: ok fine

src/Reference.php
DarkSide666: fine with me if that helps IDEs

src/Reference/HasOne_SQL.php
```diff
@@ -45,8 +48,12 @@ public function addField($field, $their_field = null) : Field_SQL_Expression
             $their_field = $field;
         }
 
+        // if caption is not defined in $defaults -> get it directly from the linked model field $their_field
+        $defaults['caption'] = $defaults['caption'] ?? $this->owner->refModel($this->link)->getField($their_field)->getCaption();
```
DarkSide666: I plan to implement this differently by using new `Field->getSeed()` method to pass all field properties (including caption) to new field. Anyway this is good for now, but will conflict with my PR later.

abbadon1334: OK, IMHO what is needed is to not redefining "caption" for reference imported fields, when you will break it we will go in this point and change the code, but final implementation will not have to change nothing.
What you think if i add a Test for this case, to see when this BC happen?

abbadon1334: Add Test. Not bad idea to have a test even for small things.

romaninsh (approved):
ok

src/Reference/HasOne_SQL.php
mvorisek, line 52: this is problematic when there is more than 1 reference, when the meaning of imported field is too broad - like `name` - in `user` table it is clear what is means, but if imported into `order` table, we `name` has completely different meaning

@abbadon1334 what is your motivation behind this?

would you be ok if we drop this and the test you added?

PR #453, opened 2019-08-23
Only call \DateTime::format if its an instance

PhilippGrashoff:
The value could be null. In this case this error is thrown:
``Error: Call to a member function format() on null``


PhilippGrashoff:
@DarkSide666 @romaninsh could one of you check if casting null to an empty string here is the correct behaviour in general?

PR #454, opened 2019-08-27
model default add field property

funyx:
i think its a bug preventing the model from using extended default_add_field class

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -473,7 +473,7 @@ class Model implements ArrayAccess, IteratorAggregate
         $seed = $this->mergeSeeds(
             $seed,
             isset($seed['type']) ? ($this->typeToFieldSeed[$seed['type']] ?? null) : null,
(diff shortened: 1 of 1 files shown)


src/Model.php
```diff
@@ -473,7 +473,7 @@ public function fieldFactory($seed = [])
         $seed = $this->mergeSeeds(
             $seed,
             isset($seed['type']) ? ($this->typeToFieldSeed[$seed['type']] ?? null) : null,
-            [Field::class]
+            [$this->_default_seed_addField]
```
romaninsh: i think you don't need brackets here. If it's a class (string) it should still work OK.

romaninsh:
Also - regarding defining model's properties from persistence - it's done here: https://github.com/atk4/data/blob/develop/src/Persistence/SQL.php#L139

Probably not the best approach, but it works. Can be refactored.

Issue #456, opened 2019-08-30
Add is_update to beforeSave Hook as well


PhilippGrashoff:
Hi, some month ago the afterSave hook got the very sensible second parameter is_update. beforeSave does not have this. It could be done very easily switching some lines:
Current:
```
            if ($this->hook('beforeSave') === false) {
                return $this;
            }

            $is_update = $this->loaded();
```

Switched:
```
            $is_update = $this->loaded();

            if ($this->hook('beforeSave', $is_update) === false) {
                return $this;
            }
```

Of course its not as important as in the afterSave hook, as you can easily check inside the beforeSave hook if the model is loaded or not, but that would make both hooks behave the same way.

abbadon1334:
LGTM even if we already have beforeInsert and beforeUpdate to clearly separate the operation.
This can shortens by 2 or 3 line the code.

romaninsh:
I'm all for consistency, so please go ahead and PR. No need to discuss even.

DarkSide666:
Yeah, looks very good. Can you make PR?

PhilippGrashoff:
Will do

DarkSide666:
implemented in #462

PR #460, opened 2019-09-06
add is_update to beforeSave hook

PhilippGrashoff:
add second parameter is_update to beforeSave hook to be in line with afterSave hook. Also added a new Test Class for all Model hook related things.
Fixes https://github.com/atk4/data/issues/456


PhilippGrashoff:
After losing a fight against git, could someone remove all commits but the last two ones? Thanks!

acicovic:
@PhilippGrashoff are you OK with closing this?

PR #461, opened 2019-09-06
Fix #460 - add is_update to beforeSave hook

### LINKED ISSUE #456: Add is_update to beforeSave Hook as well

acicovic:
This fixes #460. Code author is @PhilippGrashoff, credit goes to him.

> add second parameter is_update to beforeSave hook to be in line with afterSave hook. Also added a new Test Class for all Model hook related things.
> Fixes #456

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1735,11 +1735,12 @@ class Model implements ArrayAccess, IteratorAggregate
             if (($errors = $this->validate('save')) !== []) {
                 throw new ValidationException($errors, $this);
             }
-            if ($this->hook('beforeSave') === false) {
+
+            $is_update = $this->loaded();
+            if ($this->hook('beforeSave', [$is_update]) === false) {
                 return $this;
             }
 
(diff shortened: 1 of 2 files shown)


DarkSide666:
@acicovic, @PhilippGrashoff  Why did you close this PR? Where is actual implementation of this?

acicovic:
@DarkSide666 it was implemented in #462.

Issue #175, opened 2016-10-02
model->atomic() should have hook for a failure.


romaninsh:
When model's atomic operation is failing, don't just roll stuff back but also execute a special hook. This can be used for various reasons:

Save information into auditLog about failure.

```
$m->addHook('onRollback', function($m){ 
  $m->auditLog->registerFailure();
});
```

Upgrade schema

```
$m->addHook('onRollback', function($m, $exception){ 
  if($exception instanceof \PDOException) {
     $m->schema->upgrade();
  }
});
```

PR #466, opened 2019-09-19
Implements onRollback hook

DarkSide666:
onRollback Hook (#175 )
---------------

This hook is executed right after transaction fails and rollback is done.
This can be used in various situations.

Save information into auditLog about failure:
```
    $m->addHook('onRollback', function($m){ 
        $m->auditLog->registerFailure();
    });
```

Upgrade schema:
```
    $m->addHook('onRollback', function($m, $exception) { 
        if ($exception instanceof \PDOException) {
            $m->schema->upgrade();
            $m->breakHook(false); // exception will not be thrown
        }
    });
```

In first example we will register failure in audit log, but afterwards still throw exception.
In second example we will upgrade model schema and will not throw exception at all because we
break hook and return false boolean value.

**Includes:**
 - [x] implementation
 - [x] tests
 - [x] documentation
 - [x] small fix for UserAction (should use model atomic method)


DarkSide666:
One thing I was not sure about is - maybe we should call it `afterRollback` instead because it's actually run only after rollback is already done.

Maybe in some cases it would be nice to also have `beforeRollback` hook to be able to see model data before it is rolled back. But it's not that easy, because that can't be done in atk4/data repo level.
In SQL persistence atk4/dsql does actual rollback and only then throws exception which we can catch on atk4/data Model level :(

From atk4/dsql/Connection:
```
    public function atomic($f)
    {
        $this->beginTransaction();

        try {
            $res = call_user_func($f);
            $this->commit();

            return $res;
        } catch (\Exception $e) {
            $this->rollBack();

            throw $e;
        }
    }
```

DarkSide666:
* Model values are preserved after rollback, so we can get access to them in onRollback hook too.
* beforeRollback hook is useless because you can't not rollback if it needs to be rolled back. DB transaction will anyway rollback after connection close and that will be misleading.
* if we don't have beforeRollback hook, then i think it's better to name afterRollback hook as onRollback because _after_ implies that we also have _before_ which is not the case here.

PR #477, opened 2019-11-22
Implement #386

### LINKED ISSUE #386: Convert camel case field names to captions in model
It would be a good idea to also parse a camel case field names to captions

E.g.

```
$model->addField('ProductCode'); // caption here can be calclated to "Product Code" 
$model->addField('productCode'); // caption here can be calclated to "Product Code" as well
```

DarkSide666:
Implements #386 (fix #386).

- introduce Helper class as a place to put some shared functionality which is used in multiple classes
- `Helper::readableCaption($s)` now is used everywhere to generate model, field or action captions from object class names, field names or any other string. It also is aware of camelCase words now!
- added testcases

For example, this ugly string
`this\\ _isNASA_MyBigBull shit_123\Foo`
will be translated into nice looking, human readable caption
`This Is NASA My Big Bull Shit 123 Foo`


src/Util/Helper.php
```diff
@@ -0,0 +1,35 @@
...
+     */
+    public static function readableCaption($s)
+    {
+        //$s = 'this\\ _isNASA_MyBigBull shit_123\Foo';
+
+        // first remove not allowed characters and uppercase words
+        $s = ucwords(preg_replace('/[^a-z0-9]+/i', ' ', $s));
+
+        // and then run regex to split camelcased words too
+        $s = array_map('trim', preg_split('/^[^A-Z\d]+\K|[A-Z\d][^A-Z\d]+\K/', $s, -1, PREG_SPLIT_NO_EMPTY));
```
romaninsh: i think this should go to ATK Core..

DarkSide666: implemented

PR #490, opened 2019-12-01
Feature/set custom edit exec button

georgehristov:


georgehristov:
In current `develop` version the add / edit modal button has the same label as the action so 'edit' action has label 'Edit'. Introducing the 'execButton' option would allow devs to assign a custom button.
I data should not depend on ui then I will update so that only an array is passed and ui can then create the button from the array.

DarkSide666:
I think that would be better, because really, data should not have ui dependency.
Maybe you can pass only array with caption there, and let ui decide which class to use by default?

I will add @romaninsh and @ibelar to help decide how to do this.

PhilippGrashoff:
I also vote for the array seed, as that wouldnt instanciate the Button object right away but only when needed.

georgehristov:
I was considering between a seed or only the button label and initially opted for the seed.
Now I believe it would be better if data passes the button label only.
Will update the branch.

ibelar:
The actual implementation of UserAction executor allow for changing button label via the $action->caption and $action->description property . 

For example, changing execBtn caption and description for add action would be:

````
$ac = $country->getAction('add');
$ac->caption = 'Click to Add '; // the execBtn label in UserAction executor, i.e. perform the action.
$ac->description = 'Add New';  // The btn label that trigger the action, i.e will show modal.
```` 
I think did should be ok for now. Perhaps later on we could implement a better way to support custom button in ActionExecutor\UserAction::class and most propably via the 'ui' property of the genrica action itself.

File: src/Model.php [hunk 1]
```diff
@@ -416,7 +416,7 @@ public function init()
             'fields'  => true,
             'scope'   => UserAction\Generic::SINGLE_RECORD,
             'callback'=> 'save',
-            'ui'      => ['icon'=>'edit', 'button'=>[null, 'icon'=>'edit']],
+            'ui'      => ['icon'=>'edit', 'button'=>[null, 'icon'=>'edit'], 'execButton'=>new \atk4\ui\Button(['Save', 'blue'])],
```
romaninsh: can this be seed, not the class ? `execButton=>['Button', 'Save']`

DarkSide666: Yes only seed or nothing (see https://github.com/atk4/data/pull/490#issuecomment-562621547)

PR #498, opened 2020-01-06
Fix empty array condition

### LINKED ISSUE #496: Model addCondition with empty array value does not handle gracefully
When using `$model->addCondition('id', [])` the SQL select statement is created as `select * from table where id in ()` which causes SQL query to fail

DarkSide666:
fix #496 
rely on https://github.com/atk4/dsql/pull/165

### DIFF
--- a/locale/en/atk.php
+++ b/locale/en/atk.php
@@ -1,21 +1,21 @@
 <?php
 
 return [
-  'Field requires array for defaults'                                      => 'Field requires array for defaults',
-  'Field value can not be base64 encoded because it is not of string type' => 'Field value can not be base64 encoded because it is not of string type ({{field}})',
-  'Mandatory field value cannot be null'                                   => 'Mandatory field value cannot be null ({{field}})',
-  'Model is already related to another persistence'                        => 'Model is already related to another persistence',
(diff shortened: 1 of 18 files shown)


locale/en/atk.php
```diff
@@ -1,21 +1,21 @@
...
-    'other' => 'Test are {{count}}',
-  ],
-  'There was error while decoding JSON'             => 'There was error while decoding JSON',
-  'Unable to determine persistence driver from DSN' => 'Unable to determine persistence driver from DSN',
-  'Unable to serialize field value on load'         => 'Unable to serialize field value on load ({{field}})',
-  'Unable to serialize field value on save'         => 'Unable to serialize field value on save ({{field}})',
-  'Unable to typecast field value on load'          => 'Unable to typecast field value on load ({{field}})',
-  'Unable to typecast field value on save'          => 'Unable to typecast field value on save ({{field}})',
-    'Record with specified ID was not found'        => 'Record with specified ID was not found',
+    'Field requires array for defaults'                                      => 'Field requires array for defaults',
```
romaninsh: this might be just overwritten back...

romaninsh:
hard to spot changes beacuse of formatting differences.

PR #499, opened 2020-01-06
better implementation for addFields() and test cases

DarkSide666:
Now 'addFields' method will support a lot of different ways to pass arguments.
Also 2nd parameter in addFields now will always be properly used as default seed.

This is more complete implementation of https://github.com/atk4/data/pull/495.

All of these are now supported:
```
        $m->addFields(['name'], ['default' => 'anonymous']);
        $m->addFields([
            'last_name',
            'login' => ['default' => 'unknown'],
            'salary' => ['type'=>'money',CustomField::class, 'default' => 100],
            ['tax', CustomField::class, 'type'=>'money', 'default' => 20],
            'vat' => new CustomField(['type'=>'money', 'default' => 15]),
        ]);
```

Only questionable part left is - what to do if method don't understand how to resolve parameters.
Currently (like before) it simply ignores them, but maybe it should throw Exception instead. What do you think?

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1032,7 +1032,7 @@ class Model implements ArrayAccess, IteratorAggregate
      *
      * @return UserAction\Generic
      */
-    public function addAction($name, $defaults = []) : UserAction\Generic
+    public function addAction($name, $defaults = []): UserAction\Generic
     {
         if (is_callable($defaults)) {
             $defaults = ['callback'=>$defaults];
@@ -1060,7 +1060,7 @@ class Model implements ArrayAccess, IteratorAggregate
      *
      * @return array
      */
-    public function getActions($scope = null) : array
+    public function getActions($scope = null): array
     {
         return array_filter($this->actions, function ($action) use ($scope) {
             return !$action->system && ($scope === null || $action->scope === $scope);
@@ -1093,7 +1093,7 @@ class Model implements ArrayAccess, IteratorAggregate
      *
      * @return UserAction\Generic
      */
-    public function getAction($name) : UserAction\Generic
+    public function getAction($name): UserAction\Generic
     {
         return $this->_getFromCollection($name, 'actions');
     }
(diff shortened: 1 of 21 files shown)


src/Model.php
```diff
@@ -538,18 +538,25 @@ public function fieldFactory($seed = [])
...
     public function addFields($fields = [], $defaults = [])
     {
-        foreach ($fields as $field) {
-            if (is_string($field)) {
-                $this->addField($field, $defaults);
+        foreach ($fields as $key => $field) {
+            if (!is_int($key)) {
+                // field name can be passed as array key
+                $name = $key;
+                $seed = array_merge($defaults, is_array($field) ? $field : [$field]);
```
georgehristov: Maybe use the shorthand format?
`$seed = array_merge($defaults, (array) $field);`

georgehristov:
About throwing exception on not resolving parameters - one thing just popped in my mind: organizing this in a "PHP way" making it strict in dev environment and more forgiving in production environment.
This can be achieved using `$this->app->strict` bool property for instance. Throw exception on strict and ignore otherwise.
Same can be used for missing property check on `DIContainerTrait` for instance

src/Model.php
DarkSide666: sounds good, will update

DarkSide666:
Data repo is not and should not rely on existence of 'app'.
So Data repo will (if needed) always throw exceptions and it's a task for UI layer (including app) to catch exceptions and show or ignore them.

georgehristov:
Okay, but Model for instance has the `AppScopeTrait`?

DarkSide666:
Suggestion addressed, thanks!

About AppScopeTrait - I guess I will leave it as is at least for now.

romaninsh:
Please update documentation though.

DarkSide666:
OK will do that tonight and then merge. Thanks!

PR #504, opened 2020-01-21
Add microseconds persistence support for datetime/time types + fix normalization/cloning issue

mvorisek:
Very important to not loose time precision. All major database engines ignore the microseconds part if the target datatype has less or only integer seconds precision.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -165,7 +165,7 @@ class Field implements Expressionable
     /**
      * Persisting format for type = 'date', 'datetime', 'time' fields.
      *
-     * For example, for date it can be 'Y-m-d', for datetime - 'Y-m-d H:i:s' etc.
+     * For example, for date it can be 'Y-m-d', for datetime - 'Y-m-d H:i:s.u' etc.
      *
      * @var string
      */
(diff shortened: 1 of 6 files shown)


mvorisek:
PR complete, ready to be merged.

src/Field.php
```diff
@@ -409,11 +409,23 @@ public function toString($value = null)
...
+                        if ($this->type === 'date') {
+                            $format = $dateFormat;
+                        } elseif ($this->type === 'time') {
+                            $format = $timeFormat;
+                        } else {
+                            $format = $dateFormat.'\T'.$timeFormat.'P'; // ISO 8601 format 2004-02-12T15:19:21+00:00
+                        }
+
+                        return $v->format($format);
+                    } else {
```
DarkSide666: no need for `else` here - just `return` after if condition :)

src/Persistence/SQL.php
```diff
@@ -533,20 +533,27 @@ public function _typecastLoadField(Field $f, $value)
...
+                    }
+                }
 
                 // datetime only - set from persisting timezone
                 if ($f->type == 'datetime' && isset($f->persist_timezone)) {
                     $v = $dt_class::createFromFormat($format, $v, new $tz_class($f->persist_timezone));
-                    if ($v === false) {
-                        throw new Exception(['Incorrectly formatted datetime', 'format' => $format, 'value' => $value, 'field' => $f]);
+                    if ($v !== false) {
+                        $v->setTimeZone(new $tz_class(date_default_timezone_get()));
```
DarkSide666: should be `setTimezone` instead of `setTimeZone`
https://www.php.net/manual/en/datetime.settimezone.php

mvorisek: Fixed in the another PR https://github.com/atk4/data/pull/505 - merge that PR first.

mvorisek:
PR updated and ready to be merged. Please see comments as another PR should be merged first.

PR #505, opened 2020-01-21
Accept any DateTimeInterface impl. for datetime

mvorisek:


src/Field.php
```diff
@@ -337,11 +337,15 @@ public function normalize($value)
                 } elseif (is_string($value)) {
                     $value = new $class($value);
                 } elseif (!$value instanceof $class) {
-                    if (is_object($value)) {
-                        throw new ValidationException(['must be a '.$f->type, 'class' => $class, 'value class' => get_class($value)]);
+                    if ($value instanceof \DateTimeInterface) {
+                        $value = new $class($value->format('Y-m-d H:i:s.u'), $value->getTimeZone());
```
DarkSide666: should be `getTimezone` not `getTimeZone`.
https://www.php.net/manual/en/datetime.gettimezone.php

src/Persistence/SQL.php
```diff
@@ -453,7 +453,7 @@ public function _typecastSaveField(Field $f, $value)
             $dt_class = isset($f->dateTimeClass) ? $f->dateTimeClass : 'DateTime';
             $tz_class = isset($f->dateTimeZoneClass) ? $f->dateTimeZoneClass : 'DateTimeZone';
 
-            if ($v instanceof $dt_class) {
+            if ($v instanceof $dt_class || $v instanceof \DateTimeInterface) {
```
DarkSide666: Few lines below we call `setTimezone` method, but DateTimeInterface doesn't support that.
https://www.php.net/manual/en/class.datetimeinterface.php
Probably need some additional condition in line 461 ( maybe `&& $v instanceof $dt_class`)

src/Field.php
mvorisek: Good catch, I will fix it promtly. I have copied this clever code from https://github.com/atk4/data/blob/37d1722ac47f53179677e9509f5e484fd3f8ed03/src/Persistence/SQL.php#L555 , so I will fix it there too.

src/Persistence/SQL.php
mvorisek: That is correct, there was also a mistake when the class was immutable (introduced in the past). Wdyt about the new fix which fixes both?

mvorisek:
Feedback resolved, can this PR be merged?

PR #507, opened 2020-01-23
Introduce isSystem method

georgehristov:
For method consistency


romaninsh:
I'd say "system" is a private property that affects isReadable() and isVisible() methods. UI shouldn't know about "system" flag and instead should rely and work with the above existing methods. 

Otherwise we run into problem of "should form display field if isSystem=true and isEditable=true ?" and this logic would bleed into multiple UI elements rather than staying in one place.

PR #509, opened 2020-02-07
use property to define field seed properties

georgehristov:
- use single generic `Field::getSeed` method that checks seed properties for each field type
- use class constant for field type mapping
- minor changes in `Numeric` field property names for simplification
- backward compatible use of `$fx` property in `Callback` field instead of `$expr`
- add field tests


romaninsh:
is this up-to-date? could you merge develop in?

romaninsh:
OH i see you PR this into /number-field

DarkSide666:
I wonder if number-field branch is at all needed.... not sure what it was.

georgehristov:
It is separating the field logic in dedicated classes and I believe it is necessary.
I built on the direction of this branch and created a PR for review and consideration.

georgehristov:
Reference https://github.com/atk4/data/wiki/ADP1904---Typecasting-and-serialization-cleanup

PR #526, opened 2020-03-22
Exception for required field is more descriptive

mlazovla:
Add name of field to Exception. Error message "Must not be null" exception is not descriptive.


romaninsh:
my design.

PR #529, opened 2020-03-25
WIP Use MySQL 8.0

DarkSide666:
command

https://github.com/atk4/data/runs/534570568

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -21,7 +21,6 @@ jobs:
     services:
       mysql:
         image: mysql:8.0
-        command: --default-authentication-plugin=mysql_native_password
         env:
           MYSQL_ROOT_PASSWORD: password
           MYSQL_DATABASE: db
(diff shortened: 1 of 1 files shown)


DarkSide666:
Well, I have no more ideas what else to try to force MySQL 8.0 use old style authentication mechanism.

mvorisek:
@DarkSide666 And do we need the "old style authentication mechanism" at all? I will look at this, but no sooner than in a week, I am very busy now.

DarkSide666:
Ok, thanks!

mvorisek:
@DarkSide666 Solving this seems to be complicated as docker `CMD` can not be specified in Github Action service. I tried to fix it using custom `ENTRYPOINT`, but I was not able to specify even the default file.

The issue seems to be related with https://github.com/atk4/image which uses lighweight Alpine image without the new `caching_sha2_password` plugin support.

romaninsh:
Oh we could include that support there perhaps?

mvorisek:
@romaninsh Yes - the `atk4/image` docker image installs `mysql-client`:
https://github.com/atk4/image/blob/master/Dockerfile#L3
but this client can not be used to interact with the new MySQL 8 as it does not have `caching_sha2_password` support.

Do you have any idea how to fix this without changing the Alpine image to Debian?

mvorisek:
fixed/solved by https://github.com/atk4/data/pull/645

PR #532, opened 2020-03-26
Fix LIKE conditions for Array and SQL persistences

### LINKED ISSUE #530: Conditions on boolean column type work unexpectedly
When boolean field is added to a model like:
```
$m->addField('payment_is_paid', ['type' => 'boolean', 'required' => true]);
```

Then when condition like:
```
$q = '20201093';
$cond = [];
$cond[] = ['payment_is_paid', 'like', '%' . $q . '%'];
$m->addCondition($cond);
```
is added it filters all records with the flag set to 0, here is a part of the generated query:
```
... `inv`.`payment_is_paid` like 0 ...
```
instead of:
```
... `inv`.`payment_is_paid` like '%20201093%' ...
```
<br><br>Check the behaviour of integer and date(time) types too.

DarkSide666:
fix #530

### DIFF
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -38,7 +38,7 @@ class Iterator
         $this->generator = new \CallbackFilterIterator($this->generator, function ($row) use ($field, $value) {
 
             // skip row. does not have field at all
-            if (!isset($row[$field])) {
+            if (!array_key_exists($field, $row)) {
                 return false;
             }
 
@@ -66,12 +66,12 @@ class Iterator
         $this->generator = new \CallbackFilterIterator($this->generator, function ($row) use ($field, $value) {
 
             // skip row. does not have field at all
-            if (!isset($row[$field])) {
+            if (!array_key_exists($field, $row)) {
                 return false;
             }
 
             $clean_value = trim(trim($value), '%');
-            // the row field exists check the position of th "%"(s)
+            // the row field exists check the position of the "%"(s)
             switch ($value) {
                 // case "%str%"
                 case substr($value, -1, 1) == '%' && substr($value, 0, 1) == '%':
@@ -85,6 +85,9 @@ class Iterator
                 case substr($value, 0, 1) == '%':
                     return substr($row[$field], -strlen($clean_value)) === $clean_value;
                     break;
+                // full match
+                default:
+                    return $row[$field] == $clean_value;
             }
 
             return false;
--- a/src/Model.php
+++ b/src/Model.php
@@ -1178,7 +1178,7 @@ class Model implements ArrayAccess, IteratorAggregate
      *  ->addCondition('my_field', '!=', $value);
      *  ->addCondition('my_field', 'in', [$value1, $value2]);
      *
-     * Second argument could be '=', '>', '<', '>=', '<=', '!=', 'in' or 'like'.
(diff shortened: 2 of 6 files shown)


src/Model.php
```diff
@@ -1228,7 +1228,7 @@ public function addCondition($field, $operator = null, $value = null)
         $f = is_string($field) ? $this->getField($field) : ($field instanceof Field ? $field : false);
         if ($f) {
             if ($operator === '=' || func_num_args() == 2) {
-                $v = $operator === '=' ? $value : $operator;
+                $v = ($operator === '=' ? $value : $operator);
```
mvorisek: brackets are not needed

src/Model.php
```diff
@@ -1178,7 +1178,7 @@ public function removeAction($name)
      *  ->addCondition('my_field', '!=', $value);
      *  ->addCondition('my_field', 'in', [$value1, $value2]);
      *
-     * Second argument could be '=', '>', '<', '>=', '<=', '!=' or 'in'.
+     * Second argument could be '=', '>', '<', '>=', '<=', '!=', 'in' or 'like'.
```
mvorisek: fix for regexp as well

src/Action/Iterator.php
```diff
@@ -66,12 +66,12 @@ public function like($field, $value)
         $this->generator = new \CallbackFilterIterator($this->generator, function ($row) use ($field, $value) {
 
             // skip row. does not have field at all
-            if (!isset($row[$field])) {
+            if (!array_key_exists($field, $row)) {
                 return false;
             }
 
             $clean_value = trim(trim($value), '%');
```
mvorisek: The trims are not compatible with the swith below.

src/Persistence/SQL.php
```diff
@@ -441,7 +440,8 @@ public function initQueryConditions(Model $m, Query $q): Query
                 }
                 $q->where($cond[0], $cond[1]);
             } else {
-                if ($cond[0] instanceof Field) {
+                // "like" condition do not need typecasting to field type!
+                if ($cond[0] instanceof Field && strtolower($cond[1]) != 'like') {
```
mvorisek: Use `in_array` and fix for regexp as well.

src/Model.php
DarkSide666: I know. Added them for readability. I really don't like when there are to many `=` in one line :)

mvorisek: Once PSR-12 CS rules will be implemented I think they will be removed :)

src/Action/Iterator.php
DarkSide666: Why not? Trim gives us clean value which is used below.

src/Persistence/SQL.php
DarkSide666: ok, fixed

src/Model.php
DarkSide666: ok fixed

src/Action/Iterator.php
mvorisek: As the `substr`s in the switch are checking in untrimmed string.

DarkSide666: Ah, right. Fixed in https://github.com/atk4/data/pull/538

Issue #506, opened 2020-01-22
Add BLOB type and super-lazy load this field type by default


mvorisek:

mvorisek:
Lazy loading should be easily solved by introducing some value wrapper class which will have getValue() and setValue() methods.

By default the column must not be queried or queried only like `<col> IS NOT NULL` (better than `select 1` as doing so the column name will not be used thus not validated).

The value will be persisted on save only if changed.

And prefeably with `UPDATE` one by one, otherwise MySQL `max_allowed_packet` can be reached easily in normal setups (set to `64MB` by default for MySQL 8.0)

@DarkSide666 Wdyt?

DarkSide666:
Lazy loading of field values would really be good thing to have for all (non-system) fields, but especially useful for BLOBs and calculated (expression) fields. It's quite common to have multiple, slow-querying expression or calculated fields in your model which most of the time are not used in all app pages/views, but still it's always loaded. Currently you can limit that with `onlyFields()`...

Speaking about value wrapper class - probably that would be good to have in this case, but what I am a bit worried is that lately we're making much to much BC-breaking changes in all repositories and as proved in past - this makes Agile toolkit not upgradable in projects. Developers are stuck with old Agile toolkit versions because they can't upgrade to newer ones without refactoring a lot in their projects. That's not good. Especially for minor toolkit versions.

mvorisek:
> Lazy loading of field values would really be good thing to have for all (non-system) fields

Yes, then this should be called "lazy" and this issue is for "super-lazy" fields :):
- non-system fields need to be normally loaded in bulk when iterating (for ex. for a Grid),
- but "super-lazy" fields always one by one (for ex. files to download).

I removed the "help wanted" label and will write a POC in the next days.

We are using DB for files instead of FS which is great for having to backup only one data source and also for ACID. I belive we can very easily implement then file uploads in `atk4/ui`.

mvorisek:
related with https://github.com/atk4/data/issues/118

PR #535, opened 2020-03-28
Fix LIKE conditions - fix fix

### LINKED ISSUE #530: Conditions on boolean column type work unexpectedly

mvorisek:
Fixes #530 - issue
Fixes #532 - original PR

### DIFF
--- a/src/Persistence/SQL.php
+++ b/src/Persistence/SQL.php
@@ -410,7 +410,7 @@ class SQL extends Persistence
             // count($cond) == 1, we will pass the only
             // parameter inside where()
 
-            if (count($cond) == 1) {
+            if (count($cond) === 1) {
 
                 // OR conditions
                 if (is_array($cond[0])) {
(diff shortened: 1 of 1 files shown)


georgehristov:
@mvorisek you have implemented this for `OR` conditions only. It should get extended to the `AND` conditions as well.
I am working on the field classes now and will include this to  be resolved but a quickfix might be necessary

mvorisek:
> 
> 
> @mvorisek you have implemented this for `OR` conditions only. It should get extended to the `AND` conditions as well.

Really? I am convinced that I am actually the one who is fixing it - see PR description, this is a fix :)

> I am working on the field classes now and will include this to be resolved but a quickfix might be necessary

👍 I am planning to help with issue #506 - @georgehristov, will you be solving super-lazy (see that issue why not only "lazy") too?

PR #538, opened 2020-03-29
Fix value trim

DarkSide666:
Fix value trim

### DIFF
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -70,7 +70,8 @@ class Iterator
                 return false;
             }
 
-            $clean_value = trim(trim($value), '%');
+            $value = trim($value);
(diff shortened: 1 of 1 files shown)


src/Action/Iterator.php
```diff
@@ -70,7 +70,8 @@ public function like($field, $value)
                 return false;
             }
 
-            $clean_value = trim(trim($value), '%');
+            $value = trim($value);
+            $clean_value = trim($value, '%');
             // the row field exists check the position of the "%"(s)
             switch ($value) {
```
mvorisek: This PR is fine. But what is actually the purpose of `switch` (vs. `if`) when all cases are expressions (i.e. not values for which the `switch` is designed)?

Issue #497, opened 2020-01-04
PDO: addCondition('field', null) fails on statement preparation


georgehristov:
Environment
MariaDB 10.3
PHP 7.3
atk4/dsql: develop

When one of the conditions added to the model includes field value to be `null` `PDOException` is thrown on [this line](https://github.com/atk4/dsql/blob/7766d5523eeee07aaff34860f08d064e206941cb/src/Expression.php#L562)

`SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '?' at line 1`

Code:
`$model->addCondition('position', null);`

Query string generated by dsql is:
`select count(*) from `commondata` where `parent` = :a and `position` is :b`

DarkSide666:
OK, I will need more info here to reproduce this.
Condition with null value works fine everywhere and in all tests, so it's something else which causes this.
How do you set up your DB connection?
Why do you think `show tables like` is somehow related to this?

Please try to debug this further and let me know what you find (or make PR).

georgehristov:
I spent quite some time on trying o figure this out and came to the conclusion it is related to PDO or MariaDB in the case when there is a word (and not '=') before the parameter.
DB connection is setup as Laravel singleton [here](https://github.com/x-systems/epesi-core/blob/8f2167912f4200c28bed164a48d906e2c431a4cc/src/Providers/EpesiServiceProvider.php#L58) and assigned to Model [here](https://github.com/x-systems/epesi-core/blob/8f2167912f4200c28bed164a48d906e2c431a4cc/src/Data/Model.php#L16)
If no one else has this issue will have to look into it further.

DarkSide666:
Maybe it's somehow related to this line there https://github.com/atk4/dsql/blob/develop/src/Expression.php#L571 ?

georgehristov:
I was looking into it as well as some sources suggest using `\PDO::PARAM_INT` for NULL vales as well but code execution on my side does not reach that line. It fails on
https://github.com/atk4/dsql/blob/7766d5523eeee07aaff34860f08d064e206941cb/src/Expression.php#L562

But your hint about the DB connection is the only thing I can further check. Will try with direct DB connection settings.

DarkSide666:
@georgehristov any news on this?

georgehristov:
I dug into it more.

[...]

PR #540, opened 2020-03-29
introduce sql null condition test

georgehristov:
Related to #497


tests/ConditionSQLTest.php
```diff
@@ -48,6 +48,24 @@ public function testBasic()
...
+                2 => ['id' => 2, 'name' => 'Sue', 'gender' => 'F'],
+                3 => ['id' => 3, 'name' => 'Null', 'gender' => null],
+            ], ];
+        $this->setDB($a);
+
+        $m = new Model($this->db, 'user');
+        $m->addFields(['name', 'gender']);
+
+        $mm = clone $m;
+        $mm->addCondition('gender', null)->tryLoadAny();
```
PhilippGrashoff: Could make sense to use loadAny() here. If the condition has some failure and no matching record is found, an exception will be thrown which will contain more useful information than the information that the assertion failed in the next line.

georgehristov: Thanks but I improved the test a bit in another direction

PR #541, opened 2020-03-30
introduce shorthand for setting conditions on referenced models

georgehristov:
Introduce setting conditions on referenced models:
```$contact->addCondition('company/country', 'US');```

where 'company' is the name of the reference

This will limit scope of $contact model to contacts whose company country field is set to 'US'

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1236,7 +1236,7 @@ class Model implements ArrayAccess, IteratorAggregate
 
         if (is_string($field)) {
             // shorthand for adding conditions on references
-            // use chained reference names separated by dot "."
+            // use chained reference names separated by "/"
             if (stripos($field, '/') !== false) {
                 $references = explode('/', $field);
 
--- a/src/Model.php
+++ b/src/Model.php
@@ -1186,12 +1186,12 @@ class Model implements ArrayAccess, IteratorAggregate
      *  ->addCondition($expr);
      *
      * Conditions on referenced models are also supported:
-     *  $contact->addCondition('company.country', 'US');
+     *  $contact->addCondition('company/country', 'US');
      * where 'company' is the name of the reference
      * This will limit scope of $contact model to contacts whose company country is set to 'US'
      *
      * Using # in conditions on referenced model will apply the condition on the number of records:
-     * $contact->addCondition('tickets.#', '>', 5);
+     * $contact->addCondition('tickets/#', '>', 5);
      * This will limit scope of $contact model to contacts that have more than 5 tickets
      *
      * To use those, you should consult with documentation of your
@@ -1237,8 +1237,8 @@ class Model implements ArrayAccess, IteratorAggregate
         if (is_string($field)) {
             // shorthand for adding conditions on references
             // use chained reference names separated by dot "."
-            if (stripos($field, '.') !== false) {
-                $references = explode('.', $field);
+            if (stripos($field, '/') !== false) {
+                $references = explode('/', $field);
 
                 $field = array_pop($references);
 
--- a/src/Model.php
+++ b/src/Model.php
@@ -1184,11 +1184,11 @@ class Model implements ArrayAccess, IteratorAggregate
      * There are also vendor-specific expression support:
      *  ->addCondition('my_field', $expr);
      *  ->addCondition($expr);
-     *  
+     *
      * Conditions on referenced models are also supported:
      *  $contact->addCondition('company.country', 'US');
      * where 'company' is the name of the reference
-     * This will limit scope of $contact model to contacts whose company country is set to 'US' 
+     * This will limit scope of $contact model to contacts whose company country is set to 'US'
      *
      * To use those, you should consult with documentation of your
      * persistence driver.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by DarkSide666** [comment]
Theoretically model fields can have dots in their names... :(

**Comment by georgehristov** [comment]
@DarkSide666 unfortunately does not work as we discussed under https://github.com/atk4/ui/issues/965
Of course this is for atk4/ui but I suggest dots to not be used in atk4/data as well for field names as well

**Comment by DarkSide666** [comment]
Right. Then we need to add restriction to use dot in field names in `addField`, `addExpression` etc. methods to be sure.

**Comment by georgehristov** [comment]
Okay, then I will do that and also update docs for the setting conditions on references shorthand syntax

**Comment by georgehristov** [comment]
@DarkSide666 I decided to change the separator to "/". Now also docs are updated.

File: src/Model.php [hunk 1]
```diff
@@ -1225,14 +1234,45 @@ public function addCondition($field, $operator = null, $value = null)
...
+                foreach ($references as $link) {
+                    $model = $model->refLink($link);
+                }
+
+                $args = func_get_args();
+
+                // '#' will apply condition directly on the record count (has # referenced records)
+                // otherwise applying condition on the referenced model field (has referenced records where)
+                if ($field !== '#') {
+                    $model->addCondition(...$args);
```
DarkSide666, line 1255: Shouldn't we use `($field, $operator, $value)` here instead? Otherwise it will pass field as `company/country` for example, but should pass only extracted field name `country`.

For now it's OK, but maybe think about how we could use `SELECT * FROM t WHERE EXISTS (...)` syntax in this case (for SQL). This would be better for performance than counting records.

src/Model.php
```diff
@@ -1225,14 +1234,45 @@ public function addCondition($field, $operator = null, $value = null)
...
+                    $args[1] = '>';
+                    $args[2] = 0;
+                }
+                $args[0] = $model->action('count');
+
+                $this->conditions[] = $args;
```
DarkSide666, line 1261: Shouldn't we use `return $this->addCondition(...$args);` here?

tests/LookupSQLTest.php
```diff
@@ -68,8 +68,8 @@ public function init()
             ->withTitle()
             ->addFields(['country_code'=>'code', 'is_eu']);
 
-        $this->hasMany('Friends', new LFriend())
-            ->addField('friend_names', ['field'=>'friend_name', 'concat'=>',']);
+//         $this->hasMany('Friends', new LFriend())
+//             ->addField('friend_names', ['field'=>'friend_name', 'concat'=>',']);
```
DarkSide666, line 72: why these were commented out?

DarkSide666 (changes requested):

src/Model.php
georgehristov: `func_get_args` takes the last value of the argument variable. So in this case we change `$field` to the field name of the reference a few lines above.
`func_get_args` is used intentionally as it takes only the number of arguments passed onto the method (after last changes to the variables). This is important as `$operator` can also be the `$value` so if you pass `($field, $operator, $value)` you always pass on 3 arguments and the shorthand for `=` operator will not work.
What I noticed also is that `func_get_args` takes only the number of arguments originally passed onto the method. So if you have only `($field, $operator)` and you set `$value` to something it will not get included in the `func_get_args` array. That is why I used `$args` variable.
I was also considering `EXISTS` syntax as I know it is much faster. Will think of some field syntax for that: maybe using '?' for exists and '!' does not exist?

src/Model.php
georgehristov: Yes, can be done. Will switch it over.

tests/LookupSQLTest.php
georgehristov: They are not used in the test. Some more lines in the test are commented out related to this model relation.
When running the test on referenced models with these lines SQL is throwing exception for missing field 'field_names'.
I did not look into it deeper at this stage as I believe it is not related to the shorthand syntax.

georgehristov: `Column not found: 1054 Unknown column 'friend_name' in 'field list'`

```
SELECT 
	`user`.`id`,
	`user`.`name`,
	`user`.`is_vip`,
	`user`.`country_id`,
	(SELECT `name` FROM `country` `c` WHERE `id` = `user`.`country_id`) `country`,
	(SELECT `code` FROM `country` `c` WHERE `id` = `user`.`country_id`) `country_code`,
	(SELECT `is_eu` FROM `country` `c` WHERE `id` = `user`.`country_id`) `is_eu`,
	(SELECT GROUP_CONCAT(`friend_name` SEPARATOR ',') `friend_name`
		FROM `friend` `F`
	WHERE `user_id` = `user`.`id`) `friend_names`
FROM `user`
WHERE 
(SELECT COUNT(*) FROM `country` `c` WHERE `id` = `user`.`country_id` AND `code` = 'LV') > 0
```

src/Model.php
DarkSide666: Wow. I didn't know that. Good catch !

DarkSide666:
I think `addCondition('company/tickets/?');` could be great syntax for `EXISTS` queries.

Not so sure about `!` for `NOT EXISTS`. Maybe `addCondition('company/tickets/?', true or false);` (true by default), but maybe `!` is also good - `addCondition('company/tickets/!');`

How this goes together with new `Scope` class? Don't need any changes in this PR? Which PR should go first?

georgehristov:
Scope branch is a branch of this one.
The functionality is implemented in Scope (including toWords option) and not in Model any more so we can directly merge Scope (when completed).

As example `$user->addCondition('company/tickets/?');` would be 'translated' by Scope to
`User that has reference company that has reference tickets where any records exist`

romaninsh (approved):
This is so awesome!!

georgehristov:
Scope branch is ready so please review it and merge when agreed.
This functionality will then be available (still working on the 'having' usage, presently 'count' is used)

romaninsh:
what should we do with this then? close?

georgehristov:
Yes, this one we close and we keep only scopes

PR #543, opened 2020-04-01
Fix refactorability for "owner"

mvorisek:
Important for quick navigation and refactoring.

### DIFF
--- a/src/Field_SQL.php
+++ b/src/Field_SQL.php
@@ -9,6 +9,8 @@ use atk4\dsql\Expressionable;
 
 /**
  * Class description?
+ *
+ * @property Join\SQL $join
  */
 class Field_SQL extends Field implements Expressionable
 {
--- a/src/Join/SQL.php
+++ b/src/Join/SQL.php
@@ -9,6 +9,9 @@ use atk4\data\Model;
 
 /**
  * Join\SQL class.
+ *
+ * @property \atk4\data\Persistence\SQL $persistence
+ * @property SQL $join
  */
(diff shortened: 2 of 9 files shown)


src/Field_SQL.php
```diff
@@ -9,6 +9,8 @@
 
 /**
  * Class description?
+ *
+ * @property Join\SQL $join
```
mvorisek, line 13: Is this assumtion right, or can be here non-SQL Join?

src/Join/SQL.php
```diff
@@ -9,6 +9,9 @@
 
 /**
  * Join\SQL class.
+ *
+ * @property \atk4\data\Persistence\SQL $persistence
```
mvorisek, line 13: Is this assumtion right, or can be here non-SQL Persistence?

src/Join/SQL.php
```diff
@@ -9,6 +9,9 @@
...
+ * @property SQL $join
```
mvorisek, line 14: Is this assumtion right, or can be here non-SQL Join?

src/Model.php
```diff
@@ -21,6 +21,8 @@
 
 /**
  * Data model class.
+ *
+ * @property Field[]|Reference[] $elements
```
mvorisek, line 25: Is this assumtion right?

DarkSide666: Currently yes, but in future I think more element types could go there. Maybe even scopes/conditions, joins etc. Not sure, but good for now.

PR #545, opened 2020-04-08
Use null coalescing operator instead of isset

mvorisek:


mvorisek:
There is a small bug - see https://github.com/atk4/data/runs/570500993

How to solve it?

src/Persistence.php
```diff
@@ -40,7 +40,7 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
         // Process DSN string
         $dsn = \atk4\dsql\Connection::normalizeDSN($dsn, $user, $password);
 
-        $driver = isset($args['driver']) ? strtolower($args['driver']) : $dsn['driver'];
+        $driver = strtolower($args['driver'] ?? $dsn['driver']);
```
DarkSide666, line 43: This is slight actual change, but I think that's ok.

mvorisek: Step by step improvement ;-)

mvorisek:
> 
> 
> Overall looks good. I will check why value type is cast in this case. I guess `??` does that somehow.

No, see the 2nd reverted commit - the issue is that the return type is now forced to string (by function header). Probably the export function should not use toString for numerical fields.

DarkSide666:
> No, see the 2nd reverted commit - the issue is that the return type is now forced to string (by function header). Probably the export function should not use toString for numerical fields.

Oh I see. Well... I think `toString` should return string always, so maybe we should change tests anxd cast everything to string. I think it will not affect anything important because it still will be `var_dump("5"+3);` int 8

mvorisek:
@DarkSide666 Done, same assumption on my side, but the test data come from `export` method and I think this method should preserve the integer type.

Any idea how to fix this properly, i.e. without changing the test data?

DarkSide666:
No idea, but I think it's OK this way.
If we want typecasted export, then there is 3rd parameter of export method. Otherwise it's not typecasted - so doing toString.
These `vat_rate_id` fields are not actually id fields. They are not defined as hasOne fields I think.

PR #549, opened 2020-04-09
take into account model limit when counting

georgehristov:
`Persistence\SQL` does not take into account the model limit when counting records.
This is inconsistent with what `Persistence\Array_` is performing.

If Model has SQL persistence and a limit (0, 3) when performing 'count' action all model records (matching the model scope) will be counted and not only the ones within the limit.

SQL does not support 
`select count(*) from xx limit 0, 3` 
so this fix converts it to
`select count(*) from (select * from xx limit 0, 3)`
if model has a limit set

If this is not desired behavior we can consider removing the limit affecting count in `Persistence\Array_` so both act consistently

### DIFF
--- a/src/Persistence/SQL.php
+++ b/src/Persistence/SQL.php
@@ -655,14 +655,14 @@ class SQL extends Persistence
 
             case 'count':
                 $this->initQueryConditions($m, $q);
+                $this->setLimitOrder($m, $q);
                 $m->hook('initSelectQuery', [$q]);
-                if (isset($args['alias'])) {
-                    $q->reset('field')->field('count(*)', $args['alias']);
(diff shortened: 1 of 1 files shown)


romaninsh (changes requested):
count() should not take limit into account. Limit would be applied on top of count anyway and wouldn't affect it. Can you update PR description to explain what is the problem and how is this fix solving it.

georgehristov:
@romaninsh description updated

romaninsh:
Thanks - and yes i think we should do the alternative - update Array_ so it's ignoring limit when counting. SQL generally is the "reference" implementation and Array_ is modelled after it.

PR #550, opened 2020-04-09
Remove old Travis files

mvorisek:
If needed anytime, they can be restored from history, but having old/ununsed files in repo is misleading.


romaninsh:
Is that fully removing travis? I don't see it in checks above.

mvorisek:
@romaninsh Travis is already disabled, it only removes the old/unmaintained/untested files. Legit?

DarkSide666 (changes requested):
Same as in DSQL repo - I agree to remove `.old.travis.yml` file, but would like to keep phpunit config files as they are because they can be used for testing locally in localhost (host=localhost).

mvorisek:
@DarkSide666 Yes, but we should be able to map the DBs to localhost using `ports` https://help.github.com/en/actions/configuring-and-managing-workflows/about-service-containers#mapping-docker-host-and-service-container-ports

DarkSide666:
If you could update this PR with that solution and if that will work on github docker and also in my localhost when simply running `phpunit configuration=phpunit-whatever.xml` then that would be great! I'm not good in all this Docker stuff :)

DarkSide666:
Tell me when this is ready for review. It looks that this could do the trick.

mvorisek:
> 
> 
> Tell me when this is ready for review. It looks that this could do the trick.

Need help with it, the fix currently does not work - see test results for MySQL.

DarkSide666:
tests work fine on localhost

mvorisek:
On localhost yes, and this PR normalizes them that you should need only one login/user for all atk4 repos.

But not for Actions... https://github.com/atk4/data/pull/550/checks?check_run_id=578841513#step:10:6

DarkSide666:
Few notes to test:
1. in php.ini `pdo_mysql.default_socket => no value => no value`. Maybe need to set default_socket here somehow?
2.

phpunit-mysql-workflow.xml
```diff
@@ -1,25 +0,0 @@
-<phpunit colors="true" bootstrap="vendor/autoload.php" printerClass="atk4\core\PHPUnit_AgileResultPrinter">
-    <php>
-        <var name="DB_DSN" value="mysql:dbname=db;host=mysql" />
```
romaninsh, line 3: host=mysql 

this was missing from `phpunit-mysql.xml`

romaninsh:
@DarkSide666 @mvorisek - this now connect and passes MySQL, you forgot to use "host=mysql".

Also - I have added a better local testing suite:

```
docker-compose run tests
```

This will download and install all the same images and environment as used by Github and run the tests.

romaninsh (approved):
looking good!

mvorisek:
`localhost` is/was right. You should be able to run the tests with pure `phpunit` without composer on localhost. And for docker tests the database ports from services should be mapped therefore to localhost/main docker process.

The docker for local tests seems even a little too much for me as `mysql` database has almost every developer installed and has debug tools for it - like phpmyadmin etc. - that are also important for development and not that easily accessible when running thru docker.

In total, I do not see any advantage having a docker for local testing. The only advantage I see is having PostgreSQL tested locally as well, but at least I do not care about this DB and the GH Actions tests should be enought. For me it does not justify the docker local setup that need to be maintained to be working/in sync with the Github Actions tests.

romaninsh:
@mvorisek you can still use localhost - as explained to dark - for local testing. However if you wish to replicate "GithubAction" environment EXACTLY and debug a pipeline - you can use "docker-compose" now. It was able to help me debug connection issues easily.

workflow files used "mysql/pgsql" as a host when you removed them you didn't transfer that setting and that caused pipelines to break.

mvorisek:
@romaninsh This is wrong. Why should the standard local workflow not work out of box? In docker, we can use `localhost` and make it 100% compatible out of the box.

romaninsh:
If you wanna play with port routing - create separate PR - this one was only about cleaning up the .travis files, lets keep the host system as works now.

mvorisek:
No, the original `phpunit-*.xml` files used localhost and they should continue to do so.

The local docker should be maybe even reverted as it introduce code duplication and it should be solved with proper tools like: https://github.com/nektos/act

romaninsh:
Github Actions were NOT using original `phpunit-*.xml` files. That's how "develop" is working now. You replaced those yourself: https://github.com/atk4/data/pull/550/commits/666ed654f28b52523f479aa7008837e3378c02b0#diff-4794c8638bc5e5ec555e780ee618f61fL58

romaninsh:
also - nektos/act does not support services it seems: https://github.com/nektos/act/issues/173

mvorisek:
Yes, but my original ask question was how to unify it to support localhost as usual (which also @DarkSide666 mentioned to be using it) and make docker working with it too.

romaninsh:
@mvorisek I tried it when I initially implemented GitHub actions and while .Travis supported two options "local port" and "remote host" - GitHub action being a brand new system went for remote host only. 

Maybe I have missed something: https://help.github.com/en/actions/configuring-and-managing-workflows/about-service-containers

mvorisek:
We can always connect the localhost and the remote host port with some linux command if not natively supported by GitHub Actions.

mvorisek:
`mysql` service probably listen only on localhost but this localhost is of course different.

If not, maybe this can also help https://www.google.com/search?q=%22--net%3Dhost%22

mvorisek:
Should be fixed here: https://github.com/atk4/image/issues/1 , there is currently no way to modify `CMD`/`mysqld` command in Github Actions Services.

mvorisek:
@DarkSide666 it seems that the `mysql` service can not listen easily on localhost (without socat, ncat, etc.). But this is common also for other packages - is it ok to keep there `mysql` for GH Actions and local tests? For local tests, you have the option to set `mysql` in local `hosts` which solves it for other packages too (almost all packages use `mysql` service name).

If ok, I will propose this for every other atk4 package and unify the credentials like `atk4_test` user name etc.

DarkSide666:
Looks like I have no other option :)
Yeah I guess it's fine. Most important is to unify between all repos.

mvorisek:
@DarkSide666 Can you approve it?

What workflow do you use to change something across all repos? What about some monorepo that will be always synchronized between all repos and for these changes only one PR will be needed? And all - hundreds - tests will run at once :)

DarkSide666 (approved):

DarkSide666:
Approved finally.

About changing in all repos - currently we was doing that by just manually creating PRs in each of them. It's a bit more work, but on the other hand - also no need to overcomplicate workflow (like killing flies with bazooka) :)

.github/workflows/unit-tests.yml
```diff
@@ -60,10 +63,10 @@ jobs:
         run: vendor/bin/phpunit --configuration phpunit.xml --coverage-text --exclude-group dns
 
       - name: MySQL Testing
-        run: vendor/bin/phpunit --configuration phpunit-mysql-workflow.xml --exclude-group dns
+        run: vendor/bin/phpunit --configuration phpunit-mysql.xml --exclude-group dns || echo 'debug'
```
mvorisek, line 66: @DarkSide666 Upps, please fix (remove ` || echo 'debug'`) this immediatelly, I can not do it easily :(

DarkSide666: Done.

PR #551, opened 2020-04-09
Add Scope class - new filtering/where impl. for Model

### LINKED ISSUE #524: Abstracting of Model conditions via Crits class
Presently Model conditions and Fields are quite tightly coupled with the underlying persistence.
This results in necessity to define the Model condition at the lowest level.
I propose to introduce a class that will abstract this link. The idea is based on the way conditions are handled in Epesi and I believe this will enhance atk4/data.
I have developed already the refactored classes for epesi and if concept is accepted I can work on integrating them with atk4/data (as I really need this functionality).

The concept is as follows:
- a class called `Crits` contains the conditions for the Model (scope).
- when `Crits` are applied to the Model it is equivalent to applying several conditions to the Model.
- `Crits` support multi-level AND-OR grouping
- Crit value can be 'abstract', e.g. we use a text field to store multiselect values let's say company group encoded in the persistence like `__vendor__broker__customer__`. Then if we use crits ['group' => 'customer'] this is `translated` to actual condition to ['group', 'like', '%\\_\\_customer\\_\\_%']. This translation is performed by the field class.
- Having Crits like this enables saving the complete Crits class in persistence and applying them programmatically (think about saving user access and applying it as Model scope this way in the atk4/login)
- Crits can be 'translated' to human readable text (using previous example would be "Group contains Customer")
- Crit value can still be expression
- Crits can be easily represented in UI using the https://querybuilder.js.org/
- Crit value supports also 'placeholders' - think of defining user access using access of a referenced model (Grant access of invoice items where user has access to invoice)

This functionality cannot be developed as addon to Model as Crits have generic methods in the Field subclasses.

One more thing to consider is to extend the Reference class from Field class. This would enable using the conditions on references in the crits.

That is why this is all related to the https://github.com/atk4/data/tree/feature/number-field branch and the refactoring of field classes.

I can expand on details and also share some code if idea is interesting for the atk4/data.

georgehristov:
Related to #524


tests/PersistentArrayTest.php
```diff
@@ -489,6 +489,33 @@ public function testRegexp()
...
+            2 => ['id'=>2, 'number'=>'ABC8', 'items'=>12, 'active'=>0],
+            3 => ['id'=>3, 'items'=>13, 'active'=>1],
+            4 => ['id'=>4, 'number'=>'ABC6', 'items'=>14, 'active'=>0],
+            5 => ['id'=>5, 'number'=>'ABC5', 'items'=>15, 'active'=>0],
+            6 => ['id'=>6, 'number'=>'ABC4', 'items'=>16, 'active'=>1],
+            7 => ['id'=>7, 'number'=>'ABC3', 'items'=>17, 'active'=>0],
+            8 => ['id'=>8, 'number'=>'ABC2', 'items'=>18, 'active'=>1],
+            9 => ['id'=>9, 'items'=>19, 'active'=>1],
+            10 => ['id'=>20, 'items'=>0, 'active'=>1],
+            10 => ['id'=>20, 'items'=>null, 'active'=>1],
```
abbadon1334: this index is repeated, i think is 11

abbadon1334 (changes requested):

tests/PersistentArrayTest.php
georgehristov: Fixed, thanks!

PhilippGrashoff (commented):
All changes in Field.php (and maybe in other files, too) revert changes lately made which include:
- return type hinting
- usage of \Some\Class::class instead of string '\Some\Class'
- ?? operator instead of isset() ? :
I think we should keep all 3 code updates. Personally, I find return type hinting where possible is really important, as the IDE will pick it up and make coding faster and less error prone.

georgehristov:
@PhilippGrashoff  no changes are made to the Field class file. When merged this branch will keep the changes to all files recently made.
I have prepared a merge of develop into this branch but did not commit it to keep changes made clearly visible.

PhilippGrashoff:
Ah, then maybe I do not use git/github a 100% right. I always go to "Files Changed" at the top. In there it shows quite some changes to Field.php. Is there a better way?

georgehristov:
@PhilippGrashoff Now I rebased it on latest develop branch

src/Model.php
```diff
@@ -1744,22 +1710,24 @@ public function loadBy(string $field_name, $value)
...
+            $this->scope = $scope;
             $field->system = $system;
             $field->default = $default;
 
             throw $e;
         }
 
         // restore
-        array_pop($this->conditions);
+        $this->scope = $scope;
```
romaninsh: 👍

src/Model/Scope/AbstractScope.php
```diff
@@ -0,0 +1,194 @@
...
+     */
+    abstract public function setModel(Model $model = null);
+
+    /**
+     * Negate the scope object
+     * e.g from 'is' to 'is not'.
+     *
+     * @return $this
+     */
+    abstract public function negate();
```
romaninsh, line 82: can this be documented in the main documentation?

src/Model/Scope/AbstractScope.php
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /**
+     * Use a model.
+     *
+     * @param Model $model
+     *
+     * @return static
+     */
+    final public function on(Model $model)
```
romaninsh, line 62: more methods need to be explained in the main doc.

src/Model/Scope/AbstractScope.php
```diff
@@ -0,0 +1,194 @@
...
+    {
+        return $this->active && !$this->isEmpty();
+    }
+
+    /**
+     * Returns if scope contains several conditions.
+     *
+     * @return bool
+     */
+    public function isCompound()
```
romaninsh: looks like this is not implemented.

romaninsh: oh nevermind, perhaps because it's an abstract........

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,391 @@
...
+
+        $data = [1 => $values];
+
+        $model->withPersistence(new Array_($data));
+
+        $model->add($this);
+
+        return $model->export() ? [] : [$this];
+
+//     	$match = false;
```
romaninsh: debug code?

docs/conditions.rst
```diff
@@ -132,6 +132,98 @@ to set those conditions inside the init() method of your model::
...
+Using the Model::addCondition method is the basic way to limit the model scope of records. Under the hood
+Agile Data utilizes a special set of classes (Condition and Scope) to apply the conditions as filters on records retrieved.
+These classes can be used directly and independently from Model class to define and store Model scope.
+
+.. php:class:: Condition
+
+Condition represents a simple scope in a form [field, operation, value], similar to the functionality of the 
+Model::addCondition method
+
+.. php:class:: Scope
```
romaninsh, line 201: need to document methods and provide examples for methods.

src/Model.php
```diff
@@ -171,6 +172,8 @@ class Model implements \ArrayAccess, \IteratorAggregate
      */
     public $conditions = [];
 
+    protected $scope;
```
romaninsh, line 199: is there support for HAVING .. ?

src/Model/Scope/Scope.php
```diff
@@ -0,0 +1,366 @@
...
+
+            if ($scope->isEmpty()) {
+                continue;
+            }
+
+            $this->addComponent(clone $scope);
+        }
+    }
+
+    public function find($key)
```
romaninsh: no docs at all..

romaninsh (changes requested):

src/Model/Scope/AbstractScope.php
georgehristov: Done

georgehristov:
@romaninsh docs updated

src/Model/Scope/Condition.php
georgehristov: Removed

DarkSide666:
Sorry guys, I definitely would love to carefully review this PR.
Just trying to find right moment when to fully focus on this.
Same with Fields refactor PR.

DarkSide666:
I will try to review this PR some day this week. I really would love to see it released in atk4/data `v2.2`, so merged in develop after `v2.1` is released.

@georgehristov in a meantime can you please try to fix conflicts with current develop branch?

georgehristov:
@DarkSide666 I have now rebased this branch on latest `develop` so it is ready for review

georgehristov:
@DarkSide666 so this branch is now rebased on develop after the 2.1 release

romaninsh:
Awesome - I'm approving this for the merge! @DarkSide666 please approve and merge, then we can work through the consequences :)

A few things that I believe would be needed:

 - Support for arguments in scopes (apply scope and supply associative list of typed arguments)
 - Integration between scopes and CRUD (e.g. select filter in drop-down, optionally allow to enter arguments)
 - Support for conditions in related entities (e.g. "invoice.client.address.country.is_eu = true")

I think those would be natural additions once the basic scope support is there.

romaninsh (approved):

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+$operator can be one of the supported operators >, <, >=, <=, !=, in, not in, like, not like, regexp, not regexp
+$value can be Field object, Expression object, array (interpreted as 'any of the values') or other scalar value
+
+If $value is omitted as argument then $operator is considered as $value and '=' is used as operator
+
+.. php:method:: negate();
+
+Negates the condition, e.g::
+
+	// results in 'name is not John'
```
DarkSide666, line 164: `name != John` ?
what if we do `Condition::create('price', '>', 100)->negate()` ? Will it convert to `<=` ?

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+.. php:method:: validate($values);
+
+Validates $values array if matching the condition when applied on $model. Returns array of conditions that are not met or empty if $values fit ::
+
+	$condition = Condition::create('name', 'John');
+	
+	// as condition is that only contacts named John are valid
+	// $result will return array($condition) as $condition is not met
+	// this array can be used to display validation errors
+	$result = $condition->on($contactModel)->validate(['name' => 'Peter']);
```
DarkSide666, line 195: Not sure I understood purpose of this method.
Also what if instead of field name `name` we will have expression or field object?

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+	$condition = Condition::create('name', 'John');
+	
+	// as condition is that only contacts named John are valid
+	// $result will return array($condition) as $condition is not met
+	// this array can be used to display validation errors
+	$result = $condition->on($contactModel)->validate(['name' => 'Peter']);	
+
+.. php:method:: find($key);
+
+Returns array of conditions whose key property matches the $key parameter. Useful when renaming fields and updating saved scope objects
```
DarkSide666, line 199: By reading docs I had no idea what it does :)
Also how it can return array of conditions if Condition object = 1 condition.

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+
+	$contact->add($scope); // adding scope to model
+	$contact->scope()->and($conditionXYZ); // adding more conditions
+	
+	
+.. php:method::	create($scopeOrArray = null, $junction = Scope::AND);
+
+Creates a scope object from array. If scope is passed as first argument uses this as result::
+
+	// below will create 2 conditions and join them with OR
```
DarkSide666: In method description default junction is AND, but here you say it's OR by default.

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+	$scope1->negate(); 
+
+.. php:method:: and(AbstractScope $scope);
+
+Merge $scope into current scope using AND as junction. If current scope junction is AND simply adds a component.
+If it is OR then changes it to AND and uses current scope and $scope as two sub-components
+
+.. php:method:: or(AbstractScope $scope);
+
+Merge $scope into current scope using AND as junction. If current scope junction is OR simply adds a component.
```
DarkSide666: Error : _...using AND as junction._

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+If it is AND then changes it to OR and uses current scope and $scope as two sub-components
+
+.. php:method:: mergeAnd(AbstractScope $scopeA, AbstractScope $scopeB, $_ = null);
```
DarkSide666, line 267: These methods can also be used to merge Conditions not only Scopes. right?

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+
+Merge number of scopes using OR as junction. Returns the resulting scope.
+
+.. php:method:: merge(AbstractScope $scopeA, AbstractScope $scopeB, $junction = self::AND);
+
+Merge two scopes using $junction. Returns the resulting scope.
+
+.. php:method:: find($key);
+
+Returns array of conditions whose key property matches the $key parameter. Useful when renaming fields and updating saved scope objects
```
DarkSide666: Still not sure what it does. What is _key property_ etc.

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+
+.. php:method:: addComponent(AbstractScope $scope);
```
DarkSide666, line 288: No description what is term _component_ here. As result it's impossible to understand what following methods should do. Also no code examples for them, so have absolutely no idea what scope component is.

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+
+Conditions on Referenced Models
+-------------------------------
+
+Agile Data allows for adding conditions on related models for retrieval of type 'model has references where'.
+
+Setting conditions on references can be done utilizing the Model::refLink method but there is a shorthand format 
+directly integrated with addCondition method using "/" to chain the reference names::
+
+	$contact->addCondition('company/country', 'US');
```
DarkSide666, line 324: Is negation possible here too ? Probably yes.

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+This will limit the $contact model to those whose company have more than 3 tickets.
+'company' and 'tickets' are the name of the chained references ('company' is a reference in the $contact model and
+'tickets' is a reference in Company model)
+
+For applying conditions on existence of records the '?' (has any) and '!' (doesn't have any) special symbols can be used.
+Although it is similar in functionality to checking ('company/tickets/#', '>', 0) or ('company/tickets/#', '=', 0)
+'?' and '!' special symbols use optimized query and are much faster::
+
+	// Contact whose company has any tickets
+	$contact->addCondition('company/tickets/?');
```
DarkSide666, line 343: Is it possible to set condition - _Contact whose company has any tickets which are expired_ ?
Something like `$contact->addCondition('company/tickets/valid_till?', '<', new \DateTime());`

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+
+	// Contact whose company doesn't have any tickets
+	$contact->addCondition('company/tickets/!');
```
DarkSide666, line 346: Will `negate()` also work to switch between `?` and `!` ?

docs/conditions.rst
```diff
@@ -132,6 +132,223 @@ to set those conditions inside the init() method of your model::
...
+
+Condition Value Placeholder
```
DarkSide666: I don't like this placeholder thing.
Instead we should be able to define arguments for Scope and Condition like in UserAction.
That way in UI (think about some report engine) we will be able to ask user for argument values before applying scope on model - similar as we do in UserActions. Without UI we also can pass array of argument values when applying Scope on Model.

src/Model.php
```diff
@@ -203,6 +204,8 @@ class Model implements \IteratorAggregate
...
```
DarkSide666, line 199: No `@var` typehint comment here

DarkSide666: Can't we set multiple scopes on model?
To do that we have to merge them first, right?

src/Model.php
```diff
@@ -1730,28 +1695,48 @@ public function tryLoadBy(string $field_name, $value)
...
     }
 
+    /**
+     * Check if model has persistence with specified method.
+     *
+     * @param string $method
+     *
+     * @throws Exception
+     */
+    public function checkPersistence(string $method = null)
```
DarkSide666, line 1517: Love this method. I wanted to create it to many times before and never did :)

src/Action/Iterator.php
```diff
@@ -31,69 +35,179 @@ public function __construct(array $data)
...
      */
-    public function where($field, $value)
+    public function filter(AbstractScope $scope)
     {
-        $this->generator = new \CallbackFilterIterator($this->generator, function ($row) use ($field, $value) {
-            // skip row. does not have field at all
-            if (!array_key_exists($field, $row)) {
-                return false;
-            }
+        if (!$scope->isEmpty()) {
```
DarkSide666, line 42: Shouldn't it be `isActive` here instead of `isEmpty` ?

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,419 @@
...
+            $value = $operator;
+            $operator = '=';
+        }
+
+        if (is_bool($key)) {
+            if ($key) {
+                return;
+            }
+
+            $key = new Expression('false');
```
DarkSide666, line 74: Can you give example use-case of true/false here?

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    public function setModel(Model $model = null)
+    {
+        if ($this->model = $model) {
+            // if we have a definitive scalar value for a field
+            // sets it as default value for field and locks it
+            // new records will automatically get this value assigned for the field
+            // @todo: consider this when condition is part of OR scope
+            if ($this->operator === '=' && !is_object($this->value) && !is_array($this->value)) {
```
DarkSide666, line 89: I think we can allow value to be object or array too, but need some testing.

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,419 @@
...
+            return [];
+        }
+
+        if (!$model = $this->model) {
+            throw new Exception('Model must be set using setModel to validate');
+        }
+
+        $data = [1 => $values];
+
+        $model->withPersistence(new Array_($data));
```
DarkSide666, line 198: This is cloning original model, right?

src/Persistence/Array_.php
```diff
@@ -199,73 +194,88 @@ public function update(Model $m, $id, $data, $table = null)
...
+
+    /**
+     * Last ID inserted.
+     * Last inserted ID for any table is stored under '$' key.
+     *
+     * @param Model $model
+     *
+     * @return mixed
+     */
+    public function lastInsertId(Model $model = null)
```
DarkSide666, line 238: I like this solution.

src/Persistence/SQL.php
```diff
@@ -372,60 +375,38 @@ protected function setLimitOrder(Model $m, Query $q)
...
      */
-    public function initQueryConditions(Model $m, Query $q): Query
+    public function initQueryConditions(Model $model, Query $query, AbstractScope $scope = null): Query
     {
-        if (!isset($m->conditions)) {
-            // no conditions are set in the model
-            return $q;
+        $scope = $scope ?? $model->scope();
+
+        if (!$scope || $scope->isEmpty()) {
```
DarkSide666: `isEmpty()` or `isActive()` here?

tests/PersistentSQLTest.php
```diff
@@ -137,8 +137,12 @@ public function testModelInsertRows()
         $m->addField('name');
         $m->addField('surname');
 
+        $this->assertEquals(0, $m->action('exists')->getOne());
+
         $m->import($a['user']); // import data
 
+        $this->assertEquals(1, $m->action('exists')->getOne());
```
DarkSide666, line 146: Maybe it should return `true` instead?

tests/ScopeTest.php
```diff
@@ -0,0 +1,450 @@
...
+ *
+ * We will have few of them. You can lookup country by name or by code. We will also try looking up country by
+ * multiple fields (e.g. code and 'is_eu') to see if those work are used as AND conditions. For instance
+ * is_eu=true, code='US', should not be able to lookup the country.
+ *
+ * Users is a reference. You can specify it as an array containing import data for and that will be inserted
+ * recursively.
+ *
+ * We also introduced 'user_names' field, which will concatenate all user names for said country. It can also be
+ * used when importing, simply provide a comma-separated string of user names and they will be CREATED for you.
```
DarkSide666: It's possible that CSV is not supported anymore.

tests/ScopeTest.php
```diff
@@ -0,0 +1,450 @@
...
+ *
+ * When importing users, you should be able to specify country using 'country_id' or using some of the
+ * lookup fields: 'country', 'country_code' or 'is_eu'.
+ *
+ * If ID is not specified (unlike specifying null!) we will rely on the lookup fields to try and find
+ * a country. If multiple lookup fields are set, we should find a country that matches them all. If country
+ * cannot be found then null should be set for country_id.
+ *
+ * Friends is many-to-many relationship. We have 'friend_names' field which may be similar to the one we had
+ * for a country. However specifying friend_names as a comma-separated value will not create any friends.
```
DarkSide666: Check CSV support.

DarkSide666 (changes requested):
This all looks very nice and I really like it.

Tough I made few comments while reviewing and at least some of them need some attention.
1. what is component - docs not sufficient about them?
2. what is find() method - to less info in docs and will it work always even if field is set as field object not field name?
3. does scopes support HAVING syntax?
4. isEmpty() and isActive() - not exactly sure when each of them should be used.
5. CSV is probably no more supported, but need to check.
6. Placeholders vs. Arguments.
7. maybe something more from comments ...

docs/conditions.rst
georgehristov: Yes, it converts "negate >" to "<=". `Condition::$opposites` is used as mapping.

docs/conditions.rst
georgehristov: Imagine you have an array that you want to add as Model record. This method allows you to check if the resulting record will be within the Model scope.
It returns an array of conditions which are not met so the array falls within the scope.
It is very useful for advanced access control and other uses.

It works well with `Field` object as well as it simply uses the Model with `Array_` persistence, adds the array as record and checks if model outputs any records when scope applied.
Where is does not work is if `HasOne_SQL` reference is used in the model.
To overcome this I created Persistence/ArraySQLite which uses same principle but with SQLite in memory.

docs/conditions.rst
georgehristov: Fixed

docs/conditions.rst
georgehristov: Fixed

tests/PersistentSQLTest.php
georgehristov: It is the result of the query `select exists (select * from "user")` and the result is either 0 or 1.
IMHO it is good enough for the purpose. Switching to bool values will make code more complicated

src/Persistence/SQL.php
georgehristov: Switched to `isActive`

tests/ScopeTest.php
georgehristov: Not sure why you refer to CSV not working anymore. The `SUser` and `SCountry` classes were copied along with comments from another test file.
I have now removed the comments as they are irrelevant.

docs/conditions.rst
georgehristov: Updated docs

georgehristov:
@DarkSide666 I have done some updates and comments, please review one more time

romaninsh:
@georgehristov - did you address the 7 questions @DarkSide666 posted above?

georgehristov:
@romaninsh yes, but below is as a numbered list:

1. what is component - **_docs updated_**
2. what is find() method - **_docs updated_**
3. do scopes support HAVING syntax? - **_at the moment not_**
4. isEmpty() and isActive() - **_`isEmpty` checks if scope applies any conditions on the model. `isActive` checks if the scope should be applied to the model.
Deactivating a sub-condition comes handy with placeholder values. For instance in some cases, if placeholder actual value is `NULL` the condition can be deactivated._**
5. CSV is probably no more supported, but need to check - **_not sure why it would not work?_**
6. Placeholders vs. Arguments - **_if the question is about renaming I believe "placeholder" is more descriptive and not as general as "argument"_**
7. maybe something more from comments - **_comments added to review line comments_**

docs/conditions.rst
georgehristov: Yes, merging conditions and scopes

docs/conditions.rst
georgehristov: Description of component implemented in the docs

docs/conditions.rst
georgehristov: Yes, possible

docs/conditions.rst
georgehristov: yes, it is possible (just without the question mark)
`$contact->addCondition('company/tickets/valid_till', '<', new \DateTime());`

docs/conditions.rst
georgehristov: The placeholder is important as you save the scope in database.
This means the same scope is saved (using placeholders) but it will be differently applied depending on system user, etc.
E.g if you have a placeholder `__PERSPECTIVE__` which resolves to a list of companies the user is responsible for.
Then you can save a scope for all users but it will resolve to different actual scope when applied to the model based on user presently active.
All these functionalities are refined implementation based on the real use, business proven designs.

src/Model.php
georgehristov: You can add a scope to a model (default merge with AND) or you can modify the existing scope

src/Model.php
DarkSide666: no

docs/conditions.rst
DarkSide666: yeah right

docs/conditions.rst
DarkSide666: No answer on this. Will
`$contact->addCondition('company/tickets/!')->negate() === $contact->addCondition('company/tickets/?');`

docs/conditions.rst
DarkSide666: OK probably placeholders are good too, but still need to think about arguments.

For example, think about scope to select invoices issued last year.
In data alone it can be implemented with placeholders something like this, right?
```
Condition::registerPlaceholder('__FROM__', [
    'label' => 'Date From',
    'value' => function(Condition $condition) {
    	return new \DateTime('first day of last year);
    },
]);
Condition::registerPlaceholder('__TILL__', [
    'label' => 'Date Till',
    'value' => function(Condition $condition) {
    	return new \DateTime('last day of last year);
    },
]);

$condition = Condition::create([
    ['date', '>=', '__FROM__'],
    ['date', '<=', '__TILL__'],
]);
```
If I get these dates from somewhere else then I still can use registerPlaceholder() method to set them.

But what if I need to ask user (in UI) to fill in these placeholder values? We can ask values for all placeholders maybe, but still then we have no knowledge of value types here and UI will never know that it has to ask for date value instead of string.

It's not that I don't like placeholders, but there is something missing about them. Anyway I guess we can extend that later on.

src/Model/Scope/Condition.php
DarkSide666: Will see that in real use cases.

DarkSide666 (approved):

src/Model.php
```diff
@@ -1592,22 +1561,24 @@ public function loadBy(string $field_name, $value)
...
+            $this->scope = $scope;
             $field->system = $system;
             $field->default = $default;
 
             throw $e;
         }
 
         // restore
-        array_pop($this->conditions);
+        $this->scope = $scope;
```
mvorisek: This is not a restore, not sure what @romaninsh likes on this. Absolute no go from me!

mvorisek (changes requested):
more feedback on Discord

abbadon1334 (approved):

src/Model.php
georgehristov: I will change it but where do you see the problem in using cloned scope as replacement?

mvorisek: see Discord example. Also, with deep Scopes I expect cloning to be expensive op

romaninsh: @georgehristov - were you able to look into this?

georgehristov: Yes, forgot about it but I will fix that

georgehristov: @mvorisek better now?

mvorisek: bettern but any restore needs to be done in `try/finally` fashion

mvorisek:
was all my feedback from Discord integrated?

mvorisek:
seems like Discord feedback was not resolved, as phpdocs etc. are missing

georgehristov:
@mvorisek please add one inline comment example of phpdoc for var and list any outstanding comments here

src/Model/Scope/Scope.php
```diff
@@ -0,0 +1,355 @@
...
+
+    public function isCompound(): bool
+    {
+        return count($this->getActiveComponents()) > 1;
+    }
+
+    /**
+     * @return self::AND|self::OR
+     */
+    public function getJunction()
```
mvorisek: no return type

src/Model.php
mvorisek: this is quite important point at least to be easily supported in the future and we should probably use HAVING automatically where needed

src/Action/Iterator.php
```diff
@@ -33,71 +37,175 @@ public function __construct(array $data)
...
-        });
+            // any matches && all matches the same (if all required)
+            $match = array_filter($matches) && ($scope->isAnd() ? count(array_unique($matches)) === 1 : true);
+        }
 
-        return $this;
+        return $match;
+    }
+
+    protected function where($v1, $operator, $v2)
```
mvorisek: protected, but should be probably named better, this function does not add conditions, it calculates the outcome

mvorisek:
> 
> 
> @mvorisek please add one inline comment example of phpdoc for var and list any outstanding comments here

see https://discord.com/channels/657649422280425523/664534590160896006/722466104529649674

I need to be sure to not recheck twice not finished feedback previously

src/Model.php
georgehristov: Let me have an example HAVING query use case and I will work to implement it in scopes

mvorisek: Table product
varchar name
varchar category
int price;

insert into product (name, price) values
('apple', 'fruit', 10)
('potato', 'vegetable', 2)
('mango', 'fruit', 30)

SELECT category, avg(price) as avg_price
FROM product
GROUP BY category
HAVING avg_price > 15

should not fail (with SQL `WHERE` is will) and return `fruit, 20`

but this needs to be probably solved absolutely outside scopes - we should identify when HAVING is needed and then use it on the fly

src/Action/Iterator.php
```diff
@@ -146,11 +146,11 @@ protected function match(array $row, AbstractScope $scope)
         return $match;
     }
 
-    protected function where($v1, $operator, $v2)
+    protected function evaluate($v1, $operator, $v2): bool
```
mvorisek: what evaluate? `evaluateWhere` is fine...

src/Model/Scope/Scope.php
georgehristov: done

src/Action/Iterator.php
georgehristov: done

src/Action/Iterator.php
georgehristov: evaluateIf ?

mvorisek: we do not evaluate Iterator/self, nothing more...

src/Action/Iterator.php
```diff
@@ -146,11 +146,11 @@ protected function match(array $row, AbstractScope $scope)
         return $match;
     }
 
-    protected function evaluate($v1, $operator, $v2): bool
+    protected function evaluateIf($v1, $operator, $v2): bool
```
mvorisek, line 149: If is language building block which consists of expression and it consists of operators. So probably `evaluateOperator` :) (which is better than my original proposal :D )

georgehristov: evaluateIf is perfect so we keep it.
method evaluates if $v1 = $v2 for instance. Perfectly clear.
No sense in discussing further on such minor points.

mvorisek: not true

mvorisek: Imagine no knowledge. What says `evaluateIf` to you? and what `evaluateOperator`? With first example, I do not know.

georgehristov: It is a protected method and it is quite clear from the signature.

Anything else about the scopes?

mvorisek: see https://discord.com/channels/657649422280425523/664534590160896006/722466104529649674

src/Model/Scope/Scope.php
```diff
@@ -208,18 +208,14 @@ public function toWords(bool $asHtml = false): string
...
     {
-        if ($scopeOrArray instanceof AbstractScope) {
-            return $scopeOrArray;
-        }
-
-        return new static ($scopeOrArray, $junction);
+        return new static ($scopes, $junction);
     }
 
     public function __construct($scopes = null, $junction = self::AND)
```
mvorisek, line 216: we should accept an array only

src/Model/Scope/Scope.php
```diff
@@ -208,18 +208,14 @@ public function toWords(bool $asHtml = false): string
...
      * Create a scope from array of scopes or arrays.
      *
-     * @param mixed  $scopeOrArray
+     * @param string|array  $scopes
      * @param string $junction
      *
      * @return static
      */
-    public static function create($scopeOrArray = null, $junction = self::AND)
+    public static function create($scopes = null, $junction = self::AND)
```
mvorisek: bad design - should user use `new Cl` or this?

src/Action/Iterator.php
romaninsh: I don't see big problem with "evaluateIf", happy to go with @georgehristov as he is the author of the code. not a dealbreaker.

src/Model/Scope/Scope.php
georgehristov: Both are possible and equal.

romaninsh: I would only endorse static create() if it would be returning instances of various classes. Since it always returns new Scope, there is no benefit in having a static call, so I propose not to have this method. 

it would confuse by offering two identical implementations for no benefit.

src/Model/Scope/Scope.php
mvorisek: or nothing at all

src/Action/Iterator.php
romaninsh: My own opinion here is that @georgehristov should have the choice of function name. Options proposed were: "if", "evaluate", "evaluateIf", "checkIf".  I could ask "verifyIf" and "compare" too. Few arguments - "evaluate" imply a deep evaluation of expression, e.g. ability to have nested expressions of sorts. verify() sounds more like a contingency check (like some sort of validator). compare() would imply to have only 2 arguments and return 1,0,-1. 

I do NOT think that "evaluateIf" is miss-leading, but I'm happy if @georgehristov could look at all SUGGESTIONS and decide what he wants to call the function. I will support his decision, (just don't call it "download()") ! :D

mvorisek: one/my option was `evaluateOperator`

src/Model/Scope/Scope.php
georgehristov: So `create` methods removed

src/Model/Scope/Condition.php
```diff
@@ -83,26 +83,15 @@ class Condition extends AbstractScope
...
-     * @return static
-     */
-    public static function create($key, $operator = null, $value = null)
+    public function __construct($key, $operator = null, $value = null)
     {
         $args = is_array($key) ? $key : func_get_args();
 
-        return new static (...$args);
-    }
+        $key = $args[0];
```
mvorisek: there 3 lines does nothing

src/Model/Scope/Condition.php
```diff
@@ -83,26 +83,15 @@ class Condition extends AbstractScope
...
+        $operator = $args[1] ?? null;
+        $value = $args[2] ?? null;
 
-    public function __construct($key, $operator = null, $value = null)
-    {
-        if (func_num_args() == 2) {
+        if (count($args) == 2) {
```
mvorisek: I think this is not good, we should:
- require operator
- or move it to the 3rd param

src/Model/Scope/Scope.php
```diff
@@ -322,7 +317,7 @@ public function or(AbstractScope $scope)
      */
     public static function mergeAnd(AbstractScope $scopeA, AbstractScope $scopeB, $_ = null)
```
mvorisek, line 318: what is the use case?

I think is it when wrong - and self::AND will be used as a value instead of operator

exactly a reason why to not allow these flexi constructors

src/Model.php
DarkSide666: HAVING is used only when grouping and that can be done by GroupModel. See https://github.com/atk4/report/blob/develop/src/GroupModel.php#L319.
So maybe Having should be used automatically in GroupModel but normally in standard model it probably have no use.

mvorisek: I do not think so - any expr. field with non trivial expression (pure column name only) needs having:

Example:
```
SELECT CONCAT(name, 'dummy') as x FROM `test` where x
```
throws an error: `#1054 - Unknown column 'x' in 'where clause'`

src/Model/Scope/Condition.php
georgehristov: They did but I removed this support now

src/Model/Scope/Condition.php
georgehristov: The second argument is required now

georgehristov:
@DarkSide666 I have now introduced the scope instance specific placeholders which I believe are exactly the behavior required as discussed. The other placeholders I have renamed to global placeholders.
I was looking into my implementation of creating "reports" which will be quite native to ATK and placeholder functionality for scopes on instance level is one of the features that will facilitate it.

Under my implementation, a Report class defines filter fields (e.g enum for selecting an year for the data in the report).
The resulting values from the filters I call crits (can be named anything understandable but I use this term to explain the point)
Based on the crits the Report class can also generate meta-crits (e.g. based on the selected year the meta-crits `beginDate` and `endDate` are generated)
Each Report also defines display blocks (Views in ATK) like tables, charts, lists, etc each with its own Model. These blocks control the display on a web page but also the export to external formats (like Word, Excel, LibreOffice) for which I use OpenTBS or PDF (I use JasperReports).
Here comes the good part with scopes and placeholders.
So for each report block the model scope can be statically pre-defined when designing the report but using placeholders:
E.g if the report contains a table with list of invoices then scope for this block would be:
```
$invoiceScope = new Scope ([
   ['invoice_date', '>=', '::beginDate::'] // using the meta-crit as placeholder
   ['invoice_date', '<=', '::endDate::'] // using the meta-crit as placeholder
]);
```
And later when the report instance is being generated then Report object registers the crits as placeholders to the scopes of all of its blocks `$invoiceScope->registerPlaceholders($crits)`.

mvorisek:
@georgehristov `::beginDate::` is some special placeholder syntax?

mvorisek:
@georgehristov from chat with DarkSide:

image

image

image

Is it ok with you to reduce Scopes to minimal code required for the 3 mentioned points?

Placeholders etc. seems like an idea or one usecase only to me now. Is there any downside supporting them thru an extended class for your usecase and do they need to be integrated with other repos (ui, ...)?

I just want to find a way, how we can merge the base Scope code code shortly without having to wait forever for the additional questions marks...

src/Action/Iterator.php
```diff
@@ -33,71 +37,175 @@ public function __construct(array $data)
...
+            $match = array_filter($matches) && ($scope->isAnd() ? count(array_unique($matches)) === 1 : true);
+        }
 
-        return $this;
+        return $match;
+    }
+
+    protected function evaluateIf($v1, $operator, $v2): bool
+    {
+        switch (strtoupper((string) $operator)) {
```
mvorisek, line 151: when operator is not string, it is a mistake - string types exception is only the correct outcome!

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+
+        return [$field, $operator, $value];
+    }
+
+    public function isEmpty(): bool
+    {
+        return array_filter([$this->key, $this->operator, $this->value]) ? false : true;
+    }
+
+    public function validate(array $values): array
```
mvorisek, line 186: very inconsistent with anythink else - Model->set is validate immediatelly

this should always validate, even if disabled

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    public function toWords(bool $asHtml = false): string
+    {
+        if (!$this->model) {
+            throw new Exception('Model must be set using setModel to convert to words');
+        }
+
+        // make sure clones are used to avoid changes
+        $condition = clone $this;
```
mvorisek, line 236: any clone for methods like to string, to array, is absolutely bad design, state should never be mutated by theses methods

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+        $operator = $condition->operatorToWords($asHtml);
+
+        $value = $condition->valueToWords($condition->value, $asHtml);
+
+        $ret = trim("{$key} {$operator} {$value}");
+
+        return $asHtml ? $ret : html_entity_decode($ret);
+    }
+
+    protected function keyToWords(bool $asHtml = false): string
```
mvorisek, line 249: why key, why not field? when key is not a field?

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+            $field = $model->hasField($field) ? $model->getField($field) : null;
+        }
+
+        if ($field instanceof Field) {
+            $words[] = $field->getCaption();
+        } elseif ($field instanceof Expression) {
+            $words[] = "expression '{$field->getDebugQuery()}'";
+        }
+
+        $string = implode(' ', array_filter($words));
```
mvorisek, line 292: why need to filter?

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+
+        return $asHtml ? "<strong>{$string}</strong>" : $string;
+    }
+
+    protected function operatorToWords(bool $asHtml = false): string
+    {
+        return $this->operator ? (self::$dictionary[strtoupper((string) $this->operator)] ?? 'is equal to') : '';
```
mvorisek, line 299: hmm, operator should never be option

do we need to support unary expression?

if so, we should never add anything, maybe not even allow them to use in unary condition class design (more braninpower needed)

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+        if (is_object($value)) {
+            if ($value instanceof Field) {
+                return $value->owner->getModelCaption() . ' ' . $value->getCaption();
+            }
+
+            if ($value instanceof Expression || $value instanceof Expressionable) {
+                return "expression '{$value->getDebugQuery()}'";
+            }
+
+            return 'object ' . print_r($value, true);
```
mvorisek, line 328: seems more like for a debug code...

src/Model/Scope/AbstractScope.php
```diff
@@ -0,0 +1,225 @@
...
+     */
+    final public static function registerGlobalPlaceholder(string $key, $options)
+    {
+        self::$globalPlaceholders[$key] = is_array($options) ? $options : [
+            'label' => is_string($options) ? $options : $key,
+            'value' => $options,
+        ];
+    }
+
+    public static function __set_state($array)
```
mvorisek, line 214: usecase?

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+        $data = [1 => $values];
+
+        $model->withPersistence(new Array_($data));
+
+        $model->add($this);
+
+        return count($model->export()) !== 0 ? [] : [$this];
+    }
+
+    public function negate()
```
mvorisek, line 205: This is tough - it does not negate the condition because of NULL will never match with `<` or `>`.

src/Model/Scope/Condition.php
```diff
@@ -0,0 +1,384 @@
...
+        if ($this->operator && isset(self::$opposites[$this->operator])) {
+            $this->operator = self::$opposites[$this->operator];
+        } else {
+            throw new Exception('Negation of condition is not supported for ' . ($this->operator ?: 'no') . ' operator');
+        }
+
+        return $this;
+    }
+
+    public function find($key): array
```
mvorisek, line 216: not sure what it should do nor return

mvorisek:
There are so many things to fix. I would suddgest isolating it even more for the first integration - address the 1st point only, ie. "clear and/or nested conditions structure", ie. introduce very simple classes for Scope and Condition which can replace Model::where() impl. No one line of extra code. All other is additional functionality that can be added later very easily.

@georgehristov @DarkSide666 wdyt, do we need more that this now?

src/Model/Scope/Condition.php
georgehristov: Read the docs.

As everybody is saying it is not understandable I will remove it.

src/Model/Scope/Condition.php
georgehristov: Condition can be raw Expression. That is why iperator is not compulsory.

src/Model/Scope/Condition.php
georgehristov: When key is an Expression for instance

src/Model/Scope/Condition.php
georgehristov: It cannot as scope is stand-alone and may not be associated with model to validate to

georgehristov:
@mvorisek not sure where you see duplicate code.
Placeholders is a powerful construct. Originally I made them using a separate class. Here only an array is used as the construct is simple enough.
If we call them arguments or placeholders is nevermind but the functionality is there.

src/Model/Scope/Condition.php
mvorisek: I do not read docs as a rule ;-)

src/Model/Scope/Condition.php
mvorisek: I can imagine this situation, but once it is associated...  Also, it is active by default, so it will fail when not associated by default...

src/Model/Scope/AbstractScope.php
georgehristov: Method is called when using `var_export($scope)`.
I use this when creating a script or commant to be executed in the background on a separate process not dependent on current PHP process.

src/Model/Scope/Condition.php
georgehristov: Too bad for fresh piece of code where you do not have enough examples as use cases. Then you need to wait for some implementations to be done to understand what it is about.

mvorisek: I expect intuitive API :)

georgehristov:
@DarkSide666 I believe I am getting the idea you refer to as using arguments. Basically you want arguments to be defined using seed of Field as with UserAction so they can be easily used for generating ui components.
I will work on implementing it this way instead of defining it using an array.

mvorisek:
I think anything other that special placeholder object is insane if mixed with other values.

Example:
```
class Scope\Placeholder {
// do anything you want here to support placeholder
}
```

georgehristov:
Yes, this was my approach initially. Can implement it, no problem.

What I will do now is:
- definition of placeholders will be key => field seed pairs (like `args` in UserAction)
- we can use Placeholder objects do define replaceable values in the conditions `['name', new Scope\Placeholder('some_placeholder_key')]`
- we will set the arguments in the scope object using key->value pairs. `$scope->setArguments(['some_placeholder_key' => 'actual_value'])`. These values could come from ui form generated using placeholder fields or from code. Then Placeholder objects will be replaced with the respective value from the arguments (or label from the field definition in case of `toWords`) when scope used.

mvorisek:
I am not sure if you and/or I get @DarkSide666 correctly, but to me, messing in Scope with field seeds is absolutely wrong idea.

georgehristov:
This is exactly what he is talking about and I think it is good as then you can directly use the Fields to generate form in ui or validate the values, etc.
It will be absolutely same as in UserAction.

src/Model.php
DarkSide666: Yes and no.
In this case you *can* use `having x` instead of `where x`, but need to keep in mind that `having` works only after data is selected. So if you use `having` then DB engine first will select all data and only then start filtering it. That's slower and requires more resources from DB server than using `where`. `Where` instead filters data while it's selected, so resulting dataset is smaller and performance better.
So instead of using `having` you should always use `where` if possible and it's always possible if there are no grouping.
In your testcase above generated SQL should be `SELECT CONCAT(name, 'dummy') as x FROM test where CONCAT(name, 'dummy')` and atk4/data helps with that automatically if concat is defined as expression field in model.

mvorisek: yes and no

I think queries like select (select id from xx limit 1) where (select id from xx limit 1) are not guaranteed to be stable (because id can be different each time if there is no forced order)

One easy example to test this is with `RAND()` expression :D :D

I think there are also other major issues atk4 only assume instead of can rely on. But to keep this thread simple, I proposed to isolate Scope idea (ie. Scope class to define nested where conditions) from anything else and solve any freecoolin stuff later /in another PR.

Proof:
image

src/Model/Scope/Condition.php
DarkSide666: could be simply `toWords()` :)

georgehristov: Cannot as we have `Condition::toWords` and then protected `keyToWords`, `operatorToWords` and `valueToWords`

PR #552, opened 2020-04-11
Remove deprecated classes in 2.x version

mvorisek:


DarkSide666:
These could be removed only in next **major** version.

mvorisek:
Yes, comments to remove it were added in 1.x version :)

romaninsh:
@DarkSide666 - yeah, that's probably right. We usually provide realistic path to upgrade, e.g. if you fix your code from "_" to "\" then you can switch between 2.X and 1.X and both versions will work consistently. So I'm OK to merge this.

romaninsh (approved):

PR #554, opened 2020-04-11
Move types to code if possible

mvorisek:


PhilippGrashoff:
I really like this PR! Type hinting in combination with a good IDE is the way to go IMHO.

I saw there are some changes where code like this was removed:
```
        if (!is_array($defaults)) {
            $defaults = ['master_field' => $defaults];
        }
```

so before it was possible to pass any type to $defaults param, now it has to be an array. That will break some user code for sure.
Personally, I like this, I'd rather have a parameter set to a fixed type than that amount of flexibility. Lets see what the others say to this.
However, changes like this should be made to the next major release like 2.2, where its ok to have a bit of BC break I think.
Before releasing this, we definitely need to check the docs for these changed methods, I am pretty sure there will be example code which wouldn't work with these changes.

romaninsh:
Will probably be broken by https://github.com/atk4/data/pull/551 and vice versa.

georgehristov (approved):
LGTM

PR #556, opened 2020-04-13
Don't use addMethod() with non-callable. Also add return types

mvorisek:
Changes needed once PR https://github.com/atk4/core/pull/156 is merged.


romaninsh (approved):
ok

PR #559, opened 2020-04-14
Fix driver type refactor

mvorisek:
IMPORTANT: Update/commit translations before merging

Once updated, feel free to change the status from draft to mergeable.


DarkSide666 (approved):
Russian translation is still OK.

PR #564, opened 2020-04-22
Delete ConnectionInterface.php

mvorisek:
Used nowhere, `add()` method signature is not compatible with the one used in Atk.


georgehristov (approved):
LGTM

PR #565, opened 2020-04-22
Allow to add() only an object

mvorisek:
Plus fix #566

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -472,16 +472,15 @@ class Model implements \ArrayAccess, \IteratorAggregate
     }
 
     /**
-     * TEMPORARY to spot any use of $model->add(new Field(), 'bleh'); form.
+     * TEMPORARY to spot any use of $model->add(new Field(), ['bleh']); form.
      */
-    public function add($obj, $args = [])
+    public function add(object $obj, array $defaults = []): object
     {
-        $obj = $this->_add($obj, $args);
         if ($obj instanceof Field) {
             throw new Exception(['You should always use addField() for adding fields, not add()']);
         }
 
-        return $obj;
+        return $this->_add($obj, $defaults);
     }
 
     /**
@@ -1467,13 +1466,8 @@ class Model implements \ArrayAccess, \IteratorAggregate
      *
      * but will assume that both models are compatible,
      * therefore will not perform any loading.
-     *
-     * @param string|Model $class
-     * @param array        $options
-     *
-     * @return Model
      */
-    public function saveAs($class, $options = [])
+    public function saveAs(string $class, array $options = []): self
     {
         return $this->asModel($class, $options)->save();
     }
@@ -1504,13 +1498,8 @@ class Model implements \ArrayAccess, \IteratorAggregate
     /**
      * This will cast Model into another class without
      * loosing state of your active record.
-     *
-     * @param string|Model $class
-     * @param array        $options
-     *
-     * @return Model
      */
-    public function asModel($class, $options = [])
+    public function asModel(string $class, array $options = []): self
     {
         $m = $this->newInstance($class, $options);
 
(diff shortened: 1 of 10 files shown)


src/Model.php
```diff
@@ -1463,13 +1462,8 @@ public function duplicate($new_id = null)
...
      * but will assume that both models are compatible,
      * therefore will not perform any loading.
-     *
-     * @param string|Model $class
-     * @param array        $options
-     *
-     * @return Model
      */
-    public function saveAs($class, $options = [])
+    public function saveAs(string $class, array $options = []): self
```
DarkSide666, line 1470: BC break

src/Model.php
```diff
@@ -1528,28 +1517,17 @@ public function asModel($class, $options = [])
...
      * as $this.
      *
-     * @param string|Model $class
-     * @param array        $options
-     *
-     * @return Model
+     * @return static
      */
-    public function newInstance($class = null, $options = [])
+    public function newInstance(string $class = null, array $options = [])
```
DarkSide666, line 1525: this will break compatibility. not sure if that's ok.

tests/PersistentArrayTest.php
```diff
@@ -65,7 +65,7 @@ public function testSaveAs()
 
         $m = new Male($p);
         $m->load(1);
-        $m->saveAs(new Female());
+        $m->saveAs(Female::class);
```
DarkSide666, line 68: You're right that it's more logical to use class name here not object itself, but this change will break a lot of old code everywhere because historically we used `add('className')` A LOT :(

DarkSide666 (commented):
Historically we used add('className') A LOT so this will be a big backward compatibility break :(

I would like to get @romaninsh input here.

tests/PersistentArrayTest.php
mvorisek: So perfect then? :)

src/Model.php
mvorisek: That is correct, but to put this project in good shape we need it - as far as I remember these functions were used in a very few places.

tests/PersistentArrayTest.php
DarkSide666: As I said - it looks better, but will break existing projects (again) :(
Can you try to track down @romaninsh and ask what he thinks about this?

mvorisek: @DarkSide666 I reacted to this as you are writing:

> because historically we used add('className')

and this method accepts now exactly this "string className" :)

DarkSide666: Then I guess I added comment in wrong place. It should be for all this PR.
* add() will no longer support add(string $className)
* saveAs() will no longer support saveAs(object $model)
* newInstance() will no longer support newInstance(object $model)

I'm not saying that these changes are bad, but all and every of these are BC breaks...

romaninsh:
This would be one hell of BC break.

$foo->add(Some::class); is also extensively used.

We could depreciate this and target 4.0 :D

mvorisek:
I may be, but as you have already written in Discord, there is a time to move forward and enforce right semantics. Currently the user base is not that great and I strongly advice to prevent bad usage as soon as possible.

I propose to create central page with changes, so these changes can be fixed according to that. That is an industry common practice.

@romaninsh what will be different in Atk 4.0? My answed: more users, more unwanted usage. Right?

mvorisek:
@romaninsh legit?

DarkSide666 (approved):
Overall this PR is huge BC break, but on the other hand @mvorisek you're right about

> what will be different in Atk 4.0? My answed: more users, more unwanted usage. Right?

tests/PersistentArrayTest.php
mvorisek: Yes and it should be documented like so. This prevents mistakes, allow strong typing and also help the developer to understant what is really does.

Imagine `newInstance` like new, you never want to allow `new (new class())()`, it is simply wrong.

PR #567, opened 2020-04-22
Enforce array in method prototypes when appropriate

mvorisek:
Partial implementation of #554

- `typecastLoadRow()` method does no longer accepts null
- other method prototypes were cleaned up, mostly return types, needs more attention probably only if you override these methods

### DIFF
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -161,20 +161,16 @@ class Iterator
 
     /**
      * Return all data inside array.
-     *
-     * @return array
      */
-    public function get()
+    public function get(): array
     {
         return iterator_to_array($this->generator, true);
     }
 
     /**
      * Return one row of data.
-     *
-     * @return array
      */
-    public function getRow()
+    public function getRow(): array
     {
         $row = $this->generator->current();
         $this->generator->next();
--- a/src/Model.php
+++ b/src/Model.php
@@ -465,7 +465,7 @@ class Model implements \IteratorAggregate
      *
      * @return array [field => err_spec]
      */
-    public function validate($intent = null)
+    public function validate($intent = null): array
     {
         $errors = [];
         foreach ($this->hook(self::HOOK_VALIDATE, [$intent]) as $handler_error) {
@@ -652,7 +652,7 @@ class Model implements \IteratorAggregate
      *
      * @return Field[]
      */
-    public function getFields($filter = null)
+    public function getFields($filter = null): array
     {
         if ($filter === null) {
             return $this->fields;
@@ -872,10 +872,8 @@ class Model implements \IteratorAggregate
 
     /**
      * Returns array of model record titles [id => title].
-     *
-     * @return array
      */
-    public function getTitles()
+    public function getTitles(): array
     {
         $field = $this->title_field && $this->hasField($this->title_field) ? $this->title_field : $this->id_field;
 
@@ -1235,11 +1233,9 @@ class Model implements \IteratorAggregate
      * Use this instead of save() if you want to squeeze a
      * little more performance out.
      *
-     * @param array $data
-     *
      * @return $this
      */
-    public function saveAndUnload($data = [])
+    public function saveAndUnload(array $data = [])
     {
         $ras = $this->reload_after_save;
         $this->reload_after_save = false;
@@ -1538,8 +1534,6 @@ class Model implements \IteratorAggregate
     /**
      * Save record.
      *
-     * @param array $data
-     *
      * @return $this
      */
     public $_dirty_after_reload = [];
@@ -1668,8 +1662,7 @@ class Model implements \IteratorAggregate
      * This is a temporary method to avoid code duplication, but insert / import should
      * be implemented differently.
      *
-     * @param Model $m   Model where to insert
-     * @param array $row Data row to insert
+     * @param Model $m Model where to insert
      */
     protected function _rawInsert(self $m, array $row)
     {
@@ -1749,7 +1742,7 @@ class Model implements \IteratorAggregate
      * @param string     $key_field     Optional name of field which value we will use as array key
      * @param bool       $typecast_data Should we typecast exported data
      */
-    public function export($fields = null, $key_field = null, $typecast_data = true): array
+    public function export(array $fields = null, $key_field = null, $typecast_data = true): array
     {
         if (!$this->persistence->hasMethod('export')) {
             throw new Exception('Persistence does not support export()');
@@ -1765,7 +1758,7 @@ class Model implements \IteratorAggregate
         $key_field_added = false;
 
         // prepare array with field names
-        if (!is_array($fields)) {
+        if ($fields === null) {
             $fields = [];
 
             if ($this->only_fields) {
@@ -1827,7 +1820,7 @@ class Model implements \IteratorAggregate
      *
      * @return mixed
      */
-    public function getIterator()
+    public function getIterator(): iterable
     {
         foreach ($this->rawIterator() as $data) {
             $this->data = $this->persistence->typecastLoadRow($this, $data);
@@ -1870,10 +1863,8 @@ class Model implements \IteratorAggregate
 
     /**
      * Returns iterator.
-     *
-     * @return \Iterator
      */
-    public function rawIterator()
+    public function rawIterator(): iterable
     {
         return $this->persistence->prepareIterator($this);
     }
--- a/src/Model.php
+++ b/src/Model.php
@@ -1988,7 +1988,7 @@ class Model implements \IteratorAggregate
      * join will also query $foreign_table in order to find additional fields. When inserting
      * the record will be also added inside $foreign_table and relationship will be maintained.
      *
-     * @param array  $defaults
+     * @param array $defaults
      *
      * @return Join
      */
@@ -2013,7 +2013,7 @@ class Model implements \IteratorAggregate
      *
      * @see join()
      *
-     * @param array  $defaults
+     * @param array $defaults
      *
      * @return Join
      */
--- a/src/Model.php
+++ b/src/Model.php
@@ -1988,12 +1988,11 @@ class Model implements \IteratorAggregate
      * join will also query $foreign_table in order to find additional fields. When inserting
      * the record will be also added inside $foreign_table and relationship will be maintained.
      *
-     * @param string $foreign_table
      * @param array  $defaults
      *
      * @return Join
      */
-    public function join($foreign_table, $defaults = [])
+    public function join(string $foreign_table, $defaults = [])
     {
         if (!is_array($defaults)) {
             $defaults = ['master_field' => $defaults];
@@ -2014,12 +2013,11 @@ class Model implements \IteratorAggregate
      *
      * @see join()
      *
-     * @param string $foreign_table
      * @param array  $defaults
      *
      * @return Join
      */
-    public function leftJoin($foreign_table, $defaults = [])
+    public function leftJoin(string $foreign_table, $defaults = [])
     {
         if (!is_array($defaults)) {
             $defaults = ['master_field' => $defaults];
--- a/src/Persistence/Array_.php
+++ b/src/Persistence/Array_.php
@@ -78,7 +78,7 @@ class Array_ extends Persistence
     /**
      * Loads model and returns data record.
      *
-     * @param mixed  $id
+     * @param mixed $id
      */
     public function load(Model $m, $id, string $table = null): array
     {
@@ -99,7 +99,7 @@ class Array_ extends Persistence
      * Tries to load model and return data record.
      * Doesn't throw exception if model can't be loaded.
      *
-     * @param mixed  $id
+     * @param mixed $id
      */
     public function tryLoad(Model $m, $id, string $table = null): ?array
     {
@@ -142,7 +142,7 @@ class Array_ extends Persistence
     /**
      * Inserts record in data array and returns new record ID.
      *
-     * @param array  $data
+     * @param array $data
      *
      * @return mixed
      */
@@ -166,8 +166,8 @@ class Array_ extends Persistence
     /**
      * Updates record in data array and returns record ID.
      *
-     * @param mixed  $id
-     * @param array  $data
+     * @param mixed $id
+     * @param array $data
      *
      * @return mixed
      */
@@ -191,7 +191,7 @@ class Array_ extends Persistence
     /**
      * Deletes record in data array.
      *
-     * @param mixed  $id
+     * @param mixed $id
      */
     public function delete(Model $m, $id, string $table = null)
     {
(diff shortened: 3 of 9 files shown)


mvorisek:
@DarkSide666 This will break BC too, but it is needed to improve the project quality. Ok?

src/Action/Iterator.php
```diff
@@ -110,9 +110,9 @@ public function order($fields)
 
         // prepare arguments for array_multisort()
         $args = [];
-        foreach ($fields as list($field, $desc)) {
+        foreach ($fields as list($field, $direction)) {
             $args[] = array_column($data, $field);
-            $args[] = $desc ? SORT_DESC : SORT_ASC;
+            $args[] = strtolower($direction) === 'desc' ? SORT_DESC : SORT_ASC;
```
DarkSide666: Currently it also supported `$direction===true` or any other value which casts to true so this is BC break.
Also maybe we could allow to pass `SORT_ASC` and `SORT_DESC` here too as $direction.

src/Model.php
```diff
@@ -718,15 +714,14 @@ public function isDirty($fields = [])
...
-    public function getFields($filter = null)
+    public function getFields($filter = null): array
     {
-        if (!$filter) {
+        if ($filter === null) {
             return $this->fields;
+        } elseif (is_string($filter)) {
+            $filter = [$filter];
         }
-        $filter = is_string($filter) ? explode(',', $filter) : $filter;
```
DarkSide666: BR break - no more support string as CSV.

src/Model.php
```diff
@@ -754,33 +749,19 @@ public function getFields($filter = null)
...
-                return $this;
-            } elseif ($field instanceof self) {
-                $this->data = $field->data;
-                //$this->id = $field->id;
-
-                return $this;
-            } else {
-                $value = $field;
-                $field = $this->title_field;
+        if (func_num_args() === 1) {
```
DarkSide666: Not validating if $field is array at all (should be of course).

src/Model.php
```diff
@@ -754,33 +749,19 @@ public function getFields($filter = null)
...
      *
      * @return $this
      */
     public function set($field, $value = null)
     {
-        if (func_num_args() == 1) {
-            if (is_array($field)) {
-                foreach ($field as $key => $value) {
-                    if ($key === '0' || $key === 0) {
-                        $this->set($value);
```
DarkSide666: i wonder what was this strange functionality for...

src/Model.php
```diff
@@ -754,33 +749,19 @@ public function getFields($filter = null)
...
-                    } else {
-                        $this->set($key, $value);
-                    }
-                }
-
-                return $this;
-            } elseif ($field instanceof self) {
-                $this->data = $field->data;
```
DarkSide666: i guess this was like cloning model loaded data to another model - not sure if needed. i never used set() like that.
This can be replaced by `$model->set($another_model->get());` I guess. Or `$another_model->saveAs($model)` maybe.

src/Model.php
```diff
@@ -754,33 +749,19 @@ public function getFields($filter = null)
...
-                //$this->id = $field->id;
-
-                return $this;
-            } else {
-                $value = $field;
-                $field = $this->title_field;
```
DarkSide666: And this supported `$model->set('Foo');` which would automatically set model title_field to 'foo'. Could be handy. Not sure if we used it somewhere. Maybe in data import / lookup of related records somewhere?

src/Model.php
```diff
@@ -1302,57 +1279,53 @@ public function addWith(self $model, string $alias, array $mapping = [], bool $r
...
+     * @param string|array $field
+     * @param string       $direction "asc" or "desc"
      *
      * @return $this
      */
-    public function setOrder($field, $desc = null)
+    public function setOrder($field, string $direction = 'asc')
     {
-        // fields passed as CSV string
-        if (is_string($field) && strpos($field, ',') !== false) {
```
DarkSide666: Drop CSV support again - BC break.

src/Model.php
```diff
@@ -1954,33 +1923,17 @@ public function save($data = [], Persistence $to_persistence = null)
...
+     * @param Model $m Model where to insert
      */
-    protected function _rawInsert($m, $row)
+    protected function _rawInsert($m, array $row)
     {
         $m->reload_after_save = false;
         $m->unload();
 
-        if (!is_array($row)) {
-            $row = [$m->title_field => $row];
```
DarkSide666: BC break - drops support for importing / inserting title_fields alone.

src/Model.php
```diff
@@ -1954,33 +1923,17 @@ public function save($data = [], Persistence $to_persistence = null)
...
-        }
-
         // Find any row values that do not correspond to fields, and they may correspond to
         // references instead
         $refs = [];
         foreach ($row as $key => $value) {
-
-            // no field exists
-            if ($field = $this->hasField($key)) {
```
DarkSide666: why we drop this ?

src/Persistence/Array_.php
```diff
@@ -81,10 +81,8 @@ public function add($m, $defaults = [])
      * @param Model  $m
      * @param mixed  $id
      * @param string $table
-     *
-     * @return array|false
      */
-    public function load(Model $m, $id, $table = null)
+    public function load(Model $m, $id, $table = null): array
```
DarkSide666: Looks like $table can be only string, so we can set type here. I doubt that somebody will set table name as integer.

src/Persistence/Array_.php
```diff
@@ -110,17 +108,15 @@ public function load(Model $m, $id, $table = null)
      * @param Model  $m
      * @param mixed  $id
      * @param string $table
-     *
-     * @return array|false
      */
-    public function tryLoad(Model $m, $id, $table = null)
+    public function tryLoad(Model $m, $id, $table = null): ?array
```
DarkSide666: Looks like $table can be only string, so we can set type here. I doubt that somebody will set table name as integer.

src/Persistence/Array_.php
```diff
@@ -132,17 +128,15 @@ public function tryLoad(Model $m, $id, $table = null)
      *
      * @param Model $m
      * @param mixed $table
-     *
-     * @return array|false
      */
-    public function tryLoadAny(Model $m, $table = null)
+    public function tryLoadAny(Model $m, $table = null): ?array
```
DarkSide666: Looks like $table can be only string, so we can set type here. I doubt that somebody will set table name as integer.

src/ValidationException.php
```diff
@@ -22,32 +22,22 @@ public function __construct($errors, $model = null, $intent = null)
     {
```
DarkSide666: Please set array type for $errors argument of constructor

tests/BusinessModelTest.php
```diff
@@ -222,12 +222,9 @@ public function testSetTitle()
     {
         $m = new Model();
         $m->addField('name');
-        $m->set('foo');
+        $m->set('name', 'foo');
```
DarkSide666: I think we can/should introduce `$model->setTitle()` method which could cover this functionality. Especially because we already have `$model->getTitle()`, so having also setTitle() is only logical, right?

tests/LimitOrderTest.php
```diff
@@ -46,9 +46,8 @@ public function testReverse()
         $ii->addExpression('total_gross', '[total_net]+[total_vat]');
         $ii->getField($ii->id_field)->system = false;
 
-        // pass parameters as CSV string
         $i = clone $ii;
-        $i->setOrder('total_net desc, total_gross desc');
```
DarkSide666: In few old projects there are a lot of such syntax, but ... well, I guess it have to go...

DarkSide666 (changes requested):

src/Action/Iterator.php
mvorisek: Yes, this is slighly questionable. I discovered some bugs because of this (like 'asc' was passed and evaluated to true), but with strict types (bug need to be enable on calling side) it might be worth to isolate this change into another PR. Wdyt?

src/Model.php
mvorisek: Explode like this should never ever be allowed. Strings may contain commas (or other delimiters) itself. Very important to get rid of this.

src/Model.php
mvorisek: Yes - important one, title field should be used solely for display.

src/Persistence/Array_.php
mvorisek: This PR probably does not solve this :)

tests/BusinessModelTest.php
mvorisek: No, title is not that special, this is bad, we should not use title in data at all (only support title field for ui)

title is non-sense from data perspective, title it often also not unique and so on...

tests/LimitOrderTest.php
mvorisek: This should be never allowed

tests/BusinessModelTest.php
DarkSide666: Title is also very useful in data not just UI. And I think it's nice to have shortcut method
`$model->setTitle('foo');` instead of `$model->set($model->title_field, 'foo');`
Same as `$model->getTitle()` instead of longer `$model->get($model->title_field);` (plus fallback to $model->id)

src/Persistence/Array_.php
mvorisek: @DarkSide666 I checked now - yes, `$table` must be scalar (it is used as an array key later) and for these reasons `string` is the only valid type. But SQL does not have this parameter - do we need table at all (any strong usecase)? We can remove it, user can still set it using `Model::table` then.

DarkSide666: Maybe some use-case related to audit (historical) table.
Can't imagine anything else now which can't be solved by setting Model::table.

DarkSide666 (approved):
LGTM just please fix tests first

src/Model.php
mvorisek: seems no longer presented...

src/Model.php
mvorisek: seems no longer presented...

src/ValidationException.php
mvorisek: done

src/Persistence/Array_.php
```diff
@@ -144,14 +142,13 @@ public function tryLoadAny(Model $m, $table = null): ?array
...
      * Inserts record in data array and returns new record ID.
      *
-     * @param array  $data
-     * @param string $table
+     * @param array $data
      *
      * @return mixed
      */
-    public function insert(Model $m, $data, $table = null)
+    public function insert(Model $m, $data, string $table = null)
```
DarkSide666, line 149: no array type for $data

src/Persistence/Array_.php
```diff
@@ -169,15 +166,14 @@ public function insert(Model $m, $data, $table = null)
...
-     * @param mixed  $id
-     * @param array  $data
-     * @param string $table
+     * @param mixed $id
+     * @param array $data
      *
      * @return mixed
      */
-    public function update(Model $m, $id, $data, $table = null)
+    public function update(Model $m, $id, $data, string $table = null)
```
DarkSide666, line 174: no array type for $data

src/Persistence/Array_.php
```diff
@@ -210,14 +205,13 @@ public function delete(Model $m, $id, $table = null)
...
      * Generates new record ID.
      *
-     * @param Model  $m
-     * @param string $table
+     * @param Model $m
      *
      * @return string
      */
-    public function generateNewId($m, $table = null)
+    public function generateNewId($m, string $table = null)
```
DarkSide666, line 212: No Model type for $m

src/Persistence/Array_.php
```diff
@@ -210,14 +205,13 @@ public function delete(Model $m, $id, $table = null)
     /**
      * Generates new record ID.
      *
-     * @param Model  $m
-     * @param string $table
+     * @param Model $m
      *
      * @return string
```
DarkSide666, line 210: return type is mixed not always string

src/Persistence/Array_.php
```diff
@@ -242,23 +236,18 @@ public function generateNewId($m, $table = null)
...
      * Export all DataSet.
      *
-     * @param array|null $fields
-     * @param bool       $typecast_data Should we typecast exported data
-     *
-     * @return array
+     * @param bool $typecast_data Should we typecast exported data
      */
-    public function export(Model $m, $fields = null, $typecast_data = true)
+    public function export(Model $m, array $fields = null, $typecast_data = true): array
```
DarkSide666, line 250: why not setting bool type for $typecast_data?

src/Persistence/Array_.php
```diff
@@ -293,11 +282,8 @@ public function initAction(Model $m, $fields = null)
 
     /**
      * Will set limit defined inside $m onto data.
-     *
-     * @param Model         $m
-     * @param ArrayIterator $action
      */
-    protected function setLimitOrder($m, $action)
+    protected function setLimitOrder(Model $m, \atk4\data\Action\Iterator &$action)
```
DarkSide666: Why reference markup - $action is object so it should be already passed as reference.

src/Persistence/Csv.php
```diff
@@ -351,37 +340,34 @@ public function insert(Model $m, $data)
...
      * Updates record in data array and returns record ID.
      *
-     * @param mixed  $id
-     * @param array  $data
-     * @param string $table
+     * @param mixed $id
+     * @param array $data
      */
-    public function update(Model $m, $id, $data, $table = null)
+    public function update(Model $m, $id, $data, string $table = null)
```
DarkSide666, line 346: type array for $data ?

src/Persistence/Csv.php
```diff
@@ -351,37 +340,34 @@ public function insert(Model $m, $data)
...
      * Generates new record ID.
      *
-     * @param Model  $m
-     * @param string $table
+     * @param Model $m
      *
      * @return string
      */
-    public function generateNewId($m, $table = null)
+    public function generateNewId($m, string $table = null)
```
DarkSide666, line 368: type Model for $m?

src/Persistence/Sql.php
```diff
@@ -859,12 +851,9 @@ public function insert(Model $m, array $data)
...
      * Export all DataSet.
      *
-     * @param array|null $fields
-     * @param bool       $typecast_data Should we typecast exported data
-     *
-     * @return array
+     * @param bool $typecast_data Should we typecast exported data
      */
-    public function export(Model $m, $fields = null, $typecast_data = true)
+    public function export(Model $m, array $fields = null, $typecast_data = true): array
```
DarkSide666, line 856: type bool for $typecast_data?

DarkSide666 (changes requested):

DarkSide666 (approved):

PR #568, opened 2020-04-22
Simplify locale path getter

mvorisek:


mvorisek:
Tests should pass, retest once https://github.com/atk4/core/pull/174 is merged

tests/LocaleTest.php
```diff
@@ -19,8 +19,8 @@ public function testException()
 
     public function testGetPath()
     {
-        $path = implode(DIRECTORY_SEPARATOR, [dirname(__DIR__), 'src', '..', 'locale']).DIRECTORY_SEPARATOR;
-        $this->assertEquals($path, Locale::getPath());
+        $rootDir = realpath(dirname(__DIR__).'/src/..');
```
georgehristov, line 22: Test may fail having `/` here

mvorisek: Provide a enviroment conditions where it will fail.

Can you even imagine using it like: `require __DIR__ . DIRECTORY_SEPARATOR . 'example.php'`. I do not.

georgehristov (approved):
LGTM

PR #569, opened 2020-04-23
Model::setNull() - Allow field value unset

### LINKED ISSUE #542: Field value can not be unset (set to null) if field is not nullable
Setting of null value should be probably always allowed.

Use case: https://github.com/atk4/ui/blob/92150346e6ab5e81150d15759a2e866be38ea510/src/FormField/Lookup.php#L435

image

mvorisek:
Fix https://github.com/atk4/data/issues/542

Use case: https://github.com/atk4/ui/blob/06fc859685cdf5f9f3d19772f4305e9eac2d99c0/src/FormField/Lookup.php#L435


tests/FieldTest.php
```diff
@@ -889,4 +889,77 @@ public function testDateTimeFieldsToString()
...
+        $m->setNull('a');
+        $m->setNull('b');
+        $m->getField('c')->setNull();
+        $this->assertNull($m->get('a'));
+        $this->assertNull($m->get('b'));
+        $this->assertNull($m->get('c'));
+
+        // invalid value for set() - normalization must fail
+        $this->expectException(\atk4\data\Exception::class);
+        $m->set('c', null); // @TODO even "b"/mandatory field should fail!
```
mvorisek, line 924: @DarkSide666 Am I right, that with "b" field / mandatory field this should fail too?

DarkSide666: Yes `b` (`mandatory`) should also not allow `null` value, but allows empty string value.
Basically `mandatory` should be called `notNull` and `required` should be called `notEmpty`, but changing these property names would be big compatibility break and a lot of code to refactor again in actual projects which use atk4/data :(
Also one of these properties (not sure now which one it was) is more like UI specific property. I guess it's `required` (notEmpty) one, but again, changing that will break a lot of projects.

DarkSide666 (changes requested):

DarkSide666:
Please add:
```
        $this->expectException(\atk4\data\Exception::class);
        $m->set('b', null);
```

Besides that `setNull` method looks fine to me - like a bit of workaround, but fine.
Important is to not allow to set `null` value in _normal_ way with `set`. But if developer uses `setNull` then he should know what exactly he is doing - it can't happen accidentally, so it's fine then.

tests/FieldTest.php
mvorisek: @DarkSide666 Yes, I always need to look at the phpdoc for this. `notNull` and `notEmpty` will be perfect - ok to refactor? This project is still in very early stage. If we want to make this project big, I think we should not wait with these changes.

To the issue - can you propose a PR with fix before this is merged? The `mandatory` normalization/validation currently does not work.

mvorisek:
> 
> 
> Please add:
> 
> ```
>         $this->expectException(\atk4\data\Exception::class);
>         $m->set('b', null);
> ```
> 
> Besides that `setNull` method looks fine to me - like a bit of workaround, but fine.
> Important is to not allow to set `null` value in _normal_ way with `set`. But if developer uses `setNull` then he should know what exactly he is doing - it can't happen accidentally, so it's fine then.

See above, I want (and actually did), but the current implementation is broken...

mvorisek:
@DarkSide666 As this PR is complete, you can merge it and solve the `mandatory` bug later and change then the `924` line in tests to use `b` field.

tests/FieldTest.php
DarkSide666: Yes `notNull` and `notEmpty` would be perfect names for these properties, but changing that will break almost every single data model we have in all our projects :(
Can we change them in at least semi-compatible manner? Maybe introduce `isNotNull()` and `isNotEmpty()` methods for that and also use both property names just to keep some compatibility.
On the other hand of course it would be much better (cleaner code) to just change these property names everywhere. Also would like to have @romaninsh input on this. Sadly he's quite hard to catch lately.

DarkSide666:
> @DarkSide666 As this PR is complete, you can merge it and solve the `mandatory` bug later and change then the `924` line in tests to use `b` field.

OK I agree.

DarkSide666 (approved):

tests/FieldTest.php
mvorisek: @DarkSide666 The current design choose is actually a slightly different - `required` does not allow everything that evalueated with `empty()` to true, thus even not null, I understand that it may be handy to set one `required` property, but it is horrible design choice as `mandatory` can be false even if the field is effectively `mandatory`.

Rejecting empty values but still allowing null may be perfectly valid, at least in `atk4/data`.

Wdyt and what is your prefference? Two bool properties `notNull` and `notEmpty` ("empty" excl. null) or one enum property for all 4 possibilities?

DarkSide666: I do not see 4 possibilities here - only two.
* only restrict `null` values - everything else (empty string, zero, false etc.) is allowed
* restrict `null` *and* empty values - do not allow empty string, false, (not sure about this - i think zero should be allowed - it's legit value) zero, empty array etc.

Also there was idea that actually one of these properties more belong to UI than Data level, but it would be quite painful to move it somewhere else because of backward compatibility...

mvorisek: @DarkSide666 - no restriction, prevent null, prevent empty but allow null, prevent null and empty

DarkSide666: _prevent empty but allow null_ - I can't imagine of any use-case for this.

mvorisek: @DarkSide666 In data it is very common - null has meaning of N/A, not set yet, while empty means empty value is desired - and not always empty value should be allowed.

PR #571, opened 2020-04-23
Normalize Unit Testing workflow across repos

mvorisek:
Normalize unit testing workflow across repos.

Testing every push is important to test branches without PR.

Scheduled testing is important to provide actual build status of default branch.


mvorisek:
Trigger rules needs to be updated to not trigger twice if PR is done on branch in this repo.

mvorisek:
@DarkSide666 Can you provide feedback on this and approve it? I am still dooing small changes on this but 99% work is done.

composer.json
```diff
@@ -18,14 +18,14 @@
   "homepage": "https://github.com/atk4/data",
   "license": "MIT",
   "authors": [
-    {
-      "name": "Romans Malinovskis",
-      "email": "romans@agiletoolkit.org",
-      "homepage": "https://nearly.guru/"
-    }
+    {"name": "Romans Malinovskis", "email": "romans@agiletoolkit.org", "homepage": "https://nearly.guru/" },
+    {"name": "Michael Voříšek", "homepage": "https://mvorisek.cz/"}
```
romaninsh: @mvorisek can we set it up so the entire contributor team would get credit?

mvorisek: Yes, based on https://github.com/atk4/data/graphs/contributors at least @DarkSide666 for this repo - @DarkSide666, I added you in https://github.com/atk4/data/pull/571/commits/bb1c13734d3aced37b57ee7ebaa12ea08b8f1553 - ok?

I added authors to the other repos too and asked them for an approval to be added and to provide their homepages.

DarkSide666: Yes, I'm absolutely fine with that :)
It would be great if composer.json could read these names from atk4/image or atk4/infrastructure, to always be up to date, but I guess it's not possible.

mvorisek:
@DarkSide666 can you merge? The last big normalization PR of CS/CI.

composer.json
```diff
@@ -18,9 +18,19 @@
   "homepage": "https://github.com/atk4/data",
   "license": "MIT",
   "authors": [
-    {"name": "Romans Malinovskis", "email": "romans@agiletoolkit.org", "homepage": "https://nearly.guru/" },
-    {"name": "Imants Horsts", "homepage": "https://darkside.lv/"},
-    {"name": "Michael Voříšek", "homepage": "https://mvorisek.cz/"}
+    {
```
mvorisek: @DarkSide666 By hand, or is there any composer command to prettify/normalize composer.json?

DarkSide666: I did it by hand :)
Thing is that json will be prettified anyway when making release, but to keep changes small when releasing it's good to have already prettified json.

mvorisek: I put it there previously "minified" to lower the total real estate... And that is what I usually see also in other repos, but ok with me if you do not want to change the release processes...

DarkSide666 (approved):
Good, thanks!

PR #573, opened 2020-04-23
WIP Add super lazy BLOB type

### LINKED ISSUE #506: Add BLOB type and super-lazy load this field type by default

mvorisek:
Fixes #506

### DIFF
--- a/src/BlobValue.php
+++ b/src/BlobValue.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace atk4\data;
+
+/**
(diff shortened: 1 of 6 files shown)


mvorisek:
Closing, might reopen in the future...

Slightly related with https://github.com/atk4/data/issues/577 , this PR should solve only "lazy load" and not implementation of "binary string" (currently strings are not binary safe as for ex. "\r" are transformed to "\n")

PR #582, opened 2020-04-30
Require PHP 7.4 or higher

mvorisek:
PHP 8.0 is out in winter, is it ok to drop support for PHP 7.2 and 7.3 in favor of only PHP 7.4 and higher?

It would allow:
- typed properties
- short closures which will help to refactor all array callbacks (used a lot for hooks)

Newer PHP versions have also better trait support - see https://github.com/atk4/core/pull/120

PHP version is actually quite easy to upgrade and if someone can't/don't want, composer will prevent them to upgrade..

Please submit your feedback below:
- What PHP versions do you use?
- Do you agree that this is ok for 2.1.x?

PHP on Ubuntu and Debian can be very easily upgraded with https://deb.sury.org/#php-packages repo which is very popular and very stable also for older OS releases. On Windows, PHP can be easily upgraded with Chocolatey, WAMP, .. Almost all webhostings provides PHP 7.4 already.

If you can not upgrade easily, please comment below the reasons to better understand, like you use some packages that are not compatible with PHP 7.4 etc.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,13 +38,13 @@
     "sort-packages": true
   },
   "require": {
-    "php": ">=7.2.0",
+    "php": ">=7.4",
     "ext-intl": "*",
     "atk4/core": "dev-develop",
     "atk4/dsql": "dev-develop"
   },
   "require-release": {
-    "php": ">=7.2.0",
+    "php": ">=7.4",
     "ext-intl": "*",
     "atk4/core": "^2.0",
     "atk4/dsql": "^2.0"


DarkSide666:
No! That's to early.
Multiple projects still use PHP 7.2. Maybe can be upgraded to 7.3 in near future, but dropping 7.2 and 7.3 so soon looks real bad to me. At winter when 8.0 will be out and stable then probably we can drop 7.2 and 7.3, but not now.

mvorisek:
@DarkSide666 Yes, collecting feedback when. Guys, can we agree that once PHP 8.0.1 will be released we drop PHP 7.2 and 7.3 support?

DarkSide666:
I would say yes when 8.0.1 is out.

romaninsh:
Screenshot 2020-05-05 at 01 27 56

7.2 can be dropped EOY but not 7.3.

at the risk of repeating myself - What benefits users will have if we use typed properties and short closures? Those can be added in 1-2 years time.

mvorisek:
Orange color is for security fixes only. Currently one of the Atk issue is that there is very less typing and a lot of mistakes. When I do some refactoring to enforce some types, I spot a lot of them. Like passing strings to bools, which work on the first sight, but it is wrong.

PR #583, opened 2020-04-30
No incomplete tests by default for Sqlite and MySQL

mvorisek:
To be rebased after: https://github.com/atk4/data/pull/571

@DarkSide666 UPDATE: done


romaninsh (approved):
LG

PR #586, opened 2020-05-10
Add hinting support to Model

### LINKED ISSUE #547: Make field names refactorable/hintable
Is there currently any possibility how to refer to data with refactorability/hinting support?

If no, find a way to fix it.

It should work without the need of any generated stubs. Properties with some special phpdoc seems to be the best option.

mvorisek:
fix #547

merge after https://github.com/atk4/core/pull/192

### ⚠️ DANGER ZONE, EXTREMELY MAGICAL ENVIROMENT ⚠️

Now you can continue using everything and update your models as you want.

Then enjoy the full freedom of full IDE support.

Example usage:
```
/**
 * @property string $_name @Atk\Field(field_name="name")
 * @property string $town @Atk\Field()
 */
class Address extends HintableModel
{
    public function init(): void
    {
        parent::init();

        // i. define fields - old way
//      $this->addField('name');
//      $this->addField('town');

        // i. define fields - new/refactorable way
        $this->addField($this->fieldName()->_name); // notice different field support
        $this->addField($this->fieldName()->town);

        // both code parts are 100% equivalent
        // properties of `$this->fieldName()` magical instance are mapped
        // to their field names
    }
}

$addr = new Address();

// ii. manipulate data - old way
$addr->set('town', 'Prague');
var_dump($addr->get('town')); // will display "Prague"

// ii. manipulate data - new/refactorable way
$addr->town = 'Is the best!';
var_dump($addr->town); // will display "Prague"

// both code parts are 100% equivalent
// there is no actual/physical property in the class,
// properties of `$addr` are mapped to standard Model data
```

Want to refactor `town` to `city`? Just do it, now!, one command :)

Notice `$this->fieldName()->town` will refactor to `$this->fieldName()->city` as well.

Work in progress, but feel free to your feedback. Will be integrated to Model soon.

### DIFF
--- a/tests/ExpressionSQLTest.php
+++ b/tests/ExpressionSQLTest.php
@@ -126,13 +126,6 @@ class ExpressionSQLTest extends \atk4\schema\PhpunitTestCase
 
     public function testExpressions()
     {
-        if ($this->driverType === 'pgsql') {
-            $this->markTestIncomplete('This test is not supported on PostgreSQL');
-        }
-        if ($this->driverType === 'mysql') {
-            $this->markTestIncomplete('This test is not supported on Mysql (|| does not concatenate strings on mysql)');
-        }
-
         $a = [
             'user' => [
                 1 => ['id' => 1, 'name' => 'John', 'surname' => 'Smith', 'cached_name' => 'John Smith'],
@@ -144,12 +137,22 @@ class ExpressionSQLTest extends \atk4\schema\PhpunitTestCase
         $m = new Model($db, 'user');
         $m->addFields(['name', 'surname', 'cached_name']);
 
-        $m->addExpression('full_name', '[name] || " " || [surname]');
+        if ($this->driverType === 'sqlite') {
+            $m->addExpression('full_name', '[name] || " " || [surname]');
+        } else {
+            $m->addExpression('full_name', 'CONCAT([name], \' \', [surname])');
+        }
+
         $m->addCondition($m->expr('[full_name] != [cached_name]'));
 
         if ($this->driverType === 'sqlite') {
             $this->assertSame(
                 'select "id","name","surname","cached_name",("name" || " " || "surname") "full_name" from "user" where ("name" || " " || "surname") != "cached_name"',
+                $m->action('select')->render()
+            );
+        } elseif ($this->driverType === 'mysql') {
+            $this->assertSame(
+                'select `id`,`name`,`surname`,`cached_name`,(CONCAT(`name`, \' \', `surname`)) `full_name` from `user` where (CONCAT(`name`, \' \', `surname`)) != `cached_name`',
                 $m->action('select')->render()
             );
         }


PhilippGrashoff:
Hi,

I see the bonus of this. I guess that `get_class($addr->town)` would resolve to \atk4\data\Field\Line or something similar?

But, especially to newcomers, I doubt if the added complexity and less readable code it produces is totally worth this. What I see primarily when reading the demo code is:

```
$addr->set($addr->fieldName()->town, 'Prague');
$addr->set('town', 'Prague');
```
The second one IMO is easier to understand. Quite straightforward, whereas the first is longer and has an extra method call in it...


```
$addr->town = 'Is the best!';
```
1) What I would wonder here as someone new to ATK: There is no property called "town" in that model, so how can this work? I have an equal thought on array access to model fields (`$addr['town']`).
The feeling I have is that ATK newbies already struggle with a quite steep learning curve. This would make the curve even steeper.
2) What would happen if the Address class has a property "town" AND a field "town"? In a complex model inheriting from others, stuff like this could get accidently get messed up I think.


But, would it harm to offer coding like this and as before? Generally not I think , but I wouldn't make it the standard in demo code, at least thats my opinion now.
Im interested in seeing more of this :)

mvorisek:
> `get_class($addr->town)`
resolves to ->get('town')

demo - I do not care, I will not refactor it, but I do not thinks this is so less unreadable..

Ad 1: property is annotated, you any new user shoul click and see/understand it
Ad 2: Having different property name is already possible, also presented in the short example

In short, for the end-user (programmer) there are only 3 important points:
- you can completely opt-out + full BC
- you can use properties for data access directly (extremely usefull for quick, hintable usage + full refactoring for maintaince)
- you can use get field name (for use in other functions) using the property using `->fieldName()->` syntax

@PhilippGrashoff Do you have any related idea / feature request / improvement?

PhilippGrashoff:
> `$addr->field()->town->set('Prague');` feature can be added in a 30 seconds :)

Now this I like much more! Its much more readable IMO. What do the others think?

Looking forward to playing around with this when I finally upgrade to new versions.

mvorisek:
@PhilippGrashoff for better understanding I updated the example, `addr->field()->...` might help to access fields definitions more easily, but at least for data this way should never be prefered over direct/magical properties of model instance.

src/Hintable/HintablePropertyDef.php
```diff
@@ -6,6 +6,9 @@
 
 use atk4\data\Exception;
 
+// This code and any related parts is licenced to Michael Voříšek and anything except reading, commenting and deleting it is strictly prohibited under flat penalty of 100k EUR.
+// By continue reading/using/copying this code you fully understand it and agree.
```
romaninsh, line 10: that conflicts with the MIT license 🗡️

src/Hintable/HintablePropertyDef.php
```diff
@@ -6,6 +6,9 @@
...
+
```
DarkSide666, line 11: :) :) :)

src/Hintable/HintablePropertyDef.php
romaninsh: licensed "to" or "by" Michael ?

mvorisek: Conflict is right, my question if these and other breakig changes are wanted or not. This PR with the related PR in core can be easily put in another repo, but the other breaking changes can not. Did you come to a conclusion on this on the meeting?

mvorisek: Strictly taken, "to" is stronger, I am not allowed to put the code here!

mvorisek:
@romaninsh please delete this PR from GH as there is problem with licensing

src/Hintable/HintablePropertyDef.php
romaninsh: Thanks for clearing up. Don’t contribute other people code. ;) I’ll delete this PR as you requested.

romaninsh: Ok it appears we would need to submit GitHub support ticket for deleting this.

PR #587, opened 2020-05-11
[feature] - Setup UI executor

ibelar:
Adding method for setting up UI executor for the action.

Then we could use:
```
$action->setUiExecutor(new ActionExecutor\UserAction())
       ->onHook('afterExecute', function($ex, $return) use ($g) {
           return [
               new jsToast($return),
               $g->container->jsReload()
           ];
});
````


ibelar:
close in favor of #602

PR #589, opened 2020-05-15
Use MB functions

DarkSide666:


mvorisek:
Closing, I do not see anything to change. The tests works on known ASCII texts...

PR #591, opened 2020-05-15
Isolate ArrayAccess from Model to optional/BC trait

mvorisek:
->get('x'), ->set('x', 'v') are fully enough, originally also very unclear when to use when.

Also misleading as model is iterable on rows, not values (like you would expect when you iterate over an array).

For BC trait can be used or even the composer data installation can be patched to add this trait to Model by default.

This PR is big, but solely to isolate the ArrayAccess to make it optional, there is actually not additional different except presence/non-presence of the ArrayAccess impl. on Model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,11 +258,11 @@ While in most cases modern SQL sub-queries have comparable speed to JOIN, Agile 
 
 You can, however, [import fields through joins too](http://agile-data.readthedocs.io/en/develop/joins.html)
 
-#### Q: I don't like the `$book['field'] = 123`, I prefer properties
+#### Q: I don't like the `$book->set('field', 123)`, I prefer properties
 
 Agile Models are not Entities. They don't represent a single record, but rather a set of records. Which is why Model has some important properties: `$model->id`, `$model->persistence` and `model->data`.
 
-To simplify work with model records, you can use `$model['field']` which will be routed to `$model->data['field']`. Read more on [working with individual data records](http://agile-data.readthedocs.io/en/develop/persistence.html).
+Read more on [working with individual data records](http://agile-data.readthedocs.io/en/develop/persistence.html).
 
 #### Q: I do not like to use class `\atk4\data\Model` as a parent
 
--- a/README.md
+++ b/README.md
@@ -258,11 +258,11 @@ While in most cases modern SQL sub-queries have comparable speed to JOIN, Agile 
 
 You can, however, [import fields through joins too](http://agile-data.readthedocs.io/en/develop/joins.html)
 
-#### Q: I don't like the `$book->set('field', 123)`, I prefer properties
+#### Q: I don't like the `$book['field'] = 123`, I prefer properties
 
 Agile Models are not Entities. They don't represent a single record, but rather a set of records. Which is why Model has some important properties: `$model->id`, `$model->persistence` and `model->data`.
 
-Read more on [working with individual data records](http://agile-data.readthedocs.io/en/develop/persistence.html).
+To simplify work with model records, you can use `$model['field']` which will be routed to `$model->data['field']`. Read more on [working with individual data records](http://agile-data.readthedocs.io/en/develop/persistence.html).
 
 #### Q: I do not like to use class `\atk4\data\Model` as a parent
 
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -434,10 +434,10 @@ inside your model are unique::
                 if ($m->dirty[$field]) {
                     $mm = clone $m;
                     $mm->addCondition($mm->id_field != $this->id);
-                    $mm->tryLoadBy($field, $m->get($field));
+                    $mm->tryLoadBy($field, $m[$field]);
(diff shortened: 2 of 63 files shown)


DarkSide666:
Huge BC break as array access is used a lot for model everywhere.
I think this is not worth removing because of BC. But let @romaninsh also say his thoughts.

mvorisek:
Yes, BC break as previously one thing was solved using two solutions. I think it is time to say goodbye to some not-useful, duplicated, strange, badly defined... features.

georgehristov:
What is the problem with using `ArrayAccess` for Model?

mvorisek:
> 
> 
> What is the problem with using `ArrayAccess` for Model?

What is real advantage or having two methods for exactly the same?

georgehristov:
Flexibility
What is the benefit of removing it?
Generally more flexible frameworks are preferred especially for devs with various backgrounds and experience.
Framework resolves whatever syntax you throw at it. This reduces frustration.

mvorisek:
Flexibility of writing 10 like X. And IV like IIII. 10th century.

I can disable this locally plus I use the new hintable access :).

But in my experience, frameworks with too flexible freedom diy.

Speaking from myself, I hate having to decide between two equal things and you should never let someone do that in enviroments that should be normalized by definition.

mvorisek:
I can now confirm that with this change I was able to identify internal bugs. So one more plus of this - get is easy to seach, array access is usually not even supported by ide for refactoring.

mvorisek:
Another example of wrong usage - here used array like:
https://github.com/atk4/data/blob/90ab68ac063b8fc2c72dcd66115f1bd3f70a3a92/src/Field.php#L465
one method below used with `set()` method...

georgehristov:
I fully support we have to make code consistent and decide for a default way to use but I will not vote for leaving only one option.
Laravel for instance has `Model::getAttribute` method, uses `ArrayAccess` and supports retrieval by property name `$model->property`.
So we can even add one more way 😄

mvorisek:
Yes! But in user's solution, custom user's Model, but these can not be used anywhere outside the custom enviroments as these solutions may behave differently.

For BC we can solve this using trait which user can add if he wants to continue using this.

georgehristov:
Not sure what you mean by custom environment. For Laravel these are hardcoded default access methods in the Model class (via a trait). All three of them.

mvorisek:
We are not in Lavarel...

mvorisek:
another example:
https://github.com/atk4/ui/blob/371d09168ae294b9b4185eecf84d76f2534b885f/src/TableColumn/Link.php#L162
I acknowledge some users may be using these, but most of the usages are random - someone coded it and it just worked... This was one of misleading feature of Atk and in past I was asked about these by others - what to use, what is the difference?...

DarkSide666:
In theory I think it's good to stick to get() and set() and remove array access because that becomes even more frustrating if you use field with type=array :)
BUT what I am worried about is that this change will break a lot of other code and projects. Actually - all of them.

> For BC we can solve this using trait which user can add if he wants to continue using this.

This idea sounds good, so we still can be backward compatible and at the same time it will be easy to remove this trait after a year or so in toolkit 3.x version, for example.

mvorisek:
@DarkSide666 can we merge this? ui already fixed with https://github.com/atk4/ui/pull/1196

This PR is 100% BC if you do not use array access or you create a Model with the new optional array access trait.

If someone has some project with large number of models that extend the Model class directly, these projects can be easily fixed by patching the Model class (to use the new trait and add/implements ArrayAccess interface) via composer by `cweagans/composer-patches` pkg.

DarkSide666:
I like it now when there is also a trait for BC :)
But I have to admit that I am to lazy to review all this one more time line by line.

DarkSide666 (approved):

mvorisek:
@DarkSide666 That is what I am saying all the time - reviewing is nonse, too costly for the added value.

DarkSide666 (approved):
Please resolve conflicts and then it's good to merge.

bedengler:
How do you iterate through a multi-dimensional array with using get?
How do you read `$variable1['ab']['xy']` with get?

Is this then `$variable1->get('ab')->get('xy')`?

And how to replace nested foreach (which I'm using very often)?

```
foreach($variable1 as $key => $val) {
   $s[$key] = $val;
      foreach($val as $key1 => $val1) {
         $p[$key1] = $val1;
      }
   }
}
```

mvorisek:
@bedengler Iteration did not changed in any way - keep your iterators/foreach as they are, but instead of accessing the values with `[]`, access them thru `->get()`/`->set()`. If the 2nd level is an array however, use regular `[]`.

So `$variable1->get('ab')->get('xy')` or `$variable1->get('ab')['xy']` is valid depending on usecase (check `var_dump($val)` / `var_dump($variable1->get('ab'))` output and use the 2nd variant iff the type is array).

bedengler:
Ok, thanks!

PR #585, opened 2020-05-02
move lastInsertID to persistence

DarkSide666:
see https://github.com/atk4/dsql/pull/195

https://github.com/atk4/dsql/pull/195 should be merged in first

PR #592, opened 2020-05-15
Remove lastInsertID() from Model

mvorisek:


mvorisek:
Like having getTime() method there, nonsense... This should stay solely in `Persistence`.

What should the method provide? Did you considered inserting some related records like records from refs or log records inserted on post insert callbacks? This method can provide easily out-of-date result. Never put methods like this in Model.

DarkSide666 (changes requested):
I think this is invalid PR. Why not having this proxy method in model? It's no way worse or better than the one in Persistence class.
Also removing it will be BC break and if that change nothing then why remove?

mvorisek:
Did you read why having it on this level is absolutely wrong? - it needs to be called always in the persistence layer and as soon as possible after an insert is done.

DarkSide666:
I understand that this method shouldn't be normally used by developer in any way and it's only usage is after insert and when we insert related records. That's what I am worried about - not sure if related record functionality is all in Persistence.

mvorisek:
```
$this->persistence->lastInsertID($this)
```
Should be simply used where needed, but not via Model, Model is too "high level" for this. If you are working with persistence, you know what you are doing.

Personally, I do not use AI for anything, I use globally-unique integers distributed by a special service. Bad very long and complex story as random nor monotonic numbers are good...

mvorisek:
It is mainly in Model as it was not in Persistence before https://github.com/atk4/data/pull/585/files#diff-0f1e0d8a9aac18820a740efdb5e0f7afR1037

BC-break here is welcomed

DarkSide666 (approved):
now is good

PR #594, opened 2020-05-16
Remove NameTrait trait from Model

mvorisek:
It makes almost no sense to have it in Model - why is it there?

Added in https://github.com/atk4/data/commit/c3bd270fadbad7305fa8ca57559b404d45d8cf58 by @romaninsh

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -10,7 +10,6 @@ use atk4\core\DynamicMethodTrait;
 use atk4\core\FactoryTrait;
 use atk4\core\HookTrait;
 use atk4\core\InitializerTrait;
-use atk4\core\NameTrait;
 use atk4\core\ReadableCaptionTrait;
 use atk4\data\UserAction\Generic;
 use atk4\dsql\Query;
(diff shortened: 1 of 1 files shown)


DarkSide666 (approved):
I think it's not needed. Probably was added for historical reasons when we thought that will store models in persistence object _elements_. Like fields in model. And then we would need model->name.

mvorisek:
Perfect, now `name` is free to be used for standard `name` field/property/hinting. 👍

DarkSide666:
Not really - atk4/ui tests start to fail because of SessionTrait somehow uses these names so I reverted back at least for now.

mvorisek:
Did you checked what the name of object normally is? and if session trait should even be allowed on Model class (I am not againt), but it makes more sense on App/View only or fixing it that way `name` can be used in Models as normal `name` field/property.

mvorisek:
recreated under #596

DarkSide666:
Session trait probably is used in session persistence and probably use model name as a key in session.
Not sure - I didn't check that.

mvorisek:
I will fix it

PR #596, opened 2020-05-16
Remove NameTrait trait from Model

mvorisek:
Added in https://github.com/atk4/data/commit/c3bd270fadbad7305fa8ca57559b404d45d8cf58 by @romaninsh

`NameTrait` is Model is extremely bad as it defines `name` property. `name` field is used a lot in database models and `name` property should not stop anyone doing so - it makes sense on in `ui`, more precisely in `View` class only.

## Incompatible code

``` php
$page->add('Model\User', 'model_abc');
$user = $page->getElement('model_abc');
```

## Better code

``` php
$user = new Model\User($app->db);
```

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -10,7 +10,6 @@ use atk4\core\DynamicMethodTrait;
 use atk4\core\FactoryTrait;
 use atk4\core\HookTrait;
 use atk4\core\InitializerTrait;
-use atk4\core\NameTrait;
 use atk4\core\ReadableCaptionTrait;
 use atk4\data\UserAction\Generic;
 use atk4\dsql\Query;
@@ -30,7 +29,6 @@ class Model implements \ArrayAccess, \IteratorAggregate
     use InitializerTrait {
         init as _init;
     }
-    use NameTrait;
     use DIContainerTrait;
     use FactoryTrait;
     use AppScopeTrait;


mvorisek:
> Session trait probably is used in session persistence and probably use model name as a key in session.
Not sure - I didn't check that.

@DarkSide666 https://github.com/atk4/ui/blob/3fd37f0bee0d43f01e9eccb41e96b9dc3466d5ca/src/TableColumn/FilterModel/Generic.php#L15 the only issue is here, the SessionTrait should never be used in Model (data/Model should be responsible for backend data only, ui is for frontend) - can you fix this in ui repo?

Then this PR is ready to be merged.

DarkSide666:
I see. OK will try to refactor this somehow in UI some day soon.

romaninsh (approved):
OK to merge it as BC-break.

romaninsh:
I do not fully understand the reasoning with the "name" field, but I agree that there is no reason for Model object to have a name. I've added a BC-break so this will have a dedicated mention on the release notes and we can work around other bugs if they pop up.

mvorisek:
> I do not fully understand the reasoning

I am happy that you understand that balast code needs to be removed/isolated from something that generic like Model. 👍

PR #598, opened 2020-05-17
Add Session persistence

DarkSide666:
Implements simple session persistence which will store model data in local session.

P.S. @mvorisek it will then be used in https://github.com/atk4/ui/blob/3fd37f0bee0d43f01e9eccb41e96b9dc3466d5ca/src/TableColumn/FilterModel/Generic.php#L15 to get rid of sessionTrait and $model->name usage there.

### DIFF
--- a/src/Persistence/Session.php
+++ b/src/Persistence/Session.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace atk4\data\Persistence;
+
+/**
+ * Array persistence which will store model data in session.
+ */
+class Session extends Array_
+{
+    /**
+     * Session container key.
+     *
+     * @var string
+     */
+    protected $session_key = '__atk_session';
+
+    /**
+     * Constructor. Can pass array of data in parameters.
+     */
+    public function __construct(&$data = [], ?string $key = null)
+    {
+        $key = $key ?? $this->name ?? get_class($this);
+
+        parent::__construct($data);
+        $_SESSION[$this->session_key][$key] = &$this->data;
+    }
+}
--- a/src/Persistence/Session.php
+++ b/src/Persistence/Session.php
@@ -17,9 +17,9 @@ class Session extends Array_
     /**
      * Constructor. Can pass array of data in parameters.
      */
-    public function __construct(&$data = [], ?string $key = null)
+    public function __construct(&$data = [], string $key = null)
     {
-        $key = $key ?? $this->name ?? get_class($this);
+        $key = $key ?? $this->name ?? static::class;
 
         parent::__construct($data);
         $_SESSION[$this->session_key][$key] = &$this->data;
--- a/tests/PersistentSessionTest.php
+++ b/tests/PersistentSessionTest.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace atk4\data\tests;
+
+use atk4\core\AtkPhpunit;
+use atk4\data\Model;
+use atk4\data\Persistence;
+
+/**
+ * @coversDefaultClass \atk4\data\Persistence\Session
+ */
+class PersistentSessionTest extends AtkPhpunit\TestCase
+{
+    /**
+     * Tests.
+     */
+    public function testSession()
+    {
+        $a = [
+            'user' => [
+                1 => ['name' => 'John'],
+                2 => ['name' => 'Sarah'],
+            ],
+        ];
+
+        $p = new Persistence\Session($a);
+        $this->assertSame($a, $p->data);
+
+        $m = new Model($p, 'user');
+        $m->addField('name');
+        $this->assertSame($a, $p->data);
+        $this->assertSame($a[$m->table], $m->export());
+
+        $m->load(1);
+        $this->assertSame('John', $m['name']);
+
+        $m->load(2);
+        $m['name'] = 'Jane';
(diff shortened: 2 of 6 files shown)


mvorisek:
In short, I do not support this much. I think this actually only creates additional Persistence class which does not solve sessions from A-Z, it only constructs array persistence with $_SESSION array...

So I think this should be documented only as an example and used so.

DarkSide666:
This will simplify developers work - especially for new developers. They will be able to just use session persistence and don't care about passing $_SESSION[whatever] as data array in Array session.

mvorisek:
I fully understand, but having to maintain one extra class which does not even solve all session issues and without any possibility (excl. trait and extending both) is horrible design choose.

As proposed in ui - https://github.com/atk4/ui/pull/1127#issuecomment-629204170 - these non-full-mature or otherwise not-framework-specific-enough code should be put in another repo - and I support it, as it is good for new users and also for us, as we see what others tried to solve and how.

DarkSide666:
Of course it would be better if we could use SessionTrait here, but I couldn't imagine how to pass value as reference into session trait methods without breaking them.
Maybe you have some ideas how to improve this class?

mvorisek:
> how to pass value as reference

create anonymous class implementing ArrayAccess for session trait - but thas is bad again, as this trait can be then not added to Model or any class that does not implement Session solely - there is always only one ArrayAccess...

actually a wrapper like class for arrays will be good, then we do not have to use array references at all

but I still think manipulation with session should be not in data and it is usually managed thru something else pure $_SESSION array

mvorisek:
@DarkSide666 draft here https://github.com/atk4/data/pull/600 - wdyt? (but still do not add session here ;-) )

DarkSide666:
That looks useful in some cases at least. For example for Persistence\Array_.

Btw, why didn't you want having SessionTrait and NameTrait on Model class so badly???
Maybe we should not change anything actually?

mvorisek:
> Btw, why didn't you want having SessionTrait and NameTrait on Model class so badly???
> Maybe we should not change anything actually?

`SessionTrait` is theoretically ok - but it should never be added to Model and it currently depends on `NameTrait`.

`NameTrait` is Model is extremely bad as it defines `name` property. `name` field is used a lot in database models and `name` property should not stop anyone doing so - it makes sense on in `ui`, more precisely in `View` class only. Clear enough?

DarkSide666:
I do not see any correlation between `$model->name` as property and `$model->getField('name')` as model name field or `$model->get('name')` to get field value.
I don't think that `$model->field_name` should be allowed syntax in atk. It's even worse than ArrayAccess.

mvorisek:
> I don't think that `$model->field_name` should be allowed syntax in atk. It's even worse than ArrayAccess.

not in Atk directly, but user should have the option to do so - with current PHP, it is not possible to unset the property if the class is created at least thru reflection without constructor.

One possibility to solve the session issue is to add these two lines:
https://github.com/atk4/data/pull/598/files#diff-abdbf49c7ee4a983ed244c2d3c5aa723R24
to `App` as `sessionDb` property (there is already `db` property).
The SessionTrait  should be then removed or migrated to require the element added within `App` and use this persistence. So I support the SessionPersistence idea, but I do not think we need this extra class as it does not solve and (can not as normally managed by something else then arrays - it can be also DB!) solve the session from A-Z.

src/Persistence/Session.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Constructor. Can pass array of data in parameters.
+     */
+    public function __construct(&$data = [], string $key = null)
+    {
+        $key = $key ?? $this->name ?? static::class;
+
+        parent::__construct($data);
+        $_SESSION[$this->session_key][$key] = &$this->data;
```
romaninsh, line 25: you may need to do session_start at some point.

tests/PersistentSessionTest.php
```diff
@@ -0,0 +1,51 @@
...
+    public function testSession()
+    {
+        $a = [
+            'user' => [
+                1 => ['name' => 'John'],
+                2 => ['name' => 'Sarah'],
+            ],
+        ];
+
+        $p = new Persistence\Session($a);
```
romaninsh, line 26: not sure how this works - phpunit does not initialize session by default. Will probably fail in practice.

romaninsh (changes requested):

src/Persistence/Session.php
mvorisek: Exactly, if not set to start by default... Shoul be simply never in data...

DarkSide666: At this point headers probably are already sent.

tests/PersistentSessionTest.php
DarkSide666: but strangely it didn't fail :)
Anyway I got that - this class is not as useful as I imagined. To much confusion and cons.

src/Persistence/Session.php
mvorisek: It is very config dependent - autostart, session backend, buffering, ...

DarkSide666:
OK no need for this for now I guess. To much confusion and things what could go wrong.

PR #600, opened 2020-05-18
Improve array persistence and disallow passing array by reference

mvorisek:
full BC except:
- array was previous indended to be passed by reference
- array stored in data property is no longer accessible from public

merge after: https://github.com/atk4/data/pull/604


src/Persistence/Array_/ArrayByRef.php
```diff
@@ -0,0 +1,109 @@
+<?php
```
DarkSide666: Maybe this class could be useful somewhere else too and if so then we can move it to atk4/core ?

tests/JoinArrayTest.php
```diff
@@ -60,7 +60,7 @@ public function testDirection2()
     public function testJoinSaving1()
     {
         $a = ['user' => [], 'contact' => []];
-        $db = new Persistence\Array_($a);
+        $db = new Persistence\Array_(new Persistence\Array_\ArrayByRef($a));
```
DarkSide666: Shouldn't Persistence/Array_ automatically use ArrayByRef by default somehow? That way we wouldn't need to change this line in our projects. Or ArrayByRef would be one level up in namespace tree Persistence/ArrayByRef.

tests/PersistentArrayTest.php
```diff
@@ -159,7 +158,7 @@ public function testInsert()
             ],
         ];
 
-        $p = new Persistence\Array_($a);
+        $p = new Persistence\Array_(new Persistence\Array_\ArrayByRef($a));
```
DarkSide666: If we move ArrayByRef to atk4/core Then things like this would be simpler:
`$p = new Persistence\Array_(new \atk4\core\ArrayByRef($a));` or maybe even use static method there something like:
`$p = new Persistence\Array_(\atk4\core\Helper::ArrayByRef($a));`

DarkSide666 (commented):
Few comments below.
I would like @romaninsh to check this out too.

src/Persistence/Array_/ArrayByRef.php
mvorisek: I spend some time with it and I would advise to not to do it now - the new class can never replace arrays at all - for example set by reference is not possible (due to the ArrayAccess iface) and also there is not good type for it - the best is iterable, array accepts only arrays...

tests/JoinArrayTest.php
mvorisek: No, I am think now more about removing the ArrayByRef totally - what is the usecase of allowing arrays by reference?

You can always hack it by assigning `data` directly in array persistence...

romaninsh:
I'll have to reject this PR.

Previously when me and Dark were doing modifications to array persistence we have done it by creating a different class. I also do not see any design problem if array is passed by reference. 

The impact of this BC-break would also be pretty massive.

tests/JoinArrayTest.php
romaninsh: This looks weird. ArrayByRef shouldn't probably be this deep in the hierarchy. It's also a proxy class and some basic functionality may take a significant performance hit because of this.

mvorisek:
@romaninsh can you be specific? I think I have addressed all issues. This PR fixes also other small changes.

src/Persistence/Array_/ArrayByRef.php
```diff
@@ -0,0 +1,109 @@
...
+class ArrayByRef implements \ArrayAccess, \Iterator
+{
+    /** @var array */
+    private $arr;
+    /** @var bool */
+    private $invalidNext = false;
+    /** @var bool */
+    protected $breakReferenceOnClone;
+
+    public function __construct(array &$array, bool $breakReferenceOnClone = true)
```
romaninsh: > @mvorisek: Passing anything by reference is bad practise, also not advised to do so by Nikic in general.

mvorisek: Yes! I so so strictly againt that I have already added to the PR desc:
> If rerence is not needed at all, we can remove the ArrayByRef class completely. @DarkSide666 wdyt?

so if ok, I will remove this totally, I can also fix the tests.

romaninsh:
@mvorisek I took another, much more in-depth look into this and here is my decision and reasoning. 

## Good points
 - I agree that use of "references" is not a very good practice, so I support motivation behind this PR.
 - I like using getRawDataByTable instead of addressing data property directly
 - I like that tests focus on export()ed values instead of a specific implementation detail

## Bad points
 - I dislike that this PR introduces another reference inside a wrapper
 - I dislike that wrapper class is not functional. It virtually adds no benefit, rather appears like a "fancier way to write &"
 - I believe that this PR will have performance implications, because iteration would need to loop through the wrapper

## Suggestions
 - I'd suggest to mark Array_ persistence as "final" and provide alternative implementation for session persistence and any other persistences that try to "rely" on reference
 - I'd suggest to migrate code into StaticArray, which should implement a portion of THIS pr (minus the wrapper)
 - The above should done over a series of releases that would allow users to migrate

Bottom-line - i think using Persistence\Array(&$some_array) is bad, but instead of replacing it with Persistence\Array(wrapper(&$some_array)) we should replace it with Persistence\SomethingElse()

This approach would address the smelly code by eliminating references 100%.

mvorisek:
> Bottom-line - i think using Persistence\Array(&$some_array) is bad

I agree, I have removed the option to pass array by reference to the constructor

mvorisek:
replaced with https://github.com/atk4/data/pull/627

Issue #588, opened 2020-05-14
[proposal] - UserAction Modifier


ibelar:
There should be a way for UI, or any other external action executor, to know how to react after an action has terminated execution. 

UI elements, on the other hand, may have to handle many actions, like in the grid component, for example, thus mixing action that required acting on the grid after execution can be challenging. We could look at the return value but this is not enough for setting up proper action executor in UI in particular.

Take the 'send_statementl' action in Invoice as an example. This action required to have the model be loaded but does not change any record values and return a string instead of the model, like add or delete actions does.

Since this particular action does not modify the value of the record, i.e. only send an email, it does not make sense to reload the grid after it's execution. Thus, a special executor needs to be created in UI, apart from other action executors, for not reloading the grid.

My proposal is to add a new property ($modifier)  to the Data\UserAction class that will tell executor of the action what effect it had on the record itself. This property can be defined as constant like the scope property is.

This modifier property possible value could be: 
- insert;   when records are added;
- delete;  when records are deleted;
- edit;     when record value has changed;
- none;  when records have not been changed

It would make the life of UI executor, or any external executor, a lot easier because we could react on the modifier value once the action has executed.

mvorisek:
Are hooks not enough?

In UI, you should be able to add hooks for insert/delete/update and record what is needed...

ibelar:
@mvorisek - I think it would be easier to set executor using a specific property in action. This way, you do not need to know what the action will do, but only what it did to the record after the execution and then respond appropriately in UI or any other external executor, like an API call for example.

mvorisek:
> Since this particular action does not modify the value of the record, i.e. only send an email, it does not make sense to reload the grid after it's execution. Thus, a special executor needs to be created in UI, apart from other action executors, for not reloading the grid.

But usecase like this should be always solved with hooks.

DarkSide666:
Probably can be solved by hooks, but this is quite common core case so I think that having such property is actually a good idea. It should not contain any rules for UI, just tell whatever executor is what action just did with the record. Then UI executor can decide should it reload something or not (atk4/data don't care about that anymore).

PR #602, opened 2020-05-20
[feature] Action modifier

ibelar:
Adding modifier property to action. 
This will ease how action executor should response after an action is executed.
see proposal #588

### DIFF
--- a/src/UserAction/Generic.php
+++ b/src/UserAction/Generic.php
@@ -34,6 +34,15 @@ class Generic
     /** @var string by default - action is for a single-record */
     public $scope = self::SINGLE_RECORD;
 
+    /** Defining action modifier */
+    const CREATE = 'create'; // create new record(s).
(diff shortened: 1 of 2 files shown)


src/UserAction/Generic.php
```diff
@@ -34,6 +34,15 @@ class Generic
     /** @var string by default - action is for a single-record */
     public $scope = self::SINGLE_RECORD;
 
+    /** Defining action modifier */
+    const CREATE = 'create'; // create new record(s).
```
mvorisek: What is the purpose? If filtering/detection for some specific action set before actual execution, this is probably needed. Otherwise hook can be used with info even like what has been done.

If this is needed, declare as `public const` and rename/prefix all with `MODIFIER_`.

ibelar: done!

DarkSide666 (approved):
LGTM

PR #604, opened 2020-06-04
Remove support of default field + do not explode CSV string/value

mvorisek:


src/Model.php
```diff
@@ -1898,23 +1881,10 @@ protected function _rawInsert($m, $row)
...
 
-        if (!is_array($row)) {
-            $row = [$m->title_field => $row];
-        }
-
         // Find any row values that do not correspond to fields, and they may correspond to
         // references instead
         $refs = [];
         foreach ($row as $key => $value) {
-            // no field exists
```
DarkSide666: I don't see reason to remove this block. It's not related to default title_field or CSV and it's actually used in atk4\data\Reference\HasMany.

tests/LimitOrderTest.php
```diff
@@ -46,7 +46,6 @@ public function testReverse()
         $ii->addExpression('total_gross', '[total_net]+[total_vat]');
         $ii->getField($ii->id_field)->system = false;
 
-        // pass parameters as CSV string
```
DarkSide666, line 49: this line belongs to CSV in setOrder :)

DarkSide666 (changes requested):

tests/LimitOrderTest.php
mvorisek: True, but very unclear comment, comments like this are bad - code change a comments very often do not.

src/Model.php
mvorisek: It is actually related with CSV - see the explode there :)

DarkSide666 (approved):

PR #605, opened 2020-06-04
Use "asc"/"desc" strings for order direction instead of false/true

mvorisek:


mvorisek:
`wontfix` now, need to be carefully decised to be consistent with one clear mindset

mvorisek:
replaced with https://github.com/atk4/data/pull/626

PR #606, opened 2020-06-04
Add type validation to factory calls

mvorisek:


mvorisek:
replaced with https://github.com/atk4/data/pull/624

PR #597, opened 2020-05-16
Convert callables to closures and move hook names to constants

mvorisek:
related with https://github.com/atk4/core/pull/196

BC break as you have to use the new constants instead of the old/short hook names like `afterInsert`

## Example Old Code

``` php
$model->addHook('beforeSave', function() {});
```

## Example New code

``` php
$model->addHook(\atk4\data\Model::HOOK_BEFORE_SAVE, function() {});
```

## More info

Regex to locate old names:
```
['"](beforeLoad|afterLoad|beforeUnload|afterUnload|beforeInsert|afterInsert|beforeUpdate|afterUpdate|beforeDelete|afterDelete|beforeSave|afterSave|rollback|normalize|validate|onlyFields|afterAdd|initSelectQuery|beforeInsertQuery|afterInsertQuery|beforeUpdateQuery|afterUpdateQuery|beforeDeleteQuery|afterCopy)['"]
```

New hook names:
```
\atk4\data\Model::
'beforeLoad' => HOOK_BEFORE_LOAD
'afterLoad' => HOOK_AFTER_LOAD
'beforeUnload' => HOOK_BEFORE_UNLOAD
'afterUnload' => HOOK_AFTER_UNLOAD
'beforeInsert' => HOOK_BEFORE_INSERT
'afterInsert' => HOOK_AFTER_INSERT
'beforeUpdate' => HOOK_BEFORE_UPDATE
'afterUpdate' => HOOK_AFTER_UPDATE
'beforeDelete' => HOOK_BEFORE_DELETE
'afterDelete' => HOOK_AFTER_DELETE
'beforeSave' => HOOK_BEFORE_SAVE
'afterSave' => HOOK_AFTER_SAVE
'rollback' => HOOK_ROLLBACK
'normalize' => HOOK_NORMALIZE
'validate' => HOOK_VALIDATE
'onlyFields' => HOOK_ONLY_FIELDS

\atk4\data\Persistence::
'afterAdd' => HOOK_AFTER_ADD

\atk4\data\Persistence\SQL::
'initSelectQuery' => HOOK_INIT_SELECT_QUERY
'beforeInsertQuery' => HOOK_BEFORE_INSERT_QUERY
'afterInsertQuery' => HOOK_AFTER_INSERT_QUERY
'beforeUpdateQuery' => HOOK_BEFORE_UPDATE_QUERY
'afterUpdateQuery' => HOOK_AFTER_UPDATE_QUERY
[...]

PR #610, opened 2020-06-07
Forward-Compatibility with #597

romaninsh:
Allows a smooth transition into https://github.com/atk4/data/pull/597

This should work with v2.0 branch and your code will continue to work going into v2.1 branch.

## Example Old Code

$model->addHook('beforeSave', function() {});

## Example New code

$model->addHook(\atk4\data\Model::HOOK_BEFORE_SAVE, function() {});

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -457,7 +457,7 @@ class Model implements \ArrayAccess, \IteratorAggregate
             'fields'      => true,
             'scope'       => UserAction\Generic::NO_RECORDS,
             'callback'    => 'save',
-            'description' => 'Add '.$this->getModelCaption(),
+            'description' => 'Add ' . $this->getModelCaption(),
             'ui'          => ['icon'=>'plus'],
         ]);
         $this->addAction('edit', [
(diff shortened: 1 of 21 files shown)


src/Model.php
```diff
@@ -39,6 +39,44 @@ class Model implements \ArrayAccess, \IteratorAggregate
...
+    /** @const string */
+    public const HOOK_AFTER_DELETE = 'afterDelete';
+
+    /** @const string */
+    public const HOOK_BEFORE_SAVE = 'beforeSave';
+    /** @const string */
+    public const HOOK_AFTER_SAVE = 'afterSave';
+
+    /** @const string Executed when execution of self::atomic() failed. */
+    public const HOOK_ROLLBACK = self::class . '@rollback';
```
mvorisek: not sure it this helps

mvorisek (changes requested):
otherwise ok

mvorisek (approved):

src/Model.php
mvorisek: it will be also better to disable StyleCI. Everywhere

romaninsh: fixed.

PR #612, opened 2020-06-07
Fix Email field test

### LINKED ISSUE #574: Test FieldTypesTest::testEmail3 is failing
This is the only one test that requires DNS - DNS check function should be separated in extra method and this method should be overrided in tests.

DarkSide666:
fix #574

### DIFF
--- a/src/Field/Email.php
+++ b/src/Field/Email.php
@@ -68,13 +68,17 @@ class Email extends Field
             $p = explode('@', $email, 2);
             $user = $p[0] ?? null;
             $domain = $p[1] ?? null;
+            if (!$domain) {
+                throw new ValidationException([$this->name => 'Email address does not have domain']);
+            }
+
             if (!filter_var($user . '@' . $this->idnToAscii($domain), FILTER_VALIDATE_EMAIL)) {
-                throw new ValidationException([$this->name => 'Email format is invalid']);
+                throw new ValidationException([$this->name => 'Email address format is invalid']);
             }
 
             if ($this->dns_check) {
-                if (!$domain || !$this->isDNSValid($domain)) {
-                    throw new ValidationException([$this->name => 'Email domain does not exist']);
+                if (!$this->isDNSValid($domain)) {
+                    throw new ValidationException([$this->name => 'Email address domain does not exist']);
                 }
             }
 
@@ -94,7 +98,6 @@ class Email extends Field
 
     private function hasDNSRecord(string $domain, bool $isMX): bool
     {
-        var_dump($domain, $isMX);
         $normalizedDomain = $domain . '.';
         if (!checkdnsrr($normalizedDomain, ($isMX ? 'MX' : 'A'))) {
             return false;
@@ -102,7 +105,6 @@ class Email extends Field
 
         // dns_get_record can also return false
         $records = dns_get_record($normalizedDomain, ($isMX ? DNS_MX : DNS_A)) ?: [];
-        var_dump($records);
 
         return !empty($records);
     }
--- a/src/Field/Email.php
+++ b/src/Field/Email.php
@@ -73,9 +73,7 @@ class Email extends Field
             }
 
             if ($this->dns_check) {
-                var_dump($domain);
-                var_dump($this->idn_to_ascii($domain));
-                var_dump(checkdnsrr($this->idn_to_ascii($domain), 'MX'));
+                var_dump($this->idn_to_ascii($domain).' = '.( checkdnsrr($this->idn_to_ascii($domain), 'MX') ? 'VALID' : 'INVALID'));
                 if (!checkdnsrr($this->idn_to_ascii($domain), 'MX')) {
                     throw new ValidationException([$this->name => 'Email domain does not exist']);
                 }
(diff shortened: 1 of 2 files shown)


src/Field/Email.php
```diff
@@ -84,14 +88,31 @@ public function normalize($value)
...
+     */
+    protected function isDNSValid(string $domain): bool
+    {
+        return $this->hasDNSRecord($domain, true) || $this->hasDNSRecord($domain, false);
+    }
+
+    private function hasDNSRecord(string $domain, bool $isMX): bool
+    {
+        $normalizedDomain = $domain . '.';
+        if (!checkdnsrr($normalizedDomain, ($isMX ? 'MX' : 'A'))) {
```
mvorisek: Do we need `checkdnsrr` at all?

src/Field/Email.php
```diff
@@ -84,14 +88,31 @@ public function normalize($value)
...
+            return false;
+        }
+
+        // dns_get_record can also return false
+        $records = dns_get_record($normalizedDomain, ($isMX ? DNS_MX : DNS_A)) ?: [];
+
+        return !empty($records);
```
mvorisek: much better to test `!== false && count() > 0`, then you are do not need any comment ;-)

src/Field/Email.php
```diff
@@ -84,14 +88,31 @@ public function normalize($value)
         return parent::normalize(implode(', ', $emails));
     }
 
+    /**
+     * Validate DNS.
+     */
+    protected function isDNSValid(string $domain): bool
```
mvorisek: What about one method like `hasAnyDnsRecord(string $domain, array $types = ['MX', 'A'])`?

DarkSide666: Can do like that but what it will change? :)

src/Field/Email.php
DarkSide666: addressed

src/Field/Email.php
DarkSide666: Not sure, but I guess it's good to check and step out of method early.
Probably get_dns_records() takes more time and resources to execute.

src/Field/Email.php
mvorisek: A little cleanup, nothing more, no core functionality. I opened this mainly as this test was failing and occupying phpunit output as skipped :D

src/Field/Email.php
```diff
@@ -103,7 +103,6 @@ private function hasDNSRecord(string $domain, bool $isMX): bool
             return false;
         }
 
-        // dns_get_record can also return false
         $records = dns_get_record($normalizedDomain, ($isMX ? DNS_MX : DNS_A)) ?: [];
 
         return $records !== false && count($records) > 0;
```
mvorisek: ` ?: []` - with this you do not need `!== false`...

DarkSide666: oh yes, right :)

tests/FieldTypesTest.php
```diff
@@ -74,8 +79,9 @@ public function testEmail4()
 
         $m->set('email_name', 'Romans <me@gmail.com>');
         $m->set('email_names', 'Romans1 <me1@gmail.com>, Romans2 <me2@gmail.com>; Romans3 <me3@gmail.com>');
-        $m->set('email_idn', 'test@日本レジストリサービス.jp');
+        $m->set('email_idn', 'test@háčkyčárky.cz'); // official IDN test domain
```
mvorisek, line 82: not working with non-encoded IDN domain

src/Field/Email.php
```diff
@@ -64,17 +64,20 @@ public function normalize($value)
...
-            $p = explode('@', $email, 2);
-            $user = $p[0] ?? null;
-            $domain = $p[1] ?? null;
-            if (!filter_var($user . '@' . $this->idn_to_ascii($domain), FILTER_VALIDATE_EMAIL)) {
-                throw new ValidationException([$this->name => 'Email format is invalid']);
+            if (strpos($email, '@') === false) {
+                throw new ValidationException([$this->name => 'Email address does not have domain']);
+            }
+
+            [$user, $domain] = explode('@', $email, 2);
```
mvorisek, line 71: `@` can still be within `<>` name, but ok for now...

src/Field/Email.php
```diff
@@ -64,17 +64,20 @@ public function normalize($value)
...
+            $domain = idn_to_ascii($domain); // always convert domain to ASCII
```
mvorisek, line 72: makes sense to store in ASCII for DB normalization

src/Field/Email.php
```diff
@@ -84,15 +87,22 @@ public function normalize($value)
...
-        return idn_to_ascii($domain, IDNA_NONTRANSITIONAL_TO_ASCII, INTL_IDNA_VARIANT_UTS46);
+        foreach (array_unique(array_map('strtoupper', $types)) as $t) {
+            $dnsConsts = [
+                'MX' => DNS_MX,
+                'A' => DNS_A,
+                'AAAA' => DNS_AAAA,
+                'CNAME' => DNS_CNAME,
+            ];
+
+            $records = dns_get_record($domain . '.', $dnsConsts[$t]);
```
mvorisek, line 100: this seems to always work

tests/FieldTypesTest.php
mvorisek: was a bug in your code, you checked with ascii domain only in format

DarkSide666: Romans Yellow Horse :) :) :)

PR #613, opened 2020-06-08
Better fix for null offset limit

mvorisek:


src/Persistence/SQL.php
```diff
@@ -345,12 +345,7 @@ protected function setLimitOrder(Model $m, Query $q)
         // set limit
         if ($m->limit && ($m->limit[0] || $m->limit[1])) {
             if ($m->limit[0] === null) {
-                // This is max number which is allowed in MySQL server.
```
mvorisek, line 348: @DarkSide666 what this test should do?

mvorisek: @DarkSide666 do no know what this does?

DarkSide666: limit[0] = count of rows and limit[1] is offset. If count of rows was not set, then I had to give some huge number anyway if want to use offset. I added this solution a long time ago.
This was some limitation on max integer value you can have there.
And I somehow found out that MySQL also had some limitation.

Actually I found where I got that then. See https://dev.mysql.com/doc/refman/8.0/en/select.html

> To retrieve all rows from a certain offset up to the end of the result set, you can use some large number for the second parameter. This statement retrieves all rows from the 96th row to the last:
> SELECT * FROM tbl LIMIT 95,18446744073709551615;

DarkSide666: So I think it should stay there.

mvorisek: see PR update - should be working for all PHPs, you probably tested it previously with 32 bit PHP

DarkSide666 (approved):

PR #616, opened 2020-06-08
HasXxx() methods now return bool only + always require field to be defined first

mvorisek:
merge after: https://github.com/atk4/core/pull/233

this PR also removes option to get/set field without definition using deprecated `strict_field_check` model option

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -2449,13 +2449,11 @@ class Model implements \IteratorAggregate
     }
 
     /**
-     * Return reference field or false if reference field does not exist.
+     * Returns true if reference field exists.
      *
      * @param string $link
-     *
-     * @return Field|bool
      */
-    public function hasRef($link)
+    public function hasRef($link): bool
     {
         return $this->hasElement('#ref_' . $link);
     }
--- a/src/Model.php
+++ b/src/Model.php
@@ -315,19 +315,6 @@ class Model implements \IteratorAggregate
     public $only_fields = false;
 
     /**
-     * When set to true, you can only change the fields inside a model,
-     * that was properly declared. This helps you avoid mistake by
-     * accessing or changing the field that does not exist.
-     *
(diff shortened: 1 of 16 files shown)


romaninsh:
Check if there is anything in documentation that would need to be updated too.

mvorisek:
> Check if there is anything in documentation that would need to be updated too.

@romaninsh I try to do that, this PR is not done yet

romaninsh:
We might need a tag for PRs which obsolete things, so we can find them and remove obsolete features in V3.

src/Model.php
```diff
@@ -609,37 +594,27 @@ public function addFields(array $fields, $defaults = [])
...
 
-    /**
-     * Finds a field with a corresponding name. Returns false if field not found. Similar
-     * to hasElement() but with extra checks to make sure it's certainly a field you are
-     * getting.
-     *
-     * @return Field|false
-     */
-    public function hasField(string $name)
+    public function hasField(string $name): bool
```
DarkSide666, line 604: Why no method description at all?

src/Model.php
```diff
@@ -609,37 +594,27 @@ public function addFields(array $fields, $defaults = [])
...
 
-    /**
-     * Same as hasField, but will throw exception if field not found.
-     * Similar to getElement().
-     *
-     * @throws \atk4\core\Exception
-     *
-     * @return Field
-     */
-    public function getField(string $name)
```
DarkSide666, line 637: Why no method description at all?

DarkSide666 (approved):

src/Model.php
mvorisek: This method get's a field for you by it's name specified with string!

where is the advantage? This should be actualyl the goal ;-)

I do not say everywhere, but writing comments like these is bad. The same for describing __constructor method as "Constructor." or class X like "This is a class X for X."...

PR #617, opened 2020-06-08
Forward-Compatibility with #591 (removed array like access)

mvorisek:
Only @romaninsh can review

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,11 +29,11 @@
   "require": {
     "php": ">=7.2.0",
     "ext-intl": "*",
-    "atk4/core": "^2.0",
-    "atk4/dsql": "^2.0"
+    "atk4/core": "~2.0.0",
+    "atk4/dsql": "~2.0.0"
   },
   "require-dev": {
-    "atk4/schema": "^2.0",
+    "atk4/schema": "~2.0.0",
     "phpunit/phpunit": "<6",
     "phpunit/dbunit": ">=1.2",
     "phpunit/phpcov": "*",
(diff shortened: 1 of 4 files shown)


romaninsh:
@mvorisek, I think you need to re-create this again, too many conflicts.

As for array-access, can we implement like this (cc @DarkSide666):

``` php
foreach($user->ref('Invoices') as $row) {
  echo $row['ref'];
}

echo get_class($user->ref('Invoices'));  // Model
echo get_class($row);  // Entity
```

Additionally "Model::load(123)" would return entity. This makes it possible to refactor code in anticipation of dropping array access support and also implements a better pattern to break down Model into "set" and "entity"..

We could discuss this further, but Model->loaded() has been a thorn for a long time.

mvorisek:
@romaninsh only for forward compatibility, otherwise already fully isolated and implemented in data as well ui - and bugs discovered due this :)

your example is fully supported, if you implement array access in all your models, you can continue using array like access as long as you want

(and there should be no single merge conflict like now on GH - I opened a ticket by GH for that to fix their code as well ;-) )

romaninsh:
@mvorisek still seeing conflicts. 

Screenshot 2020-06-09 at 19 06 48

re-create or update this. anyways - i'm approving - if you create new PR i'll approve it too.

romaninsh (approved):
looks like we are on the way to drop array access, so thanks for this.

composer.json
```diff
@@ -29,11 +29,11 @@
   "require": {
     "php": ">=7.2.0",
     "ext-intl": "*",
-    "atk4/core": "^2.0",
-    "atk4/dsql": "^2.0"
+    "atk4/core": "~2.0.0",
```
mvorisek, line 32: @romaninsh I think you should require this in all `legacy/2.0` branches in all repos, right?

mvorisek:
@romaninsh after a few fixes, check and merge

PR #621, opened 2020-06-11
Add separate/explicit Model::setMulti() method, Model::set() no longer accepts array

mvorisek:
merge after https://github.com/atk4/ui/pull/1364

No method should accept different dimenisonality of the input data.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -797,11 +797,7 @@ class Model implements \IteratorAggregate
     /**
      * Helper method to call self::set() for each input array element.
      *
-     * This method does not revert the data when an exception is thrown,
-     * it is presented more for easy fix of old code which relied on self::set()
-     * to accept an array which is no longer supported.
-     *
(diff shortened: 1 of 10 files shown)


src/Model.php
```diff
@@ -727,7 +716,7 @@ public function getFields($filter = null)
      *
      * @return $this
      */
-    public function set($field, $value = null)
+    public function set(string $field, $value = null)
```
DarkSide666: No, no, no. `set(array)` is very handy and used widely.

DarkSide666 (changes requested):

src/Model.php
mvorisek: from Skype:

image


image


image

DarkSide666 (approved):

DarkSide666:
Approved, but I will try to fix tests and add doc about setMulti

PR #624, opened 2020-06-13
Add type validation to factory calls

mvorisek:


src/Model.php
```diff
@@ -1199,7 +1196,7 @@ public function asModel(string $class, array $options = []): self
      */
     public function newInstance(string $class = null, array $options = [])
     {
-        $model = $this->factory([$class ?? static::class], $options);
+        $model = (self::class)::fromSeed([$class ?? static::class], $options);
```
mvorisek, line 1199: I really hate php for this, but there is not better option now

georgehristov (approved):
LGTM

PR #627, opened 2020-06-13
Improve array persistence and disallow passing array by reference

mvorisek:
full BC except:
- array was previous indended to be passed by reference
- array stored in data property is no longer accessible from public

merge after: https://github.com/atk4/data/pull/604


src/Persistence/Array_.php
```diff
@@ -14,21 +14,23 @@
...
 class Array_ extends Persistence
 {
-    /**
-     * Array of data.
-     *
-     * @var array
-     */
-    public $data;
+    /** @var array */
+    private $data;
```
DarkSide666, line 18: Maybe better `protected` because what if I extend this class in my project? Then I still would like to access this property from inside that class.

DarkSide666 (approved):

src/Persistence/Array_.php
mvorisek: answered on Skype, can be changed later if absolutely needed, but we should target more on apis than access to a raw underlaying data...

PR #633, opened 2020-06-19
Fix generated caption for anonymous classes

mvorisek:
Table name can be secret, so we should not use it as a fallback data


georgehristov (approved):
LGTM

PR #634, opened 2020-06-20
[refactor] UserAction\Generic to Model\UserAction

### LINKED ISSUE #630: Renaming UserAction scope constants
Up for discussion/vote

I would propose for naming consistency and clarity BC renaming of `\atk4\data\UserAction\Generic` constants:

```
const SCOPE_NONE = 'none'; // e.g. add
const SCOPE_SINGLE = 'single'; // e.g. archive
const SCOPE_MULTIPLE = 'multiple'; // e.g. delete
const SCOPE_ALL = 'all'; // e.g. truncate
```

georgehristov:
- Improving naming semantics on Model User action by renaming namespace to Model\UserAction
- renaming scope constants as https://github.com/atk4/data/issues/630
- separating model methods related to actions in a trait Model\HasActionsTrait

BC break:
- `instanceof UserAction\Generic::class` should be replaced with `instanceof Model\UserAction::class`


src/Model/Action.php
```diff
@@ -0,0 +1,228 @@
...
+class Action
+{
+    use DIContainerTrait;
+    use TrackableTrait;
+    use InitializerTrait {
+        init as init_;
+    }
+
+    /** Defining scope of the action */
+    const SCOPE_NONE = 'none'; // e.g. add
```
mvorisek: wdyt about _NEW instead? Is there any other use case? (my reasoning: if not new, then it is single :) )

src/Model/Action.php
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Implements generic user action. Assigned to a model it can be invoked by a user. Model\Action class contains a
+ * meta information about the action (arguments, permissions, scope, etc) that will help UI/API or add-ons to display
+ * action trigger (button) correctly in an automated way.
+ *
+ * Action must NOT rely on any specific UI implementation.
+ *
+ * @property Model $owner
+ */
+class Action
```
mvorisek: Does actions makes sense for backend only? (or are they user UI/GUI helpers)

What is your reasoning to move it under Model?

Basically with this approach we will move everything under Model :)

georgehristov: Action is backend and UI helpers.

Namespaces will be organized as per hierarchy:
folder `Model\Scope`
class `Model\Action`
class `Model\Field`
class `Model\Field\Boolean`
class `Model\Field\Callback`, etc

The persistence part will be under `Persistence` 
`Persistence\SQL`
`Persistence\SQL\Join`
`Persistence\SQL\Field\Expression`
`Persistence\SQL\Reference\HasOne`
`Persistence\Array_`
`Persistence\Array_\Join`
`Persistence\Static_`, etc

The rest under `Util`/`Utils`

So design is: class + subfolder (with same name) for specific/extending classes
This is my intended universal NS design over all repos (as already done in atk4\dsql and atk4\schema)

src/Model/Action.php
georgehristov: We may not grasp the range of possibilities so I would keep it with original design 'NONE'.
'NEW' is just a single case of 'NONE' and will limits us.

mvorisek: The question is - is there any other case then NEW?

src/Model/Action.php
mvorisek: And Field_SQL?

georgehristov: If necessary it will be `Persistence\SQL\Field` (semantic for Field specific to SQL Persistence) but it is marked deprecated so I have not bothered moving it.

mvorisek: This seems good. Can you finish the renaming of DSQL to strict PascalCase, ie. Mysql etc.?

mvorisek: https://www.php-fig.org/psr/psr-7/ check for example this PSR from PHP and notice, they do not use `HTTP` or `URI`, but instead `Http` and `Uri`.

georgehristov: How about MySql, Postgre, etc (saw it in Laravel, to be in the middle, but I doubt you will agree 😄 )
With your logic we shall then have now HasOne but Hasone in  atk4\data?

mvorisek: No! Mysql. Would you write my_sql? No magic, no personal love or hate, clear reasoning.

mvorisek: HasOne is fine - you will write has_one :)

georgehristov: What would be the use case for converting PascalCase to snake_case?

mvorisek: There are at least 3 reasons:
- clear rule how to name it (otherwise you are inconsistent, this is actually what Microsoft and other big companies very strictly does)
- you can use it for autonaming (not only with snake_case, but also for title - with this approach, you can simply identify words easily, then you can replace the placed betteen words with anything you want - "_", "-", spaces, ...)
- Faster to read (because of previous point - you can also tell much faster where does what word start)
- and probably many others :)

So explanation phase it done. Can you now update your refactoring PRs with this naming so we can continue?

src/Model/Traits/HasActionsTrait.php
```diff
@@ -2,11 +2,11 @@
 
 declare(strict_types=1);
 
-namespace atk4\data\Model\Concerns;
+namespace atk4\data\Model\Traits;
```
mvorisek: NS should always be in singular form. Not sure if Traits trait is desired. I do not like it, also never seems it in another projects. I think Model is enought.

georgehristov: Then you will have classes and traits mixed in one folder which I do not like.
I prefer separating traits in a sub-folder. `Traits` and `Concerns` I have seen but not a `Trait` subfolder

mvorisek: Yes, I think no additional folder/NS is fine. We should be oriented on the use - nevel plural, never ever.

And `Concerns`? They says nothing to me, I do not like it.

georgehristov: Concerns is Laravel-style

mvorisek: https://stackoverflow.com/questions/40288657/how-to-make-a-trait-in-laravel - I done some research on your answer - it seems they switched from Concerns to Traits. But Without Trait suffix then.

As this clearly violates current PHP community recommendations, I am strictly against it.

georgehristov: I would agree that Model NS is already suitable for the traits (of Model) so I moved them there

src/Model/Action.php
mvorisek: ?

src/Model/Action.php
mvorisek: > Does actions makes sense for backend only? (or are they user UI/GUI helpers)

I think it may be ok to keep `UserAction` if strictly related with GUI - now it seems so, as it basically add confirmations, title, ... For new users, it is also much more explanatory.

src/Model/Action.php
georgehristov: Questuon is "could there be?". And we don't know.
So better keep it open.
Anyway this word refers to the number of records. We should not mix it with particular names of actions.
Once this refactoring series is complete I will seek to extend the functionality of Action 'scope' to accept Model\Scope object as well so action is applied to records in the scope only.

mvorisek: Ok - in sense of this PR not critical...

src/Model.php
```diff
@@ -428,34 +417,34 @@ public function init(): void
...
             'description' => 'Add ' . $this->getModelCaption(),
             'ui' => ['icon' => 'plus'],
         ]);
-        $this->addAction('edit', [
+        $this->addUserAction('edit', [
             'fields' => true,
-            'modifier' => UserAction\Generic::MODIFIER_UPDATE,
-            'scope' => UserAction\Generic::SINGLE_RECORD,
+            'modifier' => Model\UserAction::MODIFIER_UPDATE,
+            'scope' => Model\UserAction::SINGLE_RECORD,
```
mvorisek: here is `SINGLE_RECORD`, isn't is mixed with `SCOPE_`?

Do you have any ideas how to not use `SCOPE`?

tests/UserActionTest.php
```diff
@@ -154,7 +153,7 @@ public function testScope2()
     public function testScope3()
     {
         $client = new ACClient($this->pers);
-        $client->addAction('new_client', ['scope' => UserAction\Generic::NO_RECORDS, 'atomic' => false]);
+        $client->addAction('new_client', ['scope' => Model\Action::SCOPE_NONE, 'atomic' => false]);
```
mvorisek: NS seems wrong, it should be `Model\UserAction`

georgehristov: Fixed now, thanks

src/Model/Action.php
georgehristov: Switched to UserAction - it makes sense. Also changes done in atk4/ui

src/Model/UserAction.php
```diff
@@ -0,0 +1,228 @@
...
+class UserAction
+{
+    use DIContainerTrait;
+    use TrackableTrait;
+    use InitializerTrait {
+        init as init_;
+    }
+
+    /** Defining scope of the action */
+    const SCOPE_NONE = 'none'; // e.g. add
```
mvorisek: @georgehristov please explain one more time how it is related to `Scope`?

For me, it is very confusing and like with field/control terminology I think we should not use `scope` here as it refers to "when action can be applied" instead of "records filtering"

Can we change this constant to `APPLIES_TO_XXX_RECORD(S)`?

georgehristov: `AFFECTS_XXX_RECORD(S)`?

mvorisek: affects means it will change data, but I think it does not have to...

georgehristov: Yes, after a lot of thinking I changed now to APPLIES_TO prefix

src/Model.php
```diff
@@ -428,34 +417,36 @@ public function init(): void
         }
 
         // Declare our basic CRUD actions for the model.
-        $this->addAction('add', [
+        $this->addUserAction('add', [
             'fields' => true,
-            'modifier' => UserAction\Generic::MODIFIER_CREATE,
-            'scope' => UserAction\Generic::NO_RECORDS,
+            'modifier' => Model\UserAction::MODIFIER_CREATE,
+            'scope' => Model\UserAction::APPLIES_TO_NO_RECORDS,
```
mvorisek: can we change the key to `applies_to` to?

src/Model/UserAction.php
```diff
@@ -0,0 +1,228 @@
...
+/**
+ * Implements generic user action. Assigned to a model it can be invoked by a user. Model\UserAction class contains a
+ * meta information about the action (arguments, permissions, scope, etc) that will help UI/API or add-ons to display
+ * action trigger (button) correctly in an automated way.
+ *
+ * UserAction must NOT rely on any specific UI implementation.
+ *
+ * @property Model $owner
+ */
+class UserAction
```
mvorisek, line 22: (not strictly needed to be a requirement for this PR)

UserAction is related with ui only

Do you think we can easily move everything related with UserAction from data?

georgehristov: It can be moved but I believe place is here as UserActions can be used for an API for instance (without UI) so having them under atk4/data is good IMHO.

src/Model.php
georgehristov: I would propose to do this in another PR as this gets into refactoring a core functionality and better be separate.

mvorisek: I think it is ok to do it here if the change is small and if we are almost ready to merge this...

src/Model/UserAction.php
mvorisek: Makes sense to some degree, but then it should be named `Action` instead of `UserAction`.

But the current impl. is very oriented as beiing a helper for ui, see https://github.com/atk4/data/blob/develop/src/Model.php#L431 (there should be no use of `ui` in `data`, hit this several times when I was fixing some issues).

georgehristov: That is why I called it Action first. But API also has an User so it is again UserAction.
Let's keep it here and just merge the NS refactor. Now the property name is changed as well.
Further changes can be discussed on Discord with everybody

Action name also clashes with Model actions (like 'select', etc) as it is referred in the class

mvorisek: yes

mvorisek (approved):
seems ok, ui PR must be merged right after

src/Model/UserAction.php
```diff
@@ -0,0 +1,237 @@
...
+            } elseif (!is_bool($this->fields)) {
+                throw (new Exception('Argument `fields` for the action must be either array or boolean.'))
+                    ->addMoreInfo('fields', $this->fields);
+            }
+
+            // Verify some scope cases
+            switch ($this->appliesTo) {
+                case self::APPLIES_TO_NO_RECORDS:
+                    if ($this->owner->loaded()) {
+                        throw (new Exception('This action scope prevents action from being executed on existing records.'))
```
mvorisek: Should be more "User action can be executed on non-existing (new) record only."

mvorisek:
@georgehristov the deprecation messages should be fixed here too - iwthout them, xdebug nor custom error handler it is very hard to identify the issue...
image

georgehristov:
Noted, will fix it tonight.

PR #636, opened 2020-06-21
[refactor] move ModelArrayAccessTrait to correct place in NS structure

georgehristov:


src/Model/Concerns/HasArrayAccess.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace atk4\data;
+namespace atk4\data\Model\Concerns;
```
mvorisek: I do not like `Concerns`. If this is expected to be depracated, then simply `Deprecated`, but I am not sure on this (I personally do not use it and it should be never used across Atk toolkit).

src/Model/Concerns/HasArrayAccess.php
```diff
@@ -11,7 +11,7 @@
  *     use \atk4\data\ModelArrayAccessTrait;
  * }.
  */
-trait ModelArrayAccessTrait
+trait HasArrayAccess
```
mvorisek: Are you inspired with this naming by something?

- I think it is very common to put abstract (AbstractXxx, XxxBase, ...), interface (Ixxx) and Trait (XxxTrait) in the name.
- "HasXxx" is wrong to me. I much prefer "WithArrayAccessTrait". (or simply "Model\ArrayAccessTrait")

georgehristov: I would agree on `WithArrayAccess` but keep it under the Model traits folder `Model\Concerns`.
This folder already has the `HasActions` trait in it from the UserAction refactor.
As `WithArrayAccess` is for optional usage we can keep it there for devs to use if they want.

mvorisek: - `Deprecated` NS?
- `WithActions` too?
- `Trait` suffix?

georgehristov: - No need to introduce a special NS even if trait is deprecated. We use the standard NS
- `HasXxx` prefix is generally accepted
- No `Trait` suffix

mvorisek: source for 2nd two?

georgehristov: Last 2 inspired by Laravel way. Talking about what "big" others do ...

mvorisek: https://www.php-fig.org/bylaws/psr-naming-conventions/

georgehristov: If we set to follow PSR is another question, then we also need to remove the `_` in front of protected properties 😄

mvorisek: I use `_` normally only from properties that are more hacks for internal implementations, ie. properties that should not be accessed even from private.

https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions

The `Trait` suffix seems to very consistent across PHP community.

`Has` / `With` - I am still not sure with this one. `ArrayAccessTrait` `ArrayAccessibleTrait` can be also candidates. More research should be done to be really consitent with whole Atk as well as other PHP projects. I tend more to Symfony than Lavarel which contributes a lot to these standards.

mvorisek:
to be fair, I do not like `Traits` NS (I think plural is even not allowed by some PSR) nor the `With` prefix.

- I think, we do not need NS for this trait at all.

- And for the name, more research should be done

mvorisek:
closing in favor of https://github.com/atk4/data/pull/634

georgehristov:
@mvorisek but these PRs are unrelated? One is model array access trait, the other is user actions

mvorisek:
@georgehristov that is true, but you introduce another (👍 for isolating code non-strictly related with model) trait and change name also of this one...

georgehristov:
@mvorisek reopened this to relocate Model\ArrayAccessTrait to correct NS

PR #639, opened 2020-06-24
Use strict array seed

mvorisek:
no functional change, related with https://github.com/atk4/core/pull/249


mvorisek:
@georgehristov wdyt?

georgehristov (approved):
As you know I am really in favor of more versatility but referring to Seed when it is an array in specific format is good for consistency.

PR #640, opened 2020-06-27
Fix default action icon seeds

mvorisek:
related with seed refactoring

BC now and later

using `ui` in `data` is unwanted, but this is bugfix of bad design (adding a UserActions/ui code in Model/data)


ibelar (approved):
LGTM

PR #643, opened 2020-06-29
[hotfix] backward compatibility for UserAction::$scope property

georgehristov:
Needs to be merged ASAP for backward compatibility of UserAction::$scope property


mvorisek (approved):
ok, but might break app easily if both are set...

PR #650, opened 2020-07-04
Separate UserAction validation in a protected method

georgehristov:


src/Model/UserAction.php
```diff
@@ -174,6 +140,45 @@ public function execute(...$args)
         }
     }
 
+    protected function validate()
```
mvorisek: good - but it checks if action is valid to be executed, not if valid like valid :)

so rename to `validateBeforeExecute` and 👍

georgehristov: Done

mvorisek (approved):

PR #651, opened 2020-07-04
Move record related to separate Record class

mvorisek:
I prepared Model to be splitted between Model and Record. All Model properties/methods are also available virtually thru Record. Full BC, all unit tests are passing.

The goal is to separate row/record data from model - see last commit - one very simple example how load() will return data later - a Record, instead of Model. This is "still softly BC" (because all properties/methods can fallback to Model) except the returned object is no longer instance of Model (thus it can not be passed to methods when Model type is expected in their declaration).

Later, we can isolate the Model/Record logic even more to some manager and make Models more like immutable objects - data sources with defined source and conditions.

also Model foreach should return a Record then

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -224,7 +224,7 @@ class Model implements \IteratorAggregate
      *
      * @var array
      */
-    public $data = [];
+    // mapped to $this->_recordProps['data'] public $data = [];
 
     /**
      * After loading an active record from DataSet it will be stored in
@@ -240,7 +240,15 @@ class Model implements \IteratorAggregate
      *
      * @var array
      */
-    public $dirty = [];
+    // mapped to $this->_recordProps['dirty'] public $dirty = [];
+
+    /**
+     * Contains ID of the current record. If the value is null then the record
+     * is considered to be new.
+     *
+     * @var mixed
+     */
+    // mapped to $this->_recordProps['id'] public $id;
 
     /**
      * Setting model as read_only will protect you from accidentally
@@ -253,14 +261,6 @@ class Model implements \IteratorAggregate
      * @var bool
      */
     public $read_only = false;
-
-    /**
-     * Contains ID of the current record. If the value is null then the record
-     * is considered to be new.
-     *
-     * @var mixed
-     */
-    public $id;
 
     /**
      * While in most cases your id field will be called 'id', sometimes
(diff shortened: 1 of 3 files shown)


src/Model.php
```diff
@@ -240,7 +240,18 @@ class Model implements \IteratorAggregate
...
+    /**
+     * Contains ID of the current record. If the value is null then the record
+     * is considered to be new.
+     *
+     * @var mixed
+     */
+    // mapped to $this->_recordProps['id'] public $id;
+
+    /** @var Record|null Null if and only if not loaded */
+    private $_record;
```
DarkSide666, line 254: Do we need this at all? In theory I think Model should not know anything about Record. Its methods can/should return Record objects, but should not be linked to only one of them. I guess it's because of better BC, right?

DarkSide666 (commented):
Probably this is good as POC, but not as actual implementation. Looks more like workaround to implement Record class and not break compatibility. I doubt this is good to do.
I think we should set this task for some future major milestone and then should refactor Model and also Persistence to implement this properly.
For example, Model probably should not have 1:1 link to Record etc.

src/Model.php
mvorisek: This is correct, load and other loading methods should return `Record` and forget about it then. Yes, this is for migration, as currently, Model defines operations on the last loaded data/record.

mvorisek: Later, I think it should be even more splitted - record (=data), model (=definition, ie. fields/conditions) and manager (=login for data retrival/update)

mvorisek:
@DarkSide666 wdyt about returning cloned model in (more exactly in the persistence during iteration) https://github.com/atk4/data/blob/974be369185583cabe5b6114c4ceb0d4b071cd4e/src/Model.php#L1890

It will solve the issue, that Model is mutable during iteration

single record `load()` will remain as now - it is up to the user to make a clone

DarkSide666:
Oh, no way. That should be terrible performance if we clone object on each iteration.

mvorisek:
@DarkSide666 no way, I am well known that I fire people writing duplicate and slow code, so be sure every change like this one will never imply any performance issue!

so thanks for pointing this possible issue out, but I can definitely solve it before to be absolutely unmesureable vs. RTT to DB

DarkSide666:
No offense, but I can't imagine how performance can stay the same if you introduce cloning model on each iteration. If there will be 1000 iterations, then it will clone 1000 times... Even if you cache something it still will be slower than currently.

Actually even creating and returning new Record object on each iteration should be much slower than now I guess...

mvorisek:
I can cover this one :) I need feeedback if performance will be fine.

mvorisek:
this was POC/draft only, closing it, however we should probably still return another instance in load or at least in iterator shortly (theoretically we proxy field defs to the parent Model)

PR #652, opened 2020-07-07
Fix PSR class/trait/method naming

mvorisek:
related with https://github.com/atk4/core/pull/258

updated names:
```
Model::withID()
Persistence\Array_::generateNewID()
Persistence\CSV::generateNewID()
Persistence\SQL::lastInsertID()
Persistence\SQL::getFieldSQLExpression()
Join\SQL
Persistence\CSV
Persistence\SQL
Field_SQL -> FieldSql
Field_SQL_Expression -> FieldSqlExpression
Reference\HasOne_SQL -> HasOneSql
```
for 100% BC you can require the renamed classes manually (but if some other updated code required them before use is enough)


mvorisek:
@georgehristov wdyt about:
`Field_SQL` and `Field_SQL_Expression`

should we rename to `FieldSql`?

(I prefer it slightly as much consistent with other)

georgehristov (approved):

PR #654, opened 2020-07-13
Drop AppScopeTrait trait from Model & Persistence classes

mvorisek:
seems like not used, at least not for ui


DarkSide666 (approved):
Data classes should not rely on UI, so this is correct to remove that.

PR #656, opened 2020-07-15
No explicit $table param in Array & Csv persistences

mvorisek:
should have no impact on the end-user, internal cleanup only


DarkSide666 (approved):
LGTM

PR #657, opened 2020-07-15
Fix composer develop install issue

DarkSide666:


composer.json
```diff
@@ -32,6 +32,7 @@
       "homepage": "https://mvorisek.cz/"
     }
   ],
+  "version": "dev-develop",
```
romaninsh, line 35: This breaks composer

  [Composer\Json\JsonValidationException]
  "./composer.json" does not match the expected JSON schema:
   - version : Does not match the regex pattern ^v?\d+(((\.\d+)?\.\d+)?\.\d+)?

PR #660, opened 2020-07-21
Use Scope class for filtering/where impl. for Model

georgehristov:
Basic features of #551

Related to #524

### DIFF
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -318,7 +318,7 @@ CompoundCondition can be created using new CompoundCondition() statement from an
    // $condition1 will be used as child-component
 	$condition1 = new BasicCondition('name', 'like', 'ABC%');
    
-   // $condition1 will be used as child-component
+   // $condition2 will be used as child-component
 	$condition2 = new BasicCondition('country', 'US');
 	
    // $compoundCondition1 is created using AND as junction and $condition1 and $condition2 as nested conditions
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -297,15 +297,14 @@ Negates the condition, e.g::
 Sets the model of BasicCondition to a clone of $model to avoid changes to the original object.::
 
    // uses the $contact model to conver the condition to human readable words
-   $condition->on($contact)->toWords();
+   $condition->toWords($contact);
 
 .. php:method:: toWords($asHtml = false);
 
-Converts the condition object to human readable words. Model must be set first. Recommended is use of Condition::on method to set the model
-as it clones the model object first::
+Converts the condition object to human readable words. Model must be set first::
 
 	// results in 'Contact where Name is John'
-	(new BasicCondition('name', 'John'))->on($contactModel)->toWords(); 
+	(new BasicCondition('name', 'John'))->toWords($contactModel); 
 
 .. php:class:: CompoundCondition
 
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -264,7 +264,7 @@ These classes can be used directly and independently from Model class.
 
 .. php:method:: scope()
 
-This method provides access to the model scope enablind conditions to be added::
+This method provides access to the model scope enabling conditions to be added::
 
    $contact->scope()->add($condition); // adding condition to a model
 
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -2,6 +2,8 @@
 .. _DataSet:
 .. _conditions:
 
+.. php:namespace:: atk4\data
+
 ======================
 Conditions and DataSet
 ======================
@@ -51,7 +53,7 @@ Operations
 
 Most database drivers will support the following additional operations::
 
-    >, <, >=, <=, !=, in, not in
+    >, <, >=, <=, !=, in, not in, like, not like, regexp, not regexp
 
 The operation must be specified as second argument::
 
@@ -139,7 +141,7 @@ There are many other ways to set conditions, but you must always check if they
 are supported by the driver that you are using.
 
 Field Matching
--------------
+--------------
 
 Supported by: SQL   (planned for Array, Mongo)
 
@@ -164,7 +166,7 @@ inside [blah] should correspond to field names.
 
 
 SQL Expression Matching
--------------------
+-----------------------
 
 .. php:method:: expr($expression, $arguments = [])
 
@@ -248,4 +250,173 @@ do not need actual record by are only looking to traverse::
 
     $u = new Model_User($db);
     $books = $u->withId(20)->ref('Books');
+
+Advanced Usage
+==============
+
+
+Model Scope
+-----------
+
+Using the Model::addCondition method is the basic way to limit the model scope of records. Under the hood
+Agile Data utilizes a special set of classes (BasicCondition and CompoundCondition) to apply the conditions as filters on records retrieved.
+These classes can be used directly and independently from Model class.
+
+.. php:method:: scope()
+
+This method provides access to the model scope enablind conditions to be added::
+
+   $contact->scope()->add($condition); // adding condition to a model
+
+.. php:namespace:: atk4\data\Model\Scope
+
+.. php:class:: BasicCondition
+
+BasicCondition represents a simple condition in a form [field, operation, value], similar to the functionality of the 
+Model::addCondition method
+
+.. php:method:: __construct($key, $operator = null, $value = null);
+
+Creates condition object based on provided arguments. It acts similar to Model::addCondition
+
+$key can be Model field name, Field object, Expression object, FALSE (interpreted as Expression('false')), TRUE (interpreted as empty condition) or an array in the form of [$key, $operator, $value]
+$operator can be one of the supported operators >, <, >=, <=, !=, in, not in, like, not like, regexp, not regexp
+$value can be Field object, Expression object, array (interpreted as 'any of the values') or other scalar value
+
+If $value is omitted as argument then $operator is considered as $value and '=' is used as operator
+
+.. php:method:: negate();
+
+Negates the condition, e.g::
+
+	// results in 'name is not John'
+	$condition = (new BasicCondition('name', 'John'))->negate(); 
+
+.. php:method:: on(Model $model);
+
+Sets the model of BasicCondition to a clone of $model to avoid changes to the original object.::
+
+   // uses the $contact model to conver the condition to human readable words
+   $condition->on($contact)->toWords();
+
+.. php:method:: toWords($asHtml = false);
+
+Converts the condition object to human readable words. Model must be set first. Recommended is use of Condition::on method to set the model
+as it clones the model object first::
+
+	// results in 'Contact where Name is John'
+	(new BasicCondition('name', 'John'))->on($contactModel)->toWords(); 
+
+.. php:class:: CompoundCondition
+
+CompoundCondition object has a single defined junction (AND or OR) and can contain multiple nested BasicCondition and/or CompoundCondition objects referred to as nested conditions.
+This makes creating Model scopes with deep nested conditions possible, 
+e.g ((Name like 'ABC%' and Country = 'US') or (Name like 'CDE%' and (Country = 'DE' or Surname = 'XYZ')))
+
+CompoundCondition can be created using new CompoundCondition() statement from an array or joining BasicCondition objects::
+
+   // $condition1 will be used as child-component
+	$condition1 = new BasicCondition('name', 'like', 'ABC%');
+   
+   // $condition1 will be used as child-component
+	$condition2 = new BasicCondition('country', 'US');
+	
+   // $compoundCondition1 is created using AND as junction and $condition1 and $condition2 as nested conditions
+	$compoundCondition1 = CompoundCondition::mergeAnd($condition1, $condition2);
+	
+	$condition3 = new BasicCondition('country', 'DE');
+	$condition4 = new BasicCondition('surname', 'XYZ');
+	
+   // $compoundCondition2 is created using OR as junction and $condition3 and $condition4 as nested conditions
+	$compoundCondition2 = CompoundCondition::mergeOr($condition3, $condition4);
+
+	$condition5 = new BasicCondition('name', 'like', 'CDE%');
+	
+   // $compoundCondition3 is created using AND as junction and $condition5 and $compoundCondition2 as nested conditions
+	$compoundCondition3 = CompoundCondition::mergeAnd($condition5, $compoundCondition2);
+
+   // $compoundCondition is created using OR as junction and $compoundCondition1 and $compoundCondition3 as nested conditions
+	$compoundCondition = CompoundCondition::mergeOr($compoundCondition1, $compoundCondition3);
+	
+	
+CompoundCondition is an independent object not related to any model. Applying scope to model is using the Model::scope()->add($condition) method::
+
+	$contact->scope()->add($condition); // adding condition to a model
+	$contact->scope()->add($conditionXYZ); // adding more conditions
+	
+.. php:method:: __construct($nestedConditions = [], $junction = CompoundCondition::AND);
+
+Creates a CompoundCondition object from an array::
+
+	// below will create 2 conditions and nest them in a compound conditions with AND junction
+	$compoundCondition1 = new CompoundCondition([
+		['name', 'like', 'ABC%'],
+		['country', 'US']
+	]);
+	
+.. php:method:: negate();
+
+Negate method has behind the full map of conditions so any condition object can be negated, e.g negating '>=' results in '<', etc.
+For compound conditionss this method is using De Morgan's laws, e.g::
+
+	// using $compoundCondition1 defined above
+	// results in "(Name not like 'ABC%') or (Country does not equal 'US')"
+	$compoundCondition1->negate();
+
+.. php:method:: mergeAnd(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null);
+
+Merge number of conditions using AND as junction. Returns the resulting CompoundCondition object.
+
+.. php:method:: mergeOr(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null);
+
+Merge number of conditions using OR as junction. Returns the resulting CompoundCondition object.
+
+.. php:method:: simplify();
+
+Peels off single nested conditions. Useful for converting (((field = value))) to field = value.
+
+.. php:method:: clear();
+
+Clears the condition from nested conditions.
+
+.. php:method:: isOr();
+
+Checks if scope components are joined by OR
+
+.. php:method:: isAnd();
+
+Checks if scope components are joined by AND
+
+Conditions on Referenced Models
+-------------------------------
+
+Agile Data allows for adding conditions on related models for retrieval of type 'model has references where'.
+
+Setting conditions on references can be done utilizing the Model::refLink method but there is a shorthand format 
+directly integrated with addCondition method using "/" to chain the reference names::
+
+	$contact->addCondition('company/country', 'US');
+	
+This will limit the $contact model to those whose company is in US.
+'company' is the name of the reference in $contact model and 'country' is a field in the referenced model.
+
+If a condition must be set directly on the existence or number of referenced records the special symbol "#" can be
+utilized to indicate the condition is on the number of records::
+
+	$contact->addCondition('company/tickets/#', '>', 3);
+	
+This will limit the $contact model to those whose company have more than 3 tickets.
+'company' and 'tickets' are the name of the chained references ('company' is a reference in the $contact model and
+'tickets' is a reference in Company model)
+
+For applying conditions on existence of records the '?' (has any) and '!' (doesn't have any) special symbols can be used.
+Although it is similar in functionality to checking ('company/tickets/#', '>', 0) or ('company/tickets/#', '=', 0)
+'?' and '!' special symbols use optimized query and are much faster::
+
+	// Contact whose company has any tickets
+	$contact->addCondition('company/tickets/?');
+
+	// Contact whose company doesn't have any tickets
+	$contact->addCondition('company/tickets/!');
+
 
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -409,14 +409,4 @@ This will limit the $contact model to those whose company have more than 3 ticke
 'company' and 'tickets' are the name of the chained references ('company' is a reference in the $contact model and
 'tickets' is a reference in Company model)
 
-For applying conditions on existence of records the '?' (has any) and '!' (doesn't have any) special symbols can be used.
-Although it is similar in functionality to checking ('company/tickets/#', '>', 0) or ('company/tickets/#', '=', 0)
-'?' and '!' special symbols use optimized query and are much faster::
-
-	// Contact whose company has any tickets
-	$contact->addCondition('company/tickets/?');
-
-	// Contact whose company doesn't have any tickets
-	$contact->addCondition('company/tickets/!');
-
 
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -321,21 +321,21 @@ CompoundCondition can be created using new CompoundCondition() statement from an
 	$condition2 = new BasicCondition('country', 'US');
 	
    // $compoundCondition1 is created using AND as junction and $condition1 and $condition2 as nested conditions
-	$compoundCondition1 = CompoundCondition::mergeAnd($condition1, $condition2);
+	$compoundCondition1 = CompoundCondition::createAnd($condition1, $condition2);
 	
 	$condition3 = new BasicCondition('country', 'DE');
 	$condition4 = new BasicCondition('surname', 'XYZ');
 	
    // $compoundCondition2 is created using OR as junction and $condition3 and $condition4 as nested conditions
-	$compoundCondition2 = CompoundCondition::mergeOr($condition3, $condition4);
+	$compoundCondition2 = CompoundCondition::createOr($condition3, $condition4);
 
 	$condition5 = new BasicCondition('name', 'like', 'CDE%');
 	
    // $compoundCondition3 is created using AND as junction and $condition5 and $compoundCondition2 as nested conditions
-	$compoundCondition3 = CompoundCondition::mergeAnd($condition5, $compoundCondition2);
+	$compoundCondition3 = CompoundCondition::createAnd($condition5, $compoundCondition2);
 
    // $compoundCondition is created using OR as junction and $compoundCondition1 and $compoundCondition3 as nested conditions
-	$compoundCondition = CompoundCondition::mergeOr($compoundCondition1, $compoundCondition3);
+	$compoundCondition = CompoundCondition::createOr($compoundCondition1, $compoundCondition3);
 	
 	
 CompoundCondition is an independent object not related to any model. Applying scope to model is using the Model::scope()->add($condition) method::
@@ -362,11 +362,11 @@ For compound conditionss this method is using De Morgan's laws, e.g::
 	// results in "(Name not like 'ABC%') or (Country does not equal 'US')"
 	$compoundCondition1->negate();
 
-.. php:method:: mergeAnd(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null);
+.. php:method:: createAnd(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null);
 
 Merge number of conditions using AND as junction. Returns the resulting CompoundCondition object.
 
-.. php:method:: mergeOr(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null);
+.. php:method:: createOr(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null);
 
 Merge number of conditions using OR as junction. Returns the resulting CompoundCondition object.
 
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -102,7 +102,7 @@ class Iterator
 
         // simple condition
         if ($condition instanceof Model\Scope\BasicCondition) {
-            $args = $condition->toQueryArgumentsArray();
+            $args = $condition->toQueryArguments();
 
             $field = $args[0];
             $operator = $args[1] ?? null;
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -237,14 +237,14 @@ class Iterator
     /**
      * Limit Iterator.
      *
-     * @param int $length
+     * @param int $limit
      * @param int $offset
      *
      * @return $this
      */
-    public function limit($length, $offset = 0)
+    public function limit($limit, $offset = 0)
     {
-        $data = array_slice($this->get(), $offset, $length, true);
+        $data = array_slice($this->get(), $offset, $limit, true);
 
         // put data back in generator
         $this->generator = new \ArrayIterator($data);
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -102,7 +102,7 @@ class Iterator
 
         // simple condition
         if ($condition instanceof Model\Scope\BasicCondition) {
-            $args = $condition->toArray();
+            $args = $condition->toQueryArgumentsArray();
 
             $field = $args[0];
             $operator = $args[1] ?? null;
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace atk4\data\Action;
 
-use Guzzle\Iterator\FilterIterator;
+use atk4\data\Exception;
+use atk4\data\Field;
+use atk4\data\Model;
 
 /**
  * Class Array_ is returned by $model->action(). Compatible with DSQL to a certain point as it implements
@@ -33,71 +35,175 @@ class Iterator
      *
      * @return $this
      */
-    public function where($field, $value)
+    public function filter(Model\Scope\AbstractCondition $condition)
     {
-        $this->generator = new \CallbackFilterIterator($this->generator, function ($row) use ($field, $value) {
-            // skip row. does not have field at all
-            if (!array_key_exists($field, $row)) {
-                return false;
-            }
-
-            // has row and it matches
-            if ($row[$field] == $value) {
-                return true;
-            }
-
-            return false;
-        });
+        if (!$condition->isEmpty()) {
+            $this->generator = new \CallbackFilterIterator($this->generator, function ($row) use ($condition) {
+                return $this->match($row, $condition);
+            });
+        }
 
         return $this;
     }
 
     /**
-     * Applies FilterIterator condition imitating the sql LIKE operator - $field LIKE %$value% | $value% | %$value.
+     * Calculates SUM|AVG|MIN|MAX aggragate values for $field.
      *
+     * @param string $fx
      * @param string $field
-     * @param string $value
+     * @param bool   $coalesce
      *
-     * @return $this
+     * @return \atk4\data\Action\Iterator
      */
-    public function like($field, $value)
+    public function aggregate($fx, $field, $coalesce = false)
     {
-        $this->generator = new \CallbackFilterIterator($this->generator, function ($row) use ($field, $value) {
-            // skip row. does not have field at all
-            if (!array_key_exists($field, $row)) {
-                return false;
-            }
+        $result = 0;
+        $column = array_column($this->get(), $field);
 
-            $fieldValStr = (string) $row[$field];
+        switch (strtoupper($fx)) {
+            case 'SUM':
+                $result = array_sum($column);
 
-            $value = trim($value);
-            $clean_value = trim($value, '%');
-            // the row field exists check the position of the "%"(s)
-            switch ($value) {
-                // case "%str%"
-                case substr($value, -1, 1) === '%' && substr($value, 0, 1) === '%':
-                    return strpos($fieldValStr, $clean_value) !== false;
+            break;
+            case 'AVG':
+                $column = $coalesce ? $column : array_filter($column, function ($value) {
+                    return $value !== null;
+                });
 
-                    break;
-                // case "str%"
-                case substr($value, -1, 1) === '%':
-                    return substr($fieldValStr, 0, strlen($clean_value)) === $clean_value;
+                $result = array_sum($column) / count($column);
 
-                    break;
-                // case "%str"
-                case substr($value, 0, 1) === '%':
-                    return substr($fieldValStr, -strlen($clean_value)) === $clean_value;
+            break;
+            case 'MAX':
+                $result = max($column);
 
-                    break;
-                // full match
-                default:
-                    return $fieldValStr == $clean_value;
-            }
+            break;
+            case 'MIN':
+                $result = min($column);
 
-            return false;
-        });
+            break;
+            default:
+                throw (new Exception('Persistence\Array_ driver action unsupported format'))
+                    ->addMoreInfo('action', $fx);
+        }
+
+        $this->generator = new \ArrayIterator([[$result]]);
 
         return $this;
+    }
+
+    /**
+     * Checks if $row matches $condition.
+     *
+     * @return bool
+     */
+    protected function match(array $row, Model\Scope\AbstractCondition $condition)
+    {
+        $match = false;
+
+        // simple condition
+        if ($condition instanceof Model\Scope\BasicCondition) {
+            $args = $condition->toArray();
+
+            $field = $args[0];
+            $operator = $args[1] ?? null;
+            $value = $args[2] ?? null;
+            if (count($args) == 2) {
+                $value = $operator;
+
+                $operator = '=';
+            }
+
+            if (!is_a($field, Field::class)) {
+                throw (new Exception('Persistence\Array_ driver condition unsupported format'))
+                    ->addMoreInfo('reason', 'Unsupported object instance ' . get_class($field))
+                    ->addMoreInfo('condition', $condition);
+            }
+
+            if (isset($row[$field->short_name])) {
+                $match = $this->evaluateIf($row[$field->short_name], $operator, $value);
+            }
+        }
+
+        // nested conditions
+        if ($condition instanceof Model\Scope\CompoundCondition) {
+            $matches = [];
+
+            foreach ($condition->getNestedConditions() as $nestedCondition) {
+                $matches[] = $subMatch = (bool) $this->match($row, $nestedCondition);
+
+                // do not check all conditions if any match required
+                if ($condition->isOr() && $subMatch) {
+                    break;
+                }
+            }
+
+            // any matches && all matches the same (if all required)
+            $match = array_filter($matches) && ($condition->isAnd() ? count(array_unique($matches)) === 1 : true);
+        }
+
+        return $match;
+    }
+
+    protected function evaluateIf($v1, $operator, $v2): bool
+    {
+        switch (strtoupper((string) $operator)) {
+            case '=':
+                $result = is_array($v2) ? $this->evaluateIf($v1, 'IN', $v2) : $v1 == $v2;
+
+            break;
+            case '>':
+                $result = $v1 > $v2;
+
+            break;
+            case '>=':
+                $result = $v1 >= $v2;
+
+            break;
+            case '<':
+                $result = $v1 < $v2;
+
+            break;
+            case '<=':
+                $result = $v1 <= $v2;
+
+            break;
+            case '!=':
+            case '<>':
+                $result = !$this->evaluateIf($v1, '=', $v2);
+
+            break;
+            case 'LIKE':
+                $pattern = str_ireplace('%', '(.*?)', preg_quote($v2));
+
+                $result = (bool) preg_match('/^' . $pattern . '$/', (string) $v1);
+
+            break;
+            case 'NOT LIKE':
+                $result = !$this->evaluateIf($v1, 'LIKE', $v2);
+
+            break;
+            case 'IN':
+                $result = is_array($v2) ? in_array($v1, $v2, true) : $this->evaluateIf($v1, '=', $v2);
+
+            break;
+            case 'NOT IN':
+                $result = !$this->evaluateIf($v1, 'IN', $v2);
+
+            break;
+            case 'REGEXP':
+                $result = (bool) preg_match('/' . $v2 . '/', $v1);
+
+            break;
+            case 'NOT REGEXP':
+                $result = !$this->evaluateIf($v1, 'REGEXP', $v2);
+
+            break;
+            default:
+                throw (new Exception('Unsupported operator'))
+                    ->addMoreInfo('operator', $operator);
+        }
+
+        return $result;
     }
 
     /**
@@ -116,7 +222,6 @@ class Iterator
         foreach ($fields as [$field, $desc]) {
             $args[] = array_column($data, $field);
             $args[] = $desc ? SORT_DESC : SORT_ASC;
-            //$args[] = SORT_STRING; // SORT_STRING | SORT_NUMERIC | SORT_REGULAR
         }
         $args[] = &$data;
 
@@ -132,14 +237,14 @@ class Iterator
     /**
      * Limit Iterator.
      *
-     * @param int $cnt
-     * @param int $shift
+     * @param int $length
+     * @param int $offset
      *
      * @return $this
      */
-    public function limit($cnt, $shift = 0)
+    public function limit($length, $offset = 0)
     {
-        $data = array_slice($this->get(), $shift, $cnt, true);
+        $data = array_slice($this->get(), $offset, $length, true);
 
         // put data back in generator
         $this->generator = new \ArrayIterator($data);
@@ -155,6 +260,18 @@ class Iterator
     public function count()
     {
         $this->generator = new \ArrayIterator([[iterator_count($this->generator)]]);
+
+        return $this;
+    }
+
+    /**
+     * Checks if iterator has any rows.
+     *
+     * @return $this
+     */
+    public function exists()
+    {
+        $this->generator = new \ArrayIterator([[$this->generator->valid() ? 1 : 0]]);
 
         return $this;
     }
--- a/src/Action/Iterator.php
+++ b/src/Action/Iterator.php
@@ -119,9 +119,7 @@ class Iterator
                     ->addMoreInfo('condition', $condition);
             }
 
-            if (isset($row[$field->short_name])) {
-                $match = $this->evaluateIf($row[$field->short_name], $operator, $value);
-            }
+            $match = $this->evaluateIf($row[$field->short_name] ?? null, $operator, $value);
         }
 
         // nested conditions
@@ -148,7 +146,7 @@ class Iterator
     {
         switch (strtoupper((string) $operator)) {
             case '=':
-                $result = is_array($v2) ? $this->evaluateIf($v1, 'IN', $v2) : $v1 == $v2;
+                $result = is_array($v2) ? $this->evaluateIf($v1, 'IN', $v2) : $v1 === $v2;
 
             break;
             case '>':
--- a/src/Model.php
+++ b/src/Model.php
@@ -359,7 +359,7 @@ class Model implements \IteratorAggregate
      */
     public function __construct($persistence = null, $defaults = [])
     {
-        $this->scope = \Closure::bind(function() {
+        $this->scope = \Closure::bind(function () {
             return new Model\Scope();
         }, null, Model\Scope::class)();
 
--- a/src/Model.php
+++ b/src/Model.php
@@ -972,10 +972,8 @@ class Model implements \IteratorAggregate
 
     /**
      * Get the scope object of the Model.
-     *
-     * @return Model\Scope
      */
-    public function scope()
+    public function scope(): Model\Scope
     {
         return $this->scope->setModel($this);
     }
@@ -1428,7 +1426,6 @@ class Model implements \IteratorAggregate
     {
         $field = $this->getField($fieldName);
 
-        // backup
         $scopeBak = $this->scope;
         $systemBak = $field->system;
         $defaultBak = $field->default;
@@ -1440,7 +1437,6 @@ class Model implements \IteratorAggregate
 
             $this->loadAny();
         } finally {
-            // restore
             $this->scope = $scopeBak;
             $field->system = $systemBak;
             $field->default = $defaultBak;
@@ -1461,7 +1457,6 @@ class Model implements \IteratorAggregate
     {
         $field = $this->getField($fieldName);
 
-        // backup
         $scopeBak = $this->scope;
         $systemBak = $field->system;
         $defaultBak = $field->default;
@@ -1473,7 +1468,6 @@ class Model implements \IteratorAggregate
 
             $this->tryLoadAny();
         } finally {
-            // restore
             $this->scope = $scopeBak;
             $field->system = $systemBak;
             $field->default = $defaultBak;
--- a/src/Model.php
+++ b/src/Model.php
@@ -192,6 +192,9 @@ class Model implements \IteratorAggregate
      */
     public $conditions = [];
 
+    /** @var Model\Scope */
+    protected $scope;
+
     /**
      * Array of limit set.
      *
@@ -366,6 +369,8 @@ class Model implements \IteratorAggregate
      */
     public function __construct($persistence = null, $defaults = [])
     {
+        $this->scope = new Model\Scope();
+
         if ((is_string($persistence) || is_array($persistence)) && func_num_args() === 1) {
             $defaults = $persistence;
             $persistence = null;
@@ -392,6 +397,7 @@ class Model implements \IteratorAggregate
      */
     public function __clone()
     {
+        $this->scope = (clone $this->scope)->setModel($this);
         $this->_cloneCollection('elements');
         $this->_cloneCollection('fields');
         $this->_cloneCollection('userActions');
@@ -949,6 +955,15 @@ class Model implements \IteratorAggregate
      *  ->addCondition('my_field', $expr);
      *  ->addCondition($expr);
      *
+     * Conditions on referenced models are also supported:
+     *  $contact->addCondition('company/country', 'US');
+     * where 'company' is the name of the reference
+     * This will limit scope of $contact model to contacts whose company country is set to 'US'
+     *
+     * Using # in conditions on referenced model will apply the condition on the number of records:
+     * $contact->addCondition('tickets/#', '>', 5);
+     * This will limit scope of $contact model to contacts that have more than 5 tickets
+     *
      * To use those, you should consult with documentation of your
      * persistence driver.
      *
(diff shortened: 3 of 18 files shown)


docs/conditions.rst
```diff
@@ -249,3 +251,172 @@ do not need actual record by are only looking to traverse::
...
+
+Conditions on Referenced Models
+-------------------------------
+
+Agile Data allows for adding conditions on related models for retrieval of type 'model has references where'.
+
+Setting conditions on references can be done utilizing the Model::refLink method but there is a shorthand format 
+directly integrated with addCondition method using "/" to chain the reference names::
+
+	$contact->addCondition('company/country', 'US');
```
mvorisek, line 397: inspired with `/` by FS, or any popular lib? I would intuitively use `.`.

docs/conditions.rst
```diff
@@ -249,3 +251,172 @@ do not need actual record by are only looking to traverse::
...
+	
+This will limit the $contact model to those whose company is in US.
+'company' is the name of the reference in $contact model and 'country' is a field in the referenced model.
+
+If a condition must be set directly on the existence or number of referenced records the special symbol "#" can be
+utilized to indicate the condition is on the number of records::
+
+	$contact->addCondition('company/tickets/#', '>', 3);
```
mvorisek, line 405: why not `company/tickets/count(*)`? (I just want a feedback on this, `#` may have sense)

docs/conditions.rst
```diff
@@ -249,3 +251,172 @@ do not need actual record by are only looking to traverse::
...
+This will limit the $contact model to those whose company have more than 3 tickets.
+'company' and 'tickets' are the name of the chained references ('company' is a reference in the $contact model and
+'tickets' is a reference in Company model)
+
+For applying conditions on existence of records the '?' (has any) and '!' (doesn't have any) special symbols can be used.
+Although it is similar in functionality to checking ('company/tickets/#', '>', 0) or ('company/tickets/#', '=', 0)
+'?' and '!' special symbols use optimized query and are much faster::
+
+	// Contact whose company has any tickets
+	$contact->addCondition('company/tickets/?');
```
mvorisek: I think this should be not supported

with `#`, we can get query count and apply regular condition to it

and internally optimize checks of `#` > 0 to exist

`#` seems like a magical column name to me - probably ok
`?`, ... seems like a n things mixed together, actually, and, if we support unary condition (do we?, ie. without operator and value), then `#` has the same meaning

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,327 @@
...
+
+            $key = new Expression('false');
+        }
+
+        $this->key = $key;
+        $this->operator = $operator;
+        $this->value = $value;
+    }
+
+    public function onChangeModel(): void
```
mvorisek: why not setModel() ?

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,327 @@
...
+use atk4\data\Exception;
+use atk4\data\Field;
+use atk4\dsql\Expression;
+use atk4\dsql\Expressionable;
+
+class BasicCondition extends AbstractCondition
+{
+    use ReadableCaptionTrait;
+
+    public $key;
```
mvorisek, line 23: phpdoc every property like `/** @var string */`

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,209 @@
...
+        foreach ($this->elements as $nestedCondition) {
+            $nestedCondition->negate();
+        }
+
+        return $this;
+    }
+
+    public function toWords(bool $asHtml = false): string
+    {
+        if ($this->isEmpty()) {
```
mvorisek: this is unwanted code if not needed

src/Persistence/Array_.php
```diff
@@ -25,11 +24,13 @@ public function __construct(array $data = [])
...
-     *             - https://github.com/atk4/data/blob/90ab68ac063b8fc2c72dcd66115f1bd3f70a3a92/src/Reference/ContainsMany.php#L66
-     *             remove once fixed/no longer needed
+     * Array of last inserted ids per table.
+     * Last inserted ID for any table is stored under '$' key.
+     *
+     * @var array
      */
+    protected $lastInsertIds = [];
+
     public function getRawDataByTable(string $table): array
```
mvorisek: I think you removed the comment in some wrong merge in past, right? If so, introduce it back 1:1.

tests/ConditionTest.php
```diff
@@ -32,21 +32,22 @@ public function testBasicDiscrimination()
 
         $m->addCondition('gender', 'M');
 
-        $this->assertSame(1, count($m->conditions));
+        $this->assertEquals(1, count($m->scope()->getNestedConditions()));
```
mvorisek: use `assertSame`, `assertEquals` checks types weakly (and already prohibited by CS in other repos)

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,209 @@
...
+     */
+    protected $junction = self::AND;
+
+    /**
+     * Create a CompoundCondition from array of condition objects or condition arrays.
+     */
+    public function __construct(array $nestedConditions = [], $junction = self::AND)
+    {
+        // use one of JUNCTIONS values, otherwise $junction is truish means OR, falsish means AND
+        $this->junction = in_array($junction, self::JUNCTIONS, true) ? $junction : self::JUNCTIONS[$junction ? 1 : 0];
```
mvorisek: assumption instead of assertion, bad

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,209 @@
...
+     * Junction to use in case more than one element.
+     *
+     * @var self::AND|self::OR
+     */
+    protected $junction = self::AND;
+
+    /**
+     * Create a CompoundCondition from array of condition objects or condition arrays.
+     */
+    public function __construct(array $nestedConditions = [], $junction = self::AND)
```
mvorisek: phpdoc `$nestedConditions` param like `type[]`

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,327 @@
...
+    public function onChangeModel(): void
+    {
+        if ($model = $this->getModel()) {
+            // if we have a definitive scalar value for a field
+            // sets it as default value for field and locks it
+            // new records will automatically get this value assigned for the field
+            // @todo: consider this when condition is part of OR scope
+            if ($this->operator === '=' && !is_object($this->value) && !is_array($this->value)) {
+                // key containing '/' means chained references and it is handled in toArray method
+                if (is_string($field = $this->key) && stripos($field, '/') === false) {
```
mvorisek: `stripos` :)

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,327 @@
...
+    ];
+
+    public function __construct($key, $operator = null, $value = null)
+    {
+        if (func_num_args() == 2) {
+            $value = $operator;
+            $operator = '=';
+        }
+
+        if (is_bool($key)) {
```
mvorisek: usecase?

src/Model/Scope/AbstractCondition.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    abstract public function onChangeModel();
+
+    /**
+     * Temporarily assign a model to the condition.
+     *
+     * @return static
+     */
+    final public function on(Model $model)
```
mvorisek: does this method make any sense for anything other than `toWords`/`toArray`?

src/Model/Scope.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @var Model
+     */
+    protected $model;
+
+    public function __construct()
+    {
+        parent::__construct(...func_get_args());
+
+        $this->junction = self::AND;
```
mvorisek: we should definitelly define method param excl. the possibility to set junction (but cosmetic change only as used probably from Model only)

src/Model.php
```diff
@@ -1465,32 +1434,28 @@ public function tryLoadAny()
...
     {
-        // store
-        $field = $this->getField($field_name);
-        $system = $field->system;
-        $default = $field->default;
+        $field = $this->getField($fieldName);
 
-        // add condition and load record
-        $this->addCondition($field_name, $value);
+        // backup
```
mvorisek: I would remove backup/restore comment, also in tryLoadBy

src/Model.php
```diff
@@ -960,49 +975,21 @@ public function _unset(string $name)
...
-        $this->conditions[] = func_get_args();
+        $this->scope()->add(new Model\Scope\CompoundCondition([func_get_args()]));
 
         return $this;
     }
 
+    /**
+     * Get the scope object of the Model.
+     *
+     * @return Model\Scope
```
mvorisek: move type to code? (also in other placed if it makes sense)

src/Model.php
```diff
@@ -192,6 +192,9 @@ class Model implements \IteratorAggregate
      */
     public $conditions = [];
```
mvorisek: why are conditions still here?

src/Action/Iterator.php
```diff
@@ -132,14 +237,14 @@ public function order($fields)
...
      *
-     * @param int $cnt
-     * @param int $shift
+     * @param int $length
+     * @param int $offset
      *
      * @return $this
      */
-    public function limit($cnt, $shift = 0)
+    public function limit($length, $offset = 0)
```
mvorisek: count, right?

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    public function simplify()
+    {
+        if (count($this->elements) != 1) {
+            return $this;
+        }
+
+        /**
+         * @var AbstractCondition $component
```
mvorisek: use single line for inline phpdocs

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Merge number of conditions using AND as junction.
+     *
+     * @param AbstractCondition $_
+     *
+     * @return static
+     */
+    public static function mergeAnd(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null)
```
mvorisek, line 193: `fromThisAnd` might be more intuitive as static, wdyt?

docs/conditions.rst
georgehristov: This was the initial idea but it clashes with possibility to use . in field names as well.

src/Model/Scope.php
georgehristov: What do you mean exactly?

src/Model/Scope/CompoundCondition.php
georgehristov: Not sure what you mean?

src/Model/Scope/BasicCondition.php
georgehristov: `setModel` implies we are setting a model but actually we are not. Just performing actions when model is changed.
Model is only set in the `Model\Scope` object (in `$model->scope` property)

docs/conditions.rst
georgehristov: Ideas are welcome. Spent a lot of time figuring out what to use

docs/conditions.rst
georgehristov: We have 3 special symbols:
'#' -> has # referenced records
'?' -> has any referenced records
'!' -> does not have any referenced records

The '?' and '!' use Persistence `exists` action for performance purposes so they make sense.

src/Model.php
georgehristov: Fixed

src/Model.php
georgehristov: Done

src/Model.php
georgehristov: Done

src/Model/Scope/CompoundCondition.php
georgehristov: Done

src/Model/Scope/CompoundCondition.php
georgehristov: Changed

tests/ConditionTest.php
georgehristov: Done

src/Persistence/Array_.php
georgehristov: Fixed

src/Model/Scope/BasicCondition.php
georgehristov: Done

src/Model/Scope/CompoundCondition.php
georgehristov: Done

src/Action/Iterator.php
georgehristov: We limit the length of the result. The same name is used in PHP native functions as well (e.g. see `array_slice`)

docs/conditions.rst
mvorisek: I am quite strongly for `.` as it is used normally in other libs.

georgehristov: Then we have to exclude it as possible char for field name and I need @DarkSide666 review on this. We had discussions on it previously.

docs/conditions.rst
mvorisek: > and internally optimize checks of # > 0 to exist

that is why I think we do not needed, as we should optimize `#` anyway and offer exactly the same with unified design

we can even futher optimize like `(SELECT COUNT(*) FROM (SELECT 1 FROM t WHERE ... LIMIT 2)) >= 2` when the condition is `#` > 1 :)

src/Model/Scope/CompoundCondition.php
mvorisek: do not test for empty, if foreach can accept zero records and the result will be the same, this only makes the code longer and imply a little less performace in real applications

src/Model/Scope.php
mvorisek: declare function with `(array $nestedConditions = []`

and call ˙parent::__constructor($nestedConditions, self::AND)˙, better, right?

docs/conditions.rst
georgehristov: Let's start simple and then optimize if possible. I consider using the 2 additional special symbols a good solution for the extreme cases including optimization.

src/Action/Iterator.php
mvorisek: this is cosmetic change only, I did more research, `$limit` is the right name :)

src/Model/Scope/CompoundCondition.php
georgehristov: I agree to that

src/Action/Iterator.php
georgehristov: Cosmetics is all important. See what girls do with themselves 😉 
Changed now.

src/Model/Scope.php
georgehristov: It is better. Done now.

src/Model/Scope/AbstractCondition.php
georgehristov: Yes mostly for those 2 methods

src/Model/Scope/BasicCondition.php
georgehristov: Any problem with it?

src/Model/Scope/BasicCondition.php
georgehristov: `BasicCondition(false)` basically sets all records off-limit and is used in record access control.

docs/conditions.rst
mvorisek: > Let's start simple and then optimize if possible.

yes - I propose "`#` op n" only and optimize at least `= 1`, `> 0` and `>= 1` to "like ? before", and `= 0` to "like ! before"

src/Model/Scope/BasicCondition.php
mvorisek: `strpos`

src/Model/Scope/BasicCondition.php
mvorisek: I understard, there is no enable/disable now, right?

I do not like constructor that accepts "for one usecase" absolutelly different pattern/type.

what are currently other regular alternatives without allowing bool to enable/disable the condition?

as this is in constructor anyway, we can have static method like `alwaysFalse()` which can produce the same result

maybe even `BoolCondition` class with methods to easily turn it on and off

other alternative can be final immutable `AlwaysFalseCondition` class

docs/conditions.rst
georgehristov: Why do you want to remove the other special symbols?
"Everything should be made as simple as possible, but no simpler" :)

src/Model/Scope/BasicCondition.php
georgehristov: Just used to using stripos only, of course here strpos will do

src/Model/Scope/AbstractCondition.php
mvorisek: so is there anything against making `on` non-public and allowing to pass model directly to `toWords()`? (like also possibly for the future extention)

And use proper try/finally there instead of cloning...

src/Model/Scope/BasicCondition.php
georgehristov: It is not about enabling/disabling a condition. This is no longer available as option.
`BasicCondition(false)` translates to `BasicCondition(Expression('false'))`. Any expression is a legitimate condition.
In SQL terms this particular case is transformed to `SELECT * FROM xyz WHERE false AND ...`.
`BasicCondition(true)` translates to `BasicCondition()` or simply an empty condition (which is skipped when generating SQL).
It makes no sense to introduce special class or method for it, etc. This just adds unnecessary complexity.

I completely removed the activate/deactivate functionality. This is covered by the `clear` method now. 
The use case is related to placeholders and `clear` method is enough.

src/Model/Scope/BasicCondition.php
mvorisek: use php8 `str_contains`, polyfill already present in atk4/core - https://github.com/atk4/core/blob/develop/composer.json#L43

src/Model/Scope/BasicCondition.php
mvorisek: > I do not like constructor that accepts "for one usecase" absolutelly different pattern/type.

this is what should be solved - true/false should not be allowed in constructor

docs/conditions.rst
mvorisek: get rid of `?` and `!` special meaning characters

`#` is special enough

this is important for simple and understandable design and also for more freedom in the future

tests/LookupSqlTest.php
```diff
@@ -70,8 +70,8 @@ public function init(): void
             ->withTitle()
             ->addFields(['country_code' => 'code', 'is_eu']);
 
-        $this->hasMany('Friends', new LFriend())
-            ->addField('friend_names', ['field' => 'friend_name', 'concat' => ',']);
+//         $this->hasMany('Friends', new LFriend())
```
mvorisek: concat no longer supported?

src/Model/Scope/BasicCondition.php
georgehristov: What is the problem with some flexibility?
You know we are on different opinions about flexibility 😉

tests/LocaleTest.php
```diff
@@ -22,7 +22,7 @@ public function testException()
     public function testGetPath()
     {
         $rootDir = realpath(dirname(__DIR__) . '/src/..');
-        $this->assertSame($rootDir . \DIRECTORY_SEPARATOR . 'locale', realpath(Locale::getPath()));
+        $this->assertEquals($rootDir . \DIRECTORY_SEPARATOR . 'locale', realpath(Locale::getPath()));
```
mvorisek: I still see a lot `assertEquals`

src/Model/Scope/BasicCondition.php
mvorisek: Flexibility is very dangerous and high degree of freedoms is very hard to test.

Imagine code like this:
```
function getField() {
if (not loaded) {
return false;
}
}

BasicCondition(getField());
```

so if bool is not supported, you will get an error, otherwise it may take a long time to find this and flexibility equals to nighmare

so lets make a compromise - test for `func_num_args() == 1` and drop this functionality in the future if it will became common source of issues

docs/conditions.rst
mvorisek: `count(*)` might be a source of issues as `count( * )` or `count(false)` can be hard to detect or evalueate to a wrong results (if we optimize on it).

of course, we can test for `count\s*\(` regex and throw on any other non-canonical input

I like `count(*)` more as it is clear on the first sight what it does, however is may/will be confusing if other expressions are not supported (which is stronger)

or do we want to support any expressions that contains `(` like `tickets/sum(price)`? (I am slightly against it now because of added complexity)

mvorisek: if we have a data like:

company
MSFT
MMM

tickets
MSFT
MMM
MMM

and `$contact->addCondition('company/tickets/#', '>', 1);`

this should match only tickets for `MMM`, right? do we have a test for this usecase?

tests/ScopeTest.php
```diff
@@ -0,0 +1,319 @@
...
+
+        $this->assertEquals('Country Id is not equal to \'Latvia\'', $condition->on($user)->toWords());
+
+        $condition = new BasicCondition($user->getField('surname'), $user->getField('name'));
+
+        $this->assertEquals('Surname is equal to User Name', $condition->on($user)->toWords());
+
+        $country = clone $this->country;
+
+        $country->addCondition('Users/#');
```
mvorisek, line 175: need test for `$country->addCondition('#');`

src/Model/Scope/BasicCondition.php
georgehristov: Agree

tests/LocaleTest.php
georgehristov: Changed now

tests/LookupSqlTest.php
georgehristov: No changes, simply tests on this are commented out so I commented it out. Reverted now.

tests/ScopeTest.php
georgehristov: `#` is used to set condition on number of referenced records.
What would be the meaning of `$country->addCondition('#');`?

docs/conditions.rst
georgehristov: Not sure what you mean by the MSFT and MMM but based on your condition only contacts that belong to companies with more than 1 referenced tickets will be included in the scope.
So if I understand right your example only contacts of company MMM will be in the scope.
Test case is [here](https://github.com/atk4/data/blob/371d2faa1fad1ca3501f3f32d5245bd3506ba188/tests/ScopeTest.php#L186) for first level of reference. Do you mean test the second level of reference?

mvorisek: yes, test `#`, `users/#` (should be the same) and the example above, I simplified it, in reality, it will be more like:

company
id name
1 MSFT
2 MMM

tickets
id company_id
1 1
2 2
3 2

as `count(*)` aggregates, testing this is very important

tests/ScopeTest.php
mvorisek: topmost table has `#` records, but you are probably right, that in conditions, it is not that helpful - I am ok with the strict definition of "number of referenced records", but test if this throws

docs/conditions.rst
DarkSide666: If we don't implement other methods like `sum(price)`, `avg(price)` etc. this way, then I would say that `#` is much better to use than `count(*)`. It's simpler to test and simpler to write and it's quite easily understandable.

docs/conditions.rst
DarkSide666: I like `/` better than `.` because:
- dot is used in `table.field` syntax in `join()`
- theoretically dot can be used in field names `last.name`
- theoretically dot can be used in `database.table.field` syntax too somewhere or `sequence.currval()` or things like these.

So my opinion is that dot should be used for things like _properties_ or _alias.field_, but slash used for _navigating_. That way we should never mix them accidentally.

docs/conditions.rst
DarkSide666: I don't see anything bad if we use 3 special symbols, but of course it would be good if under-the-hood toolkit could optimize `#>0` to `?` and use `exists` (and other cases mentioned above).
Also need to remember that in theory we could have some persistence which doesn't support `exists` or `limit`. For example, some `Persistence_RestApi` or something like that. In that case also having multiple special symbols could (maybe) help somehow.
Anyway as I said - I don't see anything bad about having them.

tests/ScopeTest.php
DarkSide666: `$country->addCondition('Users/#');` - all countries with at least one user.
`$country->addCondition('#');` - countries with at least one country? Doesn't look useful nor spell-able :D

docs/conditions.rst
mvorisek: >     * dot is used in `table.field` syntax in `join()`

that is what I propose :) - the dot is "direct member operator"

>     * theoretically dot can be used in field names `last.name`

naming column like that is not supported by SQL nor DQL (without using backticks/escaping)

>     * theoretically dot can be used in `database.table.field` syntax too somewhere

same as above

> or `sequence.currval()` or things like these.

"direct member operator" can have 1:1 relationship

> 
> So my opinion is that dot should be used for things like _properties_ or _alias.field_, but slash used for _navigating_. That way we should never mix them accidentally.

it is about how column can be named, because `table.field` is already an navigation, but only 1:1

mvorisek: I get the point - atk is not ready for this change now, but one very good news - `/` can be very easily deprecated in the future. So let's keep `/` for traversing strictly only.

DarkSide666: > naming column like that is not supported by SQL nor DQL (without using backticks/escaping)

Toolkit automatically always adds escaping so it's _supported_ (but not well tested I guess).

BTW, if we use dot instead of slash, then this I think doesn't look understandable at all?
`$contact->addCondition('company.tickets.#', '>', 3);`

mvorisek: It does, because dot is scope is "direct member operator", ie. member can be table, field, custom/magic/internal function...

tests/ScopeTest.php
mvorisek: > `$country->addCondition('Users/#');` - all countries with at least one user.

I did not looked at the variable name, if 'Users/#' is in reality `country.users`, then

> `$country->addCondition('#');` - countries with at least one country? Doesn't look useful nor spell-able :D

should be supported as well - and spelling is easy - at least one matching record

docs/conditions.rst
DarkSide666: And in some places it even soft-escapes `table.field` to `"table"."field"` so dot is used in lot of other places in toolkit and using it here can be confusing and risky in some situations.

DarkSide666: We was discussing this with @mvorisek for about hour and came to conclusion that:
- **should use slash here**

docs/conditions.rst
DarkSide666: We was discussing this with @mvorisek for about hour and came to conclusion that:
- should use only #
- should not use ? and !. Instead internally need to optimize and use `exists` in case of `#>0` and `#=0`. In future also could optimize `#>n` to something like `(SELECT COUNT(*) FROM (SELECT 1 FROM t WHERE ... LIMIT n)) >= n`

docs/conditions.rst
mvorisek: I agree - but https://github.com/atk4/data/pull/660#discussion_r458686184 must be definitely tested, preferably also for something like `user/post/commented_by` (ie. ref. from user bad to user)

DarkSide666: We was discussing this with @mvorisek for about hour and came to conclusion that:
- currently support only #
- in future if we implement things like sum(field), avg(field), min(), max() etc, then we can also implement count(*) as longer version of #. But for now - just # is enough.

src/Model/Scope/AbstractCondition.php
DarkSide666: If it's not useful to be used public, then make private and then we will see if anyone would like to use it as public or not. It's better than making all public and afterwards set it to private and break BC for developers.

tests/ScopeTest.php
DarkSide666: > at least one matching record

What SQL select it would generate?
`SELECT * FROM country WHERE EXISTS (SELECT 1 FROM country WHERE ...)`
Looks useless.

mvorisek: `#` always translate to expr. like `(SELECT COUNT(*) FROM "table before /" WHERE ...)`

so this should answer your question, I updated my answer above - it should either be supported or throw

DarkSide666: Then throw, because I don't see any use-case for it.

mvorisek: Only if it can not be implemented easily, because in trully abstract concept I think it should not require any extra code. Only a test ;-)

docs/conditions.rst
georgehristov: I find it good idea to implement aggregates as well.
Maybe we can then use aggregares `any()` and `none()`? `#` will be the only shortcut

I am working now on level reference testing.

DarkSide666: Yes that sounds interesting - but in future, not now.
Initial implementation should be as simple and clean as possible or it will become to complex in future when adding more and more features into it. We suffered with that in toolkit code before as you can see :)

DarkSide666:
> Please don't use `str_contains` because it's supported only from PHP 8.0, but we should still support PHP 7.3 and 7.4 https://php.watch/versions/8.0/str_contains

That's solved by symphony polyfill which we have in atk4/core composer dependencies.

src/Model/Scope/BasicCondition.php
georgehristov: @DarkSide666 here is the discussion on str_contains

docs/conditions.rst
georgehristov: Yes, I will look into that separately. Target is to merge base scope functionality first then aggregates, placeholders ...

src/Model/Scope/AbstractCondition.php
georgehristov: It is necessary for public at the moment to associate a condition object with a model.
Otherwise we have to pass a model object as optional argument to `Condition::toWords` method and use it only if `Condition::getModel` is empty

src/Model.php
```diff
@@ -960,49 +965,19 @@ public function _unset(string $name)
...
 
         return $this;
     }
 
+    /**
+     * Get the scope object of the Model.
+     */
+    public function scope(): Model\Scope
+    {
+        return $this->scope->setModel($this);
```
mvorisek, line 978: In constructor, `$this` is not set, in clone, it is however set, why?

I think it should be passed in construtor and here only scope returned...

src/Model.php
```diff
@@ -960,49 +965,19 @@ public function _unset(string $name)
...
-
-                if (!is_object($v) && !is_array($v)) {
-                    $f->system = true;
-                    $f->default = $v;
-                }
-            }
-        }
-
-        $this->conditions[] = func_get_args();
+        $this->scope()->add(new Model\Scope\CompoundCondition([func_get_args()]));
```
mvorisek, line 968: I would prefer to not pass `func_get_args()` - the code can not be read easily - do we want to allow the full input here?

Also, this can add an extra level of Scope for simple condition, right? We should probably try not wrap the simple ones.

src/Model.php
```diff
@@ -1349,13 +1324,7 @@ public function withPersistence($persistence, $id = null, string $class = null)
...
     public function tryLoad($id)
     {
-        if (!$this->persistence) {
-            throw new Exception('Model is not associated with any database');
-        }
-
-        if (!$this->persistence->hasMethod('tryLoad')) {
-            throw new Exception('Persistence does not support tryLoad()');
-        }
+        $this->checkPersistence('tryLoad');
```
mvorisek, line 1327: is there anything against simplifying the passed value to `load`?

src/Model.php
```diff
@@ -1465,32 +1422,26 @@ public function tryLoadAny()
...
-            // restore
-            array_pop($this->conditions);
-            $field->system = $system;
-            $field->default = $default;
+            // add condition to cloned scope and try to load record
+            $this->scope = clone $this->scope;
+            $this->addCondition($field, $value);
 
-            throw $e;
+            $this->loadAny();
```
mvorisek: slighly better CS would be to return here and remove `return $this;` below

2x

src/Model/Scope.php
```diff
@@ -0,0 +1,45 @@
...
+ * Scope elements are joined only using AND junction.
+ */
+class Scope extends Scope\CompoundCondition
+{
+    /**
+     * @var Model
+     */
+    protected $model;
+
+    public function __construct(array $nestedConditions = [])
```
mvorisek: should topmost/Model Scope be allowed to be instatiated from something else than Model?

tests/ConditionSqlTest.php
```diff
@@ -391,12 +391,12 @@ public function testLoadBy()
         $u = (new Model($this->db, 'user'))->addFields(['name']);
 
         $u->loadBy('name', 'John');
-        $this->assertSame([], $u->conditions); // should be no conditions
+        $this->assertTrue($u->scope()->isEmpty()); // should be no conditions
```
mvorisek: comment not needed, 2x

src/Persistence/Sql.php
```diff
@@ -364,60 +364,36 @@ protected function setLimitOrder(Model $m, Query $q)
     }
 
     /**
-     * Will apply conditions defined inside $m onto query $q.
+     * Will apply scope defined inside $scope or $model->scope() onto $query.
```
mvorisek: what is `$scope`?

mvorisek: why do we need `$condition` as all?

src/Persistence/Sql.php
```diff
@@ -364,60 +364,36 @@ protected function setLimitOrder(Model $m, Query $q)
...
-                        if ($row[0] instanceof Field && (count($row) === 2 || !in_array(strtolower($row[1]), ['like', 'regexp'], true))) {
-                            $valueKey = count($row) === 2 ? 1 : 2;
-                            $row[$valueKey] = $this->typecastSaveField($row[0], $row[$valueKey]);
-                        }
-                    }
-                }
+        $condition = $condition ?? $model->scope();
 
-                $q->where($cond[0]);
+        if (!$condition || $condition->isEmpty()) {
```
mvorisek: `!$condition` seems not needed

src/Persistence/Sql.php
```diff
@@ -631,14 +607,14 @@ public function action(Model $m, $type, $args = [])
...
-                $m->hook(self::HOOK_INIT_SELECT_QUERY, [$q]);
-                if (isset($args['alias'])) {
-                    $q->reset('field')->field('count(*)', $args['alias']);
-                } else {
-                    $q->reset('field')->field('count(*)');
-                }
+                $m->hook(self::HOOK_INIT_SELECT_QUERY, [$q, $type]);
 
-                return $q;
+                return $q->reset('field')->field('count(*)', $args['alias'] ?? null);
```
mvorisek, line 608: 👍

src/Persistence/Sql.php
```diff
@@ -631,14 +607,14 @@ public function action(Model $m, $type, $args = [])
...
+            case 'exists':
+                $this->initQueryConditions($m, $q);
```
mvorisek, line 610: refactor m to model and q to query

src/Persistence/Sql.php
```diff
@@ -364,60 +364,36 @@ protected function setLimitOrder(Model $m, Query $q)
...
-                if ($cond[0] instanceof Field && !in_array(strtolower($cond[1]), ['like', 'regexp'], true)) {
-                    $cond[2] = $this->typecastSaveField($cond[0], $cond[2]);
-                }
-                $q->where($cond[0], $cond[1], $cond[2]);
+        // nested conditions
+        if ($condition instanceof Model\Scope\CompoundCondition) {
+            $expression = $condition->isOr() ? $query->orExpr() : $query->andExpr();
+
+            foreach ($condition->getNestedConditions() as $nestedCondition) {
+                $expression = $this->initQueryConditions($model, $expression, $nestedCondition);
```
mvorisek: some usages of `initQueryConditions()` does not assign, I propose to change `initQueryConditions()` to return void

tests/PersistentArrayTest.php
```diff
@@ -404,31 +419,188 @@ public function testLike()
...
+        $m->addCondition('code', [11, 12]);
+        $result = $m->action('select')->get();
+        $this->assertSame(2, count($result));
+        $this->assertSame($a['countries'][1], $result[1]);
+        $this->assertSame($a['countries'][2], $result[2]);
+        unset($result);
+        $m->unload();
+
+        $m->scope()->clear();
+        $m->addCondition('code', 'NOT IN', [11, 12, 13, 14, 15, 16, 17]);
```
mvorisek, line 559: empty IN should be tested by DSQL, but add one test for it here - should return zero results, but not throw

needed because empty `IN()` is SQL emits an error normally

src/Persistence/Array_.php
```diff
@@ -146,17 +149,15 @@ public function tryLoadAny(Model $m, string $table = null): ?array
...
 
-        $data = $this->typecastSaveRow($m, $data);
+        $data = $this->typecastSaveRow($model, $data);
 
-        $id = $this->generateNewId($m, $table);
-        if ($m->id_field) {
-            $data[$m->id_field] = $id;
+        $id = $this->generateNewId($model, $table);
+        if ($model->id_field) {
+            $data[$model->id_field] = $id;
```
mvorisek, line 160: I know this is not directly related to Scopes, but if this is for AI, we should move it, and definitelly not assing to `$data`, `$data` must represent a map of table data by table name (and possibly, id_field can be a table name)

src/Persistence/Array_.php
```diff
@@ -267,128 +277,75 @@ public function export(Model $m, array $fields = null, $typecast_data = true): a
...
     {
-        $keys = $fields ? array_flip($fields) : null;
+        $data = $this->data[$model->table];
 
-        $data = array_map(function ($r) use ($m, $keys) {
-            // typecasting moved to export() method
-            //return $this->typecastLoadRow($m, $keys ? array_intersect_key($r, $keys) : $r);
-            return $keys ? array_intersect_key($r, $keys) : $r;
-        }, $this->data[$m->table]);
+        if ($keys = array_flip((array) $fields)) {
```
mvorisek, line 284: very, very weird condition to test if `$fields` is not null (or empty, but I belive that is a mistake - we should test simply `!== null`)

src/Persistence/Array_.php
```diff
@@ -397,27 +354,33 @@ public function action($m, $type, $args = [])
...
             case 'fx0':
+                if (!isset($args[0], $args[1])) {
+                    throw (new Exception('fx action needs 2 arguments, eg: ["sum", "amount"]'))
+                        ->addMoreInfo('action', $type);
+                }
 
-                return $action->aggregate($field->short_name, $fx);
-            */
+                $fx = $args[0];
+                $field = $args[1] ?? null;
```
mvorisek: never null, see condition above... I also prefer not to combine `isset`, harder to read

src/Model/Scope/AbstractCondition.php
mvorisek: > 
> 
> It is necessary for public at the moment to associate a condition object with a model.
> Otherwise we have to pass a model object as optional argument to `Condition::toWords` method and use it only if `Condition::getModel` is empty

I think it is cleaner to allow to pass optional Model to toWords. We should not need a clone then at all.

src/Model/Scope/AbstractCondition.php
```diff
@@ -0,0 +1,108 @@
...
+     * Return if scope has any conditions.
+     */
+    abstract public function isEmpty(): bool;
+
+    /**
+     * Convert the scope to human readable words when applied on $model.
+     *
+     * @return bool
+     */
+    abstract public function toWords(bool $asHtml = false): string;
```
mvorisek: I think `$asHtml` should be dropped

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+class BasicCondition extends AbstractCondition
+{
+    use ReadableCaptionTrait;
+
+    /**
+     * Stores the condition key.
+     *
+     * @var string|Field|Expression
+     */
+    public $key;
```
mvorisek, line 23: I would like a better name than `$key`

src/Model/Scope/BasicCondition.php
mvorisek: > 
> 
> `setModel` implies we are setting a model but actually we are not. Just performing actions when model is changed.
> Model is only set in the `Model\Scope` object (in `$model->scope` property)

so `if ($this->model !== $model) {` condition in `setModel()`?

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    public function onChangeModel(): void
+    {
+        if ($model = $this->getModel()) {
+            // if we have a definitive scalar value for a field
+            // sets it as default value for field and locks it
+            // new records will automatically get this value assigned for the field
+            // @todo: consider this when condition is part of OR scope
+            if ($this->operator === '=' && !is_object($this->value) && !is_array($this->value)) {
```
mvorisek, line 101: this is interesting function, however Scopes should not change a default value, at least there should be a switch in Model/Field/Persistence to enable/disable this.

I think this has the following issues now:
- his can be also a security issue, it may set some secret value we expect (even password/hash :D )
- what if another scope is attached later?
- system property may hide column from Table, etc. which may not be desired (even very confusing - if I filter with `=` on table, should it hide that column? No!)

This should be probably completely removed and implemented in UI Persistence only.

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+                    $field->default = $this->value;
+                }
+            }
+        }
+    }
+
+    public function toArray(): array
+    {
+        // make sure clones are used to avoid changes
+        $condition = clone $this;
```
mvorisek: lets clone the whole app and we are safe, ... I think we should clone where really needed...

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+            if ($field instanceof Field && $model->persistence && !in_array(strtoupper((string) $operator), self::$skipValueTypecast, true)) {
+                $value = $model->persistence->typecastSaveField($field, $value);
+            }
+
+            // only expression contained in $field
+            if (!$operator) {
+                return [$field];
+            }
+
+            // skip explicitly using '=' as in some cases it is transformed to 'in'
```
mvorisek, line 166: then it should be addressed in dsql ;-)

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+
+                if ($field instanceof Field) {
+                    $field->system = true;
+                    $field->default = $this->value;
+                }
+            }
+        }
+    }
+
+    public function toArray(): array
```
mvorisek: we should probably use more specific name - this does create an array for Dsql

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+                    } else {
+                        // otherwise add the condition to the referenced model
+                        // and check if any records exist matching the criteria
+                        $model->addCondition($field, $operator, $value);
+                        $operator = '=';
+                        $value = 1;
+                    }
+
+                    // if not counting we check for existence only
+                    $field = $field == '#' ? $model->action('count') : $model->action('exists');
```
mvorisek: `#` needs tests and optimization, this code part needs to be rechecked before merge

src/Model/Scope/BasicCondition.php
georgehristov: Yes, we clone it and call it the virtualApp.

Here the cloning is needed.

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+            return [];
+        }
+
+        if ($model = $condition->getModel()) {
+            if (is_string($field)) {
+                // shorthand for adding conditions on references
+                // use chained reference names separated by "/"
+                if (str_contains($field, '/')) {
+                    $references = explode('/', $field);
+
```
mvorisek: remove this space, hard to read

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,199 @@
...
+    public function onChangeModel(): void
+    {
+        foreach ($this->elements as $nestedCondition) {
+            $nestedCondition->onChangeModel();
+        }
+    }
+
+    public function isEmpty(): bool
+    {
+        return empty($this->elements);
```
mvorisek: count() === 0

src/Model/Scope/BasicCondition.php
georgehristov: This is an existing feature in ATK, I have not introduced it.
We can consider updating this functionality later.

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Clears the group from nested conditions.
+     *
+     * @return static
+     */
+    public function clear()
+    {
+        foreach ($this->elements as $nestedCondition) {
+            $nestedCondition->destroy();
```
mvorisek: destroy should be leaved to GC - when is it needed?

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        return $this;
+    }
+
+    public function simplify()
```
mvorisek: missing return type

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,199 @@
...
+    {
+        if (count($this->elements) != 1) {
```
mvorisek: !==

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,199 @@
...
+        return $this;
+    }
+
+    public function toWords(bool $asHtml = false): string
+    {
+        $parts = [];
+        foreach ($this->elements as $nestedCondition) {
+            $words = $nestedCondition->toWords($asHtml);
+
+            $parts[] = $this->isCompound() && $nestedCondition->isCompound() ? "({$words})" : $words;
```
mvorisek: '(' . $words . ')'

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /**
+     * Merge number of conditions using AND as junction.
+     *
+     * @param AbstractCondition $_
+     *
+     * @return static
+     */
+    public static function mergeAnd(AbstractCondition $conditionA, AbstractCondition $conditionB, $_ = null)
```
mvorisek: why not ...$conditions ?

src/Model/Scope/BasicCondition.php
georgehristov: Any suggestions?

src/Persistence/Sql.php
georgehristov: I am for it but let's refactor this after the merge. Otherwise the PR is contaminated.

src/Model/Scope/BasicCondition.php
mvorisek: because of refs? where should then move the clone inside `if (str_contains($field, '/')) {`

src/Persistence/Sql.php
georgehristov: done

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        $string = implode(' ', array_filter($words));
+
+        return $asHtml ? "<strong>{$string}</strong>" : $string;
+    }
+
+    protected function operatorToWords(bool $asHtml = false): string
+    {
+        return $this->operator ? (self::$dictionary[strtoupper((string) $this->operator)] ?? 'is equal to') : '';
```
mvorisek: this should fail (everywhere), operator if operator should be `=`, then is must be always set

src/Model.php
georgehristov: This is intentional as it solves the problem when model is cloned, etc

src/Model/Scope/BasicCondition.php
georgehristov: No, because operator may be `NULL` if the condition is based on `Expression` (raw SQL)

tests/ScopeTest.php
```diff
@@ -162,31 +193,64 @@ public function testContitionOnReferencedRecords()
...
+        
+        // countries with users that have no tickets
+        $country->addCondition('Users/Tickets/#', 0);
 
+        $this->assertEquals(2, $country->action('count')->getOne());
+        
         foreach ($country as $c) {
-            $this->assertEquals('BR', $c->get('code'));
+            $this->assertTrue(in_array($c->get('code'), ['FR', 'BR']));
         }
```
mvorisek, line 271: I am still missing one test which will reference one Model twice

src/Model/Scope/CompoundCondition.php
georgehristov: why?

src/Model/Scope/CompoundCondition.php
georgehristov: what is the difference? Except that it will fail if `$this->elements` is not countable?

tests/PersistentArrayTest.php
georgehristov: done

src/Model.php
georgehristov: This is intentional and for backward compatibility on setting nested OR conditions

src/Model/Scope/BasicCondition.php
georgehristov: Suggestions?

tests/ScopeTest.php
georgehristov: Give me an example, please

src/Persistence/Array_.php
georgehristov: done

src/Model/Scope/BasicCondition.php
georgehristov: done

docs/conditions.rst
georgehristov: Finished with fixing and introducing tests on multilevel referenced conditions

src/Model/Scope/AbstractCondition.php
georgehristov: Any other idea how to render the scope as string nicely so that field names are highlighted for instance?

src/Persistence/Sql.php
georgehristov: done

src/Model/Scope/BasicCondition.php
georgehristov: Tests included

src/Model/Scope/BasicCondition.php
mvorisek: I formulated it badly, there should be no `is equal to` fallback

src/Model/Scope/CompoundCondition.php
mvorisek: we should use single quoted strings only, easier to read what is string and what is variable

src/Model/Scope/CompoundCondition.php
mvorisek: yes, empty never be used when the input is an array

src/Model/Scope/BasicCondition.php
mvorisek: I do not have any significantly better. The issue is that this produces "optimized" array which can not be reconstructed back which is not what you would expect from "to array" code. "to array" should mean cast object to an array only.

But you can not name/refactor this now, we can keep it.

tests/ScopeTest.php
mvorisek:
```
// users with tickets that have more than two users per country)
// test if User model can be referenced multiple times
$user->addCondition('Tickets/user/Country/Users/#', '>', 1);
// test if generated query has no duplicate column names because of counting/# field
// if added multiple times
$user->addCondition('Tickets/user/Country/Users/#', '>', 1);
$user->addCondition('Tickets/user/Country/Users/#', '>=', 2);
$user->addCondition('Tickets/user/Country/Users/Country/Users/#', '>', 1);
```

Am I understand correctly that the # is transformed to `count(*)` of EACH `Tickets/user/Country/Users` record
and
`'Tickets/user/Country/Users/#', '>, 1` produces a predicate/bool for `$user`, ie. it does not result in JOIN, ie. the returned users will be still unique if any matching user have more than one ticket?

src/Model/Scope/CompoundCondition.php
mvorisek: now I see what you want, your signature was probably better if `AbstractCondition` type would be used also for the 3rd type

but why we need mergeAnd/Or at all?

for me, it implements nesting which is supported in Scopes by definition

tests/ScopeTest.php
georgehristov: I just included the test and it works.
No JOINs are used. It is simply a chain for exists -> exists -> exists with adding the relation of the tables.

Users with tickets whose user has country with more than 1 users.

src/Model/Scope/CompoundCondition.php
georgehristov: You can see in the method what it is a shortcut for.

src/Model/Scope/AbstractCondition.php
mvorisek: traverse over the Scope objects in ui and merge the result there

src/Model/Scope/BasicCondition.php
georgehristov: Renamed to toQueryArgumentsArray

src/Model/Scope/CompoundCondition.php
georgehristov: Done

src/Model/Scope/CompoundCondition.php
georgehristov: done

src/Model/Scope/BasicCondition.php
georgehristov: Better now?

src/Model/Scope/BasicCondition.php
```diff
@@ -265,7 +265,7 @@ protected function keyToWords(bool $asHtml = false): string
 
     protected function operatorToWords(bool $asHtml = false): string
     {
-        return $this->operator ? (self::$dictionary[strtoupper((string) $this->operator)] ?? 'is equal to') : '';
+        return $this->operator ? (self::$dictionary[strtoupper((string) $this->operator)] ?? $this->operator) : '';
```
mvorisek: why? - it should emit at least php notice, you already test if operator is not empty in the first condition

if operator can be other than supported/listed, this is correct, but do we support (and want to support it) that everywhere?

src/Model/Scope/AbstractCondition.php
georgehristov: Maybe define a renderer Object which takes ($field, $operator, $value) words and returns a string?

src/Model/Scope/AbstractCondition.php
georgehristov: Changed now, resolved?

src/Model/Scope/BasicCondition.php
georgehristov: Changed

src/Action/Iterator.php
```diff
@@ -33,71 +35,175 @@ public function __construct(array $data)
...
 
-        return $this;
+        return $match;
+    }
+
+    protected function evaluateIf($v1, $operator, $v2): bool
+    {
+        switch (strtoupper((string) $operator)) {
+            case '=':
+                $result = is_array($v2) ? $this->evaluateIf($v1, 'IN', $v2) : $v1 == $v2;
```
DarkSide666: No support for `$v2 === null` (strict) ?

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,326 @@
...
+
+        return $this;
+    }
+
+    public function negate()
+    {
+        if ($this->operator && isset(self::$opposites[$this->operator])) {
+            $this->operator = self::$opposites[$this->operator];
+        } else {
+            throw new Exception('Negation of condition is not supported for ' . ($this->operator ?: 'no') . ' operator');
```
DarkSide666: Dynamic content should be moved in `addMoreInfo` not inside expression message.
That way it will be easier to localize messages.

src/Model/Scope/BasicCondition.php
```diff
@@ -0,0 +1,326 @@
...
+                    foreach ($references as $link) {
+                        $refModels[] = $model = $model->refLink($link);
+                    }
+
+                    foreach (array_reverse($refModels) as $refModel) {
+                        if ($field === '#') {
+                            $field = $value ? $refModel->action('count') : $refModel->action('exists');
+                        } else {
+                            $refModel->addCondition($field, $operator, $value);
+                            $field = $refModel->action('exists');
```
DarkSide666, line 145: Still not sure about this line why there is `exists`... Didn't get what case this is.

src/Model/Scope/CompoundCondition.php
```diff
@@ -0,0 +1,194 @@
...
+                continue;
+            }
+
+            $this->add(clone $nestedCondition);
+        }
+    }
+
+    public function __clone()
+    {
+        foreach ($this->elements as $k => $nestedCondition) {
```
DarkSide666, line 65: Maybe it would be good to not use default `$elements` property, but `$conditions` as collection name instead.
That way we could also introduce other collections like `$placeholders` etc in future.
Similarly like we moved fields in model from $elements to $fields collection, because we needed more than one type of collection in there.

DarkSide666 (changes requested):

src/Model/Scope/BasicCondition.php
mvorisek: to be solved later, issue https://github.com/atk4/data/issues/662

src/Model/Scope/BasicCondition.php
georgehristov: This is the main case. Basically we are setting a chain of `exists` ...

src/Model/Scope/BasicCondition.php
georgehristov: done

src/Action/Iterator.php
georgehristov: Fixed and tests included

src/Model/Scope/CompoundCondition.php
georgehristov: The main element of Compound condition is the nested condition and we store it under elements.
I looked into changing it to `$conditions` property but this adds unnecessary complexity in my opinion.
The placeholders concept does not rely on this (I have it ready raw)
Let's merge this like it is now. We can later modify without issues as this is internal handling.
WDYT?

DarkSide666: if conditions is only collection in there, then it's fine

src/Model/Scope/AbstractCondition.php
mvorisek: yes - but not renderer object within Scope, but helper method that will return these for your renderer, should should be responsible to traverse and render like renderer wants

src/Model/Scope/CompoundCondition.php
mvorisek: resolved by rename to createAnd/Or

src/Model/Scope/BasicCondition.php
mvorisek: fixed

src/Model/Scope/CompoundCondition.php
```diff
@@ -178,14 +178,14 @@ public function toWords(Model $model = null): string
      */
     public static function createAnd(...$conditions)
     {
-        return new self($conditions, self::AND);
+        return new static($conditions, self::AND);
```
georgehristov, line 197: `self` is intentional as using `Model\Scope::createAnd` brings unexpected results

mvorisek: I will fix

src/Persistence/Array_.php
georgehristov: well we need flipped array, this just does all the work in one line

DarkSide666 (approved):

src/Persistence/Array_.php
georgehristov: Not sure what the point is. `$data` stores the complete data including rows

src/Model/Scope/BasicCondition.php
DarkSide666: $key is ok

PR #663, opened 2020-07-23
Rename BasicCondition to Condition

mvorisek:
What is difference between CompoundCondition and Scope? Currently, Scope is topmost "scope". Everything else is "condition", but one is basic, onother is compound, ...

To simplify the terminology and reduce the number of terms, I propose to:

~~a) keep CompoundCondition more like an implementation detail and call collection of conditions a `Scope`. If it is part of Scope, we can call it `$subscope` or with specific name like `$customerScope`.~~

b) rename BasicCondition to `Condition`. Then if we speak about condition, we speak about the basic/simple one. Like in a real world, condition is undivisible

### DIFF
--- a/docs/conditions.rst
+++ b/docs/conditions.rst
@@ -259,7 +259,7 @@ Model Scope
 -----------
 
 Using the Model::addCondition method is the basic way to limit the model scope of records. Under the hood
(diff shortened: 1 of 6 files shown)


georgehristov (approved):
👍

PR #664, opened 2020-07-23
Scope improve usability - addCondition() should not wrap

mvorisek:
makes even more sense with https://github.com/atk4/data/pull/663 as we use term "condition" for simple condition only

when you, at least I, add a condition, I want to add a simple one

### DIFF
--- a/src/Model/Scope/CompoundCondition.php
+++ b/src/Model/Scope/CompoundCondition.php
@@ -65,8 +65,10 @@ class CompoundCondition extends AbstractCondition
         foreach ($this->elements as $k => $nestedCondition) {
             $this->elements[$k] = clone $nestedCondition;
             $this->elements[$k]->owner = $this;
+            $this->elements[$k]->short_name = $nestedCondition->short_name;
         }
         $this->owner = null;
+        $this->short_name = null;
     }
 
     /**
--- a/src/Model/Scope/CompoundCondition.php
+++ b/src/Model/Scope/CompoundCondition.php
@@ -72,17 +72,17 @@ class CompoundCondition extends AbstractCondition
     }
 
     /**
-     * Add nested condition.
-     *
-     * @param mixed $field
-     * @param mixed $operator
-     * @param mixed $value
-     *
-     * @return static
+     * @return $this
      */
     public function addCondition($field, $operator = null, $value = null)
     {
-        $this->add(static::createAnd(func_get_args()));
+        if (!is_array($field) && !($field instanceof Scope\AbstractCondition)) {
+            $condition = new Condition(...func_get_args());
+        } else {
+            $condition = static::createAnd(func_get_args());
+        }
+
+        $this->add($condition);
 
         return $this;
     }
--- a/src/Model/Scope/CompoundCondition.php
+++ b/src/Model/Scope/CompoundCondition.php
@@ -76,10 +76,10 @@ class CompoundCondition extends AbstractCondition
      */
     public function addCondition($field, $operator = null, $value = null)
     {
-        if (func_get_args() === 1 && $field instanceof AbstractCondition) {
+        if (func_num_args() === 1 && $field instanceof AbstractCondition) {
             $condition = $field;
-        } elseif (func_get_args() === 1 && is_array($field)) {
-            $condition = new self($field);
+        } elseif (func_num_args() === 1 && is_array($field)) {
+            $condition = static::createAnd(func_get_args());
         } else {
             $condition = new Condition(...func_get_args());
         }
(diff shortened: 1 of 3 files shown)


src/Model/Scope/CompoundCondition.php
```diff
@@ -69,17 +69,13 @@ public function __clone()
...
-     * @param mixed $field
-     * @param mixed $operator
-     * @param mixed $value
-     *
      * @return static
      */
-    public function addCondition($field, $operator = null, $value = null)
+    public function addCondition($key, $operator = null, $value = null)
     {
-        $this->add(new self([func_get_args()]));
```
georgehristov: I would only change the statement to
`$this->add(static::createAnd(func_get_args()));`

Why would you want to remove the flexibility? It does create what you expect + more and consistent with `Model::addCondition`. And code is also simpler and all handling is in one place (`__constructor`).

We can even use same method in `Model::addCondition`
`$this->scope()->addCondition(...func_get_args());`

mvorisek: > Why would you want to remove the flexibility? It does create what you expect + more and consistent with `Model::addCondition`. And code is also simpler and all handling is in one place (`__constructor`).

> We can even use same method in `Model::addCondition`
> `$this->scope()->addCondition(...func_get_args());`

I like to think in consistent design.

There are two things:

a) I do not think it is good to wrap when we add a simple condition. The structure should really look like to what you add, if you add `name = "George"`, then I would never expect a Compound condition from it. And, this was also not the case before (in case of array structure). We can agree on this, right?

b) Allow to pass a conditions as an array, ie. `addCondition([[1. cond], [2. cond]])`, that is your concern, right?

If the input is an array, we can use CompoundCondition - but what is the usecase vs. not using createAnd/createOr which I think is easier to read, because:
- of explicit And/Or in text
- you can still add CompoundCondition object using this method
- we enforce the user to use object as soon as possible, ie. to not create an array first

georgehristov: I have several concerns:
- Model will use addCondition in one way (BC considerations) and CompoundCondition in another (inconsistent)
- handling of creation of BasicCondition will be distributed on several places instead of only one (in constructor). When you addCondition you expect a condition to be added to the scope. And it is. in what way is internal workings.
- we reduce the flexibility
- we spend way too much time on theoretical discussions. At some point benefits of collaborative creation are wiped out as we cannot go further and just go in circles

src/Model/Scope/CompoundCondition.php
```diff
@@ -41,22 +41,22 @@ public function __construct(array $nestedConditions = [], string $junction = sel
         $this->junction = $junction;
 
         foreach ($nestedConditions as $nestedCondition) {
-            $nestedCondition = is_string($nestedCondition) ? new Condition($nestedCondition) : $nestedCondition;
+            if ($nestedCondition instanceof Scope\AbstractCondition) {
```
georgehristov: Should be `$nestedCondition instanceof AbstractCondition`

src/Model/Scope/CompoundCondition.php
```diff
@@ -41,22 +41,22 @@ public function __construct(array $nestedConditions = [], string $junction = sel
         $this->junction = $junction;
 
         foreach ($nestedConditions as $nestedCondition) {
-            $nestedCondition = is_string($nestedCondition) ? new Condition($nestedCondition) : $nestedCondition;
+            if ($nestedCondition instanceof AbstractCondition) {
+                $condition = $nestedCondition;
+            } else {
+                if (!is_array($nestedCondition)) {
```
georgehristov, line 47: maybe use one-liner casting 
`$nestedCondition = (array) $nestedCondition;`

mvorisek: no! we assigning the same is a bad design, really - at least much easier to read - if something, we do something, otherwise we do nothing

georgehristov (approved):
👍

src/Model/Scope/CompoundCondition.php
```diff
@@ -65,21 +65,26 @@ public function __clone()
...
-     * @return static
+     * @return $this
      */
     public function addCondition($field, $operator = null, $value = null)
     {
-        $this->add(static::createAnd(func_get_args()));
+        if (func_num_args() === 1 && $field instanceof AbstractCondition) {
+            $condition = $field;
+        } elseif (func_num_args() === 1 && is_array($field)) {
+            $condition = static::createAnd(func_get_args());
```
mvorisek, line 82: this was wrong, fixed in https://github.com/atk4/data/pull/1125

PR #666, opened 2020-07-24
Unify Model::addCondition() and CompoundCondition::addCondition()

georgehristov:
Revert to using `self` in `CompoundCondition::createXx` methods due to unpredictable behavior otherwise.
Using `static::createXx` within scope object results in creating `Scope` object and not `CompoundCondition` as intended

### DIFF
--- a/src/Model/Scope/CompoundCondition.php
+++ b/src/Model/Scope/CompoundCondition.php
@@ -79,7 +79,7 @@ class CompoundCondition extends AbstractCondition
      */
     public function addCondition($field, $operator = null, $value = null)
     {
-        $this->add(new self([func_get_args()]));
+        $this->add(static::createAnd(func_get_args()));
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


src/Model/Scope/CompoundCondition.php
```diff
@@ -194,14 +194,16 @@ public function toWords(Model $model = null): string
...
     }
 
     /**
      * @return static
      */
     public static function createOr(...$conditions)
     {
-        return new static($conditions, self::OR);
+        // intentional use of self
+        return new self($conditions, self::OR);
```
mvorisek: what did you solve, what is "unpredictable behavior", example? `Scope` was fixed to be using `parent::`, ie. `self`, but imagine this - you extend this method with some other custom class and what would you expect from `createOr`? I `new static`.

src/Model/Scope.php
```diff
@@ -44,20 +44,4 @@ public function negate()
     {
         throw new Exception('Model Scope cannot be negated!');
     }
-
-    /**
-     * @return static
-     */
-    public static function createAnd(...$conditions)
-    {
-        return parent::createAnd(...$conditions);
```
mvorisek: I see, `(parent::class)` needs to be here.

mvorisek (approved):

PR #670, opened 2020-07-24
Drop legacy/hack support of OR junction in anything else than legacy Model::addCondition()

mvorisek:


src/Model.php
```diff
@@ -965,6 +965,13 @@ public function _unset(string $name)
      */
     public function addCondition($field, $operator = null, $value = null)
     {
+        // legacy OR support before Scope was introduced
+        if (func_num_args() === 1 && is_array($field) && count($field) === 1 && is_array(reset($field))) {
```
mvorisek, line 969: we should probably drop the support later, now fully BC

src/Model/Scope/CompoundCondition.php
```diff
@@ -47,13 +47,7 @@ public function __construct(array $nestedConditions = [], string $junction = sel
...
                     $nestedCondition = [$nestedCondition];
                 }
-
-                // array of OR nested conditions
-                if (count($nestedCondition) === 1 && is_array(reset($nestedCondition))) {
-                    $condition = new self(reset($nestedCondition), self::OR);
-                } else {
-                    $condition = new Condition(...$nestedCondition);
-                }
+                $condition = new Condition(...$nestedCondition);
```
georgehristov, line 50: I would use a on-liner here:
`$condition = new Condition(...(array) $nestedCondition);`

src/Model.php
georgehristov: 👍

src/Model/Scope/CompoundCondition.php
mvorisek: I prefer to keep this is easy to understand nesting levels

The non covered/buggy cases come mostly from bad coding style ;-)

And could be possibly removed in PHP 10 :)

georgehristov (approved):

src/Model/Scope/CompoundCondition.php
georgehristov: Please ...

mvorisek: thank you :), never

PR #671, opened 2020-07-24
Rename CompoundCondition class to Scope

georgehristov:


tests/ScopeTest.php
```diff
@@ -275,45 +275,45 @@ public function testScope()
     {
         $user = clone $this->user;
 
-        $condition1 = new Condition('name', 'John');
+        $condition1 = ['name', 'John'];
```
mvorisek, line 278: this is why I do not think we should support array input, but this is short test, ok now

mvorisek (approved):

PR #673, opened 2020-07-25
Use constants to define Condition operators

georgehristov:


tests/ConditionSqlTest.php
```diff
@@ -283,7 +283,7 @@ public function testArrayCondition()
 
         $m = new Model($this->db, 'user');
         $m->addField('name');
-        $m->addCondition('name', 'not', []); // this should not fail, always should be true
+        $m->addCondition('name', 'not in', []); // this should not fail, always should be true
```
mvorisek, line 286: previous `not, []` is should be invalid, right, why did it not throw?

be very carefull and test what is/wanted behaviour of null - negate seems like to be not a negate and demorgan law can not be absolutely used then

tests/ScopeTest.php
```diff
@@ -177,6 +178,42 @@ public function testContitionToWords()
         $this->assertEquals('Country that has reference Users where any referenced record exists', $country->scope()->toWords());
     }
 
+    public function testContitionUnsupportedToWords()
+    {
+        $condition = new Condition('name', 'abc');
```
mvorisek, line 183: we should validate at create time (the operator)

but what is wrong here - I would expect `name = "abc"` which is perfectly valid

tests/ConditionSqlTest.php
georgehristov: It did not throw as there was no check of operator validity when creating the condition. Now I have fixed that so tests failed here (and rightly so). Now test also fixed.

tests/ScopeTest.php
```diff
@@ -177,6 +178,42 @@ public function testContitionToWords()
...
+        $country->addCondition('name', '==', 'abc');
+    }
+
+    public function testContitionUnsupportedNegate()
+    {
+        $condition = new Condition(new Expression('false'));
+
+        $this->expectException(Exception::class);
+
+        $condition->negate();
```
mvorisek, line 205: Hmm - is there currently a way to negate expr?

tests/ScopeTest.php
georgehristov: As mentioned above operator is now validated at condition construct time.

The wrong here is we do not know which model this condition is intended for to convert it to words.

src/Model/Scope/Condition.php
```diff
@@ -34,41 +34,75 @@ class Condition extends AbstractScope
...
-        '>' => 'is greater than',
-        '>=' => 'is greater or equal to',
-        '<=' => 'is smaller or equal to',
-        'LIKE' => 'is like',
-        'NOT LIKE' => 'is not like',
-        'IN' => 'is one of',
-        'NOT IN' => 'is not one of',
-        'REGEXP' => 'is regular expression',
-        'NOT REGEXP' => 'is not regular expression',
+    public const OPERATOR_EQUALS = '=';
```
mvorisek, line 37: 👍

tests/ScopeTest.php
georgehristov: At the moment not so it throws exception. We can work on a property `$negate` to be set to `true` if there is no operator and then prefix the query with `NOT`

tests/ConditionSqlTest.php
mvorisek: perfect

tests/ScopeTest.php
mvorisek: Hmm, is it complicated to print to words without model? If I construct some scope tree and I have not attached it yet, I would expect toWords still to work - of course, without specific typecasting, but we should support at least DateTime (instanceof) and default string. As simple as possible.

tests/ScopeTest.php
mvorisek: I think is is quite major design limitation, final version of negate should be very carefull how we implement null handling (like does not in() match etc.)

what is now behaviour of `NULL IN(NULL)`?

tests/ScopeTest.php
georgehristov: You should pass the model object as parameter, then it works (without attaching condition to model)

tests/ScopeTest.php
georgehristov: It is not a design limitation but a matter of implementation. Design allows it:)
Presently you have to add the `NOT` in your Expression (which is basically raw sql).

tests/ScopeTest.php
mvorisek: image

what is usecase of this, why not simply getField - when model is not fully constructed yet - it there any usecase for it? I would at least expect some string like `? (full field path/key)`

so `keyToWords` method then does not require a model

in `valueToWords` we should print something like `? ("casted to string")`

georgehristov: Should be okay with getField only

georgehristov: You can theoretically print out the condition toWords without the model but it might not be the best looking string

tests/ScopeTest.php
mvorisek: yes, current implementation limitation

tests/ScopeTest.php
mvorisek: lets remove the `hasField()` now (must is not to fallback to null/supress an error) and see if this strict implementation is not too strict for the real use

georgehristov: done

src/Model/Scope/Condition.php
```diff
@@ -280,11 +280,13 @@ protected function keyToWords(Model $model): string
 
                 if ($field === '#') {
                     $words[] = $this->operator ? 'number of records' : 'any referenced record exists';
-                    $field = '';
+                    $field = null;
                 }
             }
 
-            $field = $model->hasField($field) ? $model->getField($field) : null;
+            if (isset($field)) {
```
mvorisek: !== null, always

src/Model/Scope/Condition.php
```diff
@@ -87,12 +121,20 @@ public function __construct($key, $operator = null, $value = null)
...
             $value = $operator;
-            $operator = '=';
+            $operator = self::OPERATOR_EQUALS;
         }
 
         $this->key = $key;
-        $this->operator = $operator;
         $this->value = $value;
+
+        if (isset($operator)) {
```
mvorisek: `$operator !== null` - check also other code, I see the same a few lines below

`isset()` should do the same, expect `isset($iInventedNonDeclaredVariable)` typos will be supressed - very, very dangerous!

src/Model/Scope/Condition.php
```diff
@@ -256,18 +300,7 @@ protected function keyToWords(Model $model): string
...
-
-        $operator = strtoupper((string) $this->operator);
-
-        if (!isset(self::$dictionary[$operator])) {
-            throw (new Exception('Operator is not supported'))
-                ->addMoreInfo('operator', $operator);
-        }
-
-        return self::$dictionary[$operator];
+        return $this->operator ? self::$operators[$this->operator]['label'] : '';
```
mvorisek, line 303: see how things simplify with a good design :)

mvorisek (approved):
then merge

src/Model/Scope/Condition.php
georgehristov: I see. Changed ...

Issue #662, opened 2020-07-23
Do not mutate Field::default/system property in scope condition


mvorisek:
see https://github.com/atk4/data/pull/660/files/e1e0ae7947f60c268245e51ca12eafa5a45810dd#diff-d11d2f7d35e15c2e79f43737267e1c00

and related code:
https://github.com/atk4/data/blob/2f89724cfd4fd8c923ffc1761a96ac218591e9ca/src/Model/Scope/Condition.php#L101-L115

DarkSide666:
It's how it was before - BC way of doing things.
But of course we should revise that and see if that's still good approach or need to move that somehow to UI only.
In case of nested conditions this start to look less usable.

mvorisek:
you are right, this is actually hide field (because of system) with nested conditons like (a > 5 or (a = 5)) (the code will trigger becauseof `=` - but in whole complex it is not only one possible value)

georgehristov:
It can be implemented to be set to the value only when the condition is definitive - all parent scopes up to root scope have AND junction.

We have to look into alternatives to set it as well.

mvorisek:
one possibility, but this can imply default arguments for API etc., which is not fully correct (too implicit)

I think we should drop it - what code relies on it?

georgehristov:
The issue is that this functionality is used on model references.
E.g if removed [this line](https://github.com/atk4/data/blob/120d748b4c25df454c428ac756ba1075ec43cfd7/tests/SmboTransferTest.php#L83) (among many others) fails.

```
$a = new Account($this->db);

$a->save(['name' => 'AIB']);
$a->ref('Payment')->save(['amount' => 10]);
```

mvorisek:
please post PR to discuss more concrete - save() works on either loaded row/ID or a new one, right?

so if this is only a ref issue, we should be addresses there - you mean the issue is that you will otherwise have to set payment_id on a new row?

georgehristov:
The issue comes from sub-scoping models like [here](https://github.com/atk4/data/blob/10e63ef5f54e0e98fff60fd296816d16943aa600/tests/Model/Smbo/Payment.php#L13). So new entries are expected to have the value assigned permanently.
The only way I come up is to introduce a separate method on `Model` to set a condition and default field value.
Other ideas?

mvorisek:
> sub-scoping model

This is quite legit argument - only in this case it can be changed to system

But `=` equal condition definitely does not imply that.

I think, we have only two options:
[...]

PR #675, opened 2020-07-25
Refactor references NS to use consistent and explicit variable and method names

georgehristov:
Related to solving #662

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -123,24 +123,12 @@ class Reference
      */
     public function getTheirModel($defaults = []): Model
     {
-        $ourModel = $this->getOurModel();
-
         // set table_alias
-        if (!isset($defaults['table_alias'])) {
-            if (!$this->table_alias) {
-                $this->table_alias = $this->link;
-                $this->table_alias = preg_replace('/_' . ($ourModel->id_field ?: 'id') . '/', '', $this->table_alias);
-                $this->table_alias = preg_replace('/([a-zA-Z])[a-zA-Z]*[^a-zA-Z]*/', '\1', $this->table_alias);
-                if (isset($ourModel->table_alias)) {
-                    $this->table_alias = $ourModel->table_alias . '_' . $this->table_alias;
-                }
-            }
-            $defaults['table_alias'] = $this->table_alias;
-        }
+        $defaults['table_alias'] = $defaults['table_alias'] ?? $this->getTableAlias();
 
         // if model is Closure, then call it and return model
         if (is_object($this->model) && $this->model instanceof \Closure) {
-            $closure = ($this->model)($ourModel, $this, $defaults);
+            $closure = ($this->model)($this->getOurModel(), $this, $defaults);
 
             return $this->addToPersistence($closure, $defaults);
         }
@@ -154,11 +142,12 @@ class Reference
 
         // last effort - try to add model
         if (is_array($this->model)) {
-            $theirModel = [$this->model[0]];
-            $md = $this->model;
-            unset($md[0]);
+            $modelDefaults = $this->model;
+            $theirModel = [$modelDefaults[0]];
 
-            $defaults = array_merge($md, $defaults);
+            unset($modelDefaults[0]);
+
+            $defaults = array_merge($modelDefaults, $defaults);
         } elseif (is_string($this->model)) {
             $theirModel = [$this->model];
         } else {
@@ -170,6 +159,22 @@ class Reference
         }
 
         return $this->addToPersistence($theirModel, $defaults);
+    }
+
+    protected function getTableAlias(): string
+    {
+        if (!$this->table_alias) {
+            $ourModel = $this->getOurModel();
+
+            $this->table_alias = $this->link;
+            $this->table_alias = preg_replace('/_' . ($ourModel->id_field ?: 'id') . '/', '', $this->table_alias);
+            $this->table_alias = preg_replace('/([a-zA-Z])[a-zA-Z]*[^a-zA-Z]*/', '\1', $this->table_alias);
+            if (isset($ourModel->table_alias)) {
+                $this->table_alias = $ourModel->table_alias . '_' . $this->table_alias;
+            }
+        }
+
+        return $this->table_alias;
     }
 
     /**
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -11,7 +11,7 @@ namespace atk4\data;
  *
  * It's possible to extend the basic reference with more meaningful references.
  *
- * @property Model $owner
+ * @property Model $owner definition of "our model"
  */
 class Reference
 {
@@ -41,7 +41,7 @@ class Reference
     public $link;
 
     /**
-     * Definition of the destination model, that can be either an object, a
+     * Definition of the destination (their) model, that can be either an object, a
      * callback or a string. This can be defined during initialization and
      * then used inside getModel() to fully populate and associate with
      * persistence.
@@ -100,22 +100,39 @@ class Reference
         return '#ref_' . $this->link;
     }
 
+    public function getOurModel(): Model
+    {
+        return $this->owner;
+    }
+
+    /**
+     * @deprecated use getTheirModel instead - will be removed in dec-2020
+     */
+    public function getModel($defaults = []): Model
+    {
+        'trigger_error'('Method Reference::getModel is deprecated. Use Model::getTheirModel instead', E_USER_DEPRECATED);
+
+        return $this->getTheirModel($defaults);
+    }
+
     /**
      * Returns destination model that is linked through this reference. Will apply
      * necessary conditions.
      *
      * @param array $defaults Properties
      */
-    public function getModel($defaults = []): Model
+    public function getTheirModel($defaults = []): Model
     {
+        $ourModel = $this->getOurModel();
+
         // set table_alias
         if (!isset($defaults['table_alias'])) {
             if (!$this->table_alias) {
                 $this->table_alias = $this->link;
-                $this->table_alias = preg_replace('/_' . ($this->owner->id_field ?: 'id') . '/', '', $this->table_alias);
+                $this->table_alias = preg_replace('/_' . ($ourModel->id_field ?: 'id') . '/', '', $this->table_alias);
                 $this->table_alias = preg_replace('/([a-zA-Z])[a-zA-Z]*[^a-zA-Z]*/', '\1', $this->table_alias);
-                if (isset($this->owner->table_alias)) {
-                    $this->table_alias = $this->owner->table_alias . '_' . $this->table_alias;
+                if (isset($ourModel->table_alias)) {
+                    $this->table_alias = $ourModel->table_alias . '_' . $this->table_alias;
                 }
             }
             $defaults['table_alias'] = $this->table_alias;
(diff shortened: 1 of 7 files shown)


src/Reference.php
```diff
@@ -11,7 +11,7 @@
  *
  * It's possible to extend the basic reference with more meaningful references.
  *
- * @property Model $owner
+ * @property Model $owner definition of "our model"
```
mvorisek, line 14: of course, remove that ;-)

mvorisek: simply (our) model

src/Reference.php
```diff
@@ -41,7 +41,7 @@ class Reference
     public $link;
 
     /**
-     * Definition of the destination model, that can be either an object, a
+     * Definition of the destination (their) model, that can be either an object, a
```
mvorisek, line 44: simply "their model"

src/Reference.php
```diff
@@ -100,59 +100,81 @@ public function getDesiredName(): string
...
 
+    public function getOurModel(): Model
+    {
+        return $this->owner;
+    }
+
+    /**
+     * @deprecated use getTheirModel instead - will be removed in dec-2020
+     */
+    public function getModel($defaults = []): Model
```
mvorisek, line 113: makes sense if this was returning `getTheirModel()`

src/Reference.php
```diff
@@ -100,59 +100,81 @@ public function getDesiredName(): string
...
-                $this->table_alias = $this->link;
-                $this->table_alias = preg_replace('/_' . ($this->owner->id_field ?: 'id') . '/', '', $this->table_alias);
-                $this->table_alias = preg_replace('/([a-zA-Z])[a-zA-Z]*[^a-zA-Z]*/', '\1', $this->table_alias);
-                if (isset($this->owner->table_alias)) {
-                    $this->table_alias = $this->owner->table_alias . '_' . $this->table_alias;
-                }
-            }
-            $defaults['table_alias'] = $this->table_alias;
-        }
+        $defaults['table_alias'] = $defaults['table_alias'] ?? $this->getTableAlias();
```
mvorisek: if === null, then assing with method

do not assign the same value to vars, this is really a bad practise hard to read

src/Reference.php
```diff
@@ -100,59 +100,81 @@ public function getDesiredName(): string
...
 
         // if model is Closure, then call it and return model
         if (is_object($this->model) && $this->model instanceof \Closure) {
-            $c = ($this->model)($this->owner, $this, $defaults);
+            $closure = ($this->model)($this->getOurModel(), $this, $defaults);
```
mvorisek: this does not return probably another closure, so wrong var name!

src/Reference.php
```diff
@@ -100,59 +100,81 @@ public function getDesiredName(): string
...
+            return $this->addToPersistence($closure, $defaults);
         }
 
         // if model is set, then return clone of this model
         if (is_object($this->model)) {
-            $c = clone $this->model;
+            $theirModel = clone $this->model;
 
-            return $this->addToPersistence($c, $defaults);
+            return $this->addToPersistence($theirModel, $defaults);
```
mvorisek, line 140: probably clone $this->getTheirModel() here

src/Reference.php
```diff
@@ -100,59 +100,81 @@ public function getDesiredName(): string
...
+            $defaults = array_merge($modelDefaults, $defaults);
         } elseif (is_string($this->model)) {
-            $model = [$this->model];
+            $theirModel = [$this->model];
         } else {
-            $model = $this->model;
+            $theirModel = $this->model;
+        }
+
+        if (!$theirModel instanceof Model) {
```
mvorisek: drop this if, always true

src/Reference.php
```diff
@@ -100,59 +100,81 @@ public function getDesiredName(): string
...
+            $theirModel = $this->factory($theirModel, $defaults);
         }
 
-        if (!$model instanceof Model) {
-            $model = $this->factory($model, $defaults);
+        return $this->addToPersistence($theirModel, $defaults);
+    }
+
+    protected function getTableAlias(): string
```
mvorisek, line 164: make it private and name it better - getTableAlias can not return different result than `table_alias`!

src/Reference.php
georgehristov: In new php versions this can be updated to:
$var ??= 'value';

src/Reference.php
georgehristov: Yes, it was

src/Reference.php
```diff
@@ -100,59 +100,81 @@ public function getDesiredName(): string
...
+    {
+        if (!$this->table_alias) {
```
mvorisek, line 164: seems like it should be moved to init (and then drop this function)

src/Reference.php
georgehristov: I already know what you are going to comment beforehand when writing the code :)

src/Reference/ContainsOne.php
```diff
@@ -82,24 +83,21 @@ public function init(): void
      */
     protected function getDefaultPersistence($model)
```
mvorisek, line 84: can we drop this?

src/Reference.php
georgehristov: I just introduced it. Makes code cleaner and it is called from other places as well

src/Reference/ContainsOne.php
```diff
@@ -82,24 +83,21 @@ public function init(): void
...
-        $row = $m->get($this->our_field) ?: [];
-        //$row = $m->persistence->typecastLoadRow($m, $row); // we need this typecasting because we set persistence data directly
-
-        $data = [$this->table_alias => $row ? [1 => $row] : []];
-        $p = new ArrayOfStrings($data);
+        $row = $ourModel->get($this->our_field) ?: [];
+        //$row = $ourModel->persistence->typecastLoadRow($ourModel, $row); // we need this typecasting because we set persistence data directly
 
-        return $p;
+        return new Persistence\ArrayOfStrings([$this->getTableAlias() => $row ? [1 => $row] : []]);
```
mvorisek: why not 20? Isn't this ID?

src/Reference/HasMany.php
```diff
@@ -53,11 +54,12 @@ protected function referenceOurValue(): Field
...
     {
-        return $this->getModel($defaults)
-            ->addCondition(
-                $this->their_field ?: ($this->owner->table . '_' . ($this->owner->id_field ?: 'id')),
-                $this->getOurValue()
-            );
+        $ourModel = $this->getOurModel();
+
+        return $this->getTheirModel($defaults)->addCondition(
+            $this->their_field ?: ($ourModel->table . '_' . ($ourModel->id_field ?: 'id')),
```
mvorisek: we should have more like two protected methods for build and parse (I think related with table alias regexes above)

src/Reference/HasOne.php
```diff
@@ -164,8 +164,10 @@ public function init(): void
             $this->our_field = $this->link;
         }
 
-        if (!$this->owner->hasField($this->our_field)) {
-            $this->owner->addField($this->our_field, [
+        $ourModel = $this->getOurModel();
+
+        if (!$ourModel->hasField($this->our_field)) {
```
mvorisek, line 169: what is this condition for?

src/Reference/HasOneSql.php
```diff
@@ -86,31 +89,31 @@ function (Model $m) use ($their_field) {
      * addFields(['from', 'to'], ['type' => 'date']);
      *
      * @param array $fields
-     * @param array $defaults
+     * @param array $groupDefaults
      *
      * @return $this
      */
-    public function addFields($fields = [], $defaults = [])
+    public function addFields($fields = [], $groupDefaults = [])
```
mvorisek: `fieldsDefaults`

src/Reference/HasOneSql.php
```diff
@@ -86,31 +89,31 @@ function (Model $m) use ($their_field) {
...
-                        ->addMoreInfo('field', $field)
-                        ->addMoreInfo('alias', $alias);
+                        ->addMoreInfo('fieldName', $fieldName)
+                        ->addMoreInfo('alias', $defaults);
                 }
-                $alias = $alias[0];
+                $defaults = $defaults[0];
             } else {
-                $d = $defaults;
+                $field = $groupDefaults;
```
mvorisek: to be removed, right? (and if cond above)

src/Reference.php
mvorisek: this does not have any side effect, I will prefer if now, but I can live with it

it actually have - if table alias is not defined, so with array access, you "should always" use if and if you want set a fallback with `??` if key is not guaranteed to exist (like here, I think), then to be precise now you should use `if (isset($arr[k]))` and then assing. But keep this like it is now.

src/Reference.php
mvorisek: no, you coded a getter method that modify the property when called

it should be once in init (and optinally you can define private method `initTableAlias(): void`)

src/Reference/HasOne.php
georgehristov: Only add field if not existing in model I guess

src/Reference/ContainsOne.php
georgehristov: It is used in sub-classes so I would not touch it

src/Reference/HasOne.php
mvorisek: i see "if not", probably right...

src/Reference/ContainsOne.php
mvorisek: I think it should be removed

georgehristov: You can work on it if you like

src/Reference/HasMany.php
georgehristov: Also here - feel free to implement your ideas in this PR

mvorisek: related with https://github.com/atk4/data/pull/675#discussion_r460437912 , this should be probably solved in this PR

src/Reference/ContainsOne.php
mvorisek: on Tuesday we start new icescream shop, not that much time now ;-)

I would like this to be solved - but it is not directly related to this PR, thus is has to be not solved now

src/Reference.php
georgehristov: Done

src/Reference/HasOneSql.php
georgehristov: Done

src/Reference/HasOneSql.php
georgehristov: Simplified

src/Reference/HasMany.php
georgehristov: This PR is about pure renaming. Let's merge this PR and you look into it when you have time and willingness for it.

georgehristov:
@mvorisek let's merge this now as pure var name refactoring.
We also still need to place this in the "correct" NS under Model\Reference but it is a matter of further PRs

DarkSide666 (approved):
Very good work on making code more readable. Especially their/our model. That was always a pain to not mix these two.

PR #676, opened 2020-07-26
Improve Model::withPersistence

georgehristov:


src/Model.php
```diff
@@ -1281,44 +1281,44 @@ public function newInstance(string $class = null, array $options = [])
...
+                $model->addField($fieldName, clone $field);
+            }
+        }
 
-        return $m;
+        $model->data = $this->data;
+        $model->dirty = $this->dirty;
+        $model->limit = $this->limit;
+        $model->order = $this->order;
+        $model->scope = (clone $this->scope)->setModel($model);
```
mvorisek, line 1319: Why can not we simply clone $this?

src/Model.php
```diff
@@ -1281,44 +1281,44 @@ public function newInstance(string $class = null, array $options = [])
...
-        $m->dirty = $this->dirty;
+        // include any fields defined inline
+        foreach ($this->fields as $fieldName => $field) {
+            if (!$model->hasField($fieldName)) {
+                $model->addField($fieldName, clone $field);
+            }
+        }
 
-        return $m;
+        $model->data = $this->data;
```
mvorisek, line 1310: I think this method duplicates some other code.

Why is copying data prop better than `->set($model->get())` (some mention in the title?)

The code before it is a duplicate of `asModel()`, right?

georgehristov: No. `asModel` and `newInstance` use the same persistence.

>Why is copying data prop better than ->set($model->get()) (some mention in the title?)

I will remove this from the code comments as data is copied from the source model already.
What you expect from this method is a model clone with different persistence.

mvorisek: Hmm, do your bests and lets merge this

mvorisek (approved):

PR #677, opened 2020-07-26
Add `Model\UnionModel` model

georgehristov:


mvorisek:
dropping `atk4/report` completely, right?

src/Model/Aggregate.php
```diff
@@ -0,0 +1,334 @@
...
+     *
+     * @var bool
+     */
+    public $read_only = true;
+
+    /** @var Model */
+    public $master_model;
+
+    /** @var string */
+    public $id_field;
```
mvorisek: why field like this is declared when extending Model?

src/Model/Union.php
```diff
@@ -0,0 +1,462 @@
...
+
+use atk4\data\Exception;
+use atk4\data\Field;
+use atk4\data\FieldSqlExpression;
+use atk4\data\Model;
+use atk4\dsql\Expression;
+use atk4\dsql\Query;
+
+/**
+ * Union model combines multiple nested models through a UNION in order to retrieve
```
mvorisek, line 15: usecase, what is best alternative without it? We should select table name (or something unique per model, like it's object hash) and make the result rw, right?

src/Model/Aggregate.php
```diff
@@ -0,0 +1,334 @@
...
+ * If this field exist in the original model it will be added and you'll get exception otherwise. Finally you are
+ * permitted to add expressions.
+ *
+ * The base model must not be Union model or another Aggregate model, however it's possible to use Aggregate model as nestedModel inside Union model.
+ * Union model implements identical grouping rule on its own.
+ *
+ * You can also pass seed (for example field type) when aggregating:
+ * $aggregate->groupBy(['first','last'], ['salary' => ['sum([])', 'type'=>'money']];
+ */
+class Aggregate extends Model
```
mvorisek, line 38: For me, everything that except Model is very suboptimal - how can be model extended then in an user app (ok, with trait, but still)?

src/Model/Aggregate.php
georgehristov: Added phpdoc for it.

georgehristov:
> dropping atk4/report completely, right?

In its current for - yes. I will replace it with one introducing an UI component "Report".

src/Model/Union.php
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    /**
+     * Export model.
+     *
+     * @param array|null $fields        Names of fields to export
+     * @param string     $key_field     Optional name of field which value we will use as array key
+     * @param bool       $typecast_data Should we typecast exported data
+     */
+    public function export($fields = null, $key_field = null, $typecast_data = true): array
```
mvorisek, line 309: why export is different than the one in Model?

src/Model/Union.php
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    /**
+     * Specify a single field or array of fields.
+     *
+     * @param string|array $group
+     *
+     * @return $this
+     */
+    public function groupBy($group, array $aggregate = [])
```
mvorisek, line 345: probably not, union is about union only

src/Model/Union.php
```diff
@@ -0,0 +1,464 @@
...
+     * Otherwise adds condition to all nested models.
+     *
+     * @param mixed $key
+     * @param mixed $operator
+     * @param mixed $value
+     * @param bool  $forceNested Should we add condition to all nested models?
+     *
+     * @return $this
+     */
+    public function addCondition($key, $operator = null, $value = null, $forceNested = false)
```
mvorisek: is union really a SQL UNION? If yes, then we need to apply conditons on all models at query time

getFields() on this Model should return fields on resulting Model

src/Model/Union.php
```diff
@@ -0,0 +1,464 @@
...
+            // subquery should not be wrapped in parenthesis, SQLite is especially picky about that
+            $query->allowToWrapInParenthesis = false;
+
+            $args[$cnt++] = $query;
+        }
+
+        // last element is table name itself
+        $args[$cnt] = $this->table;
+
+        return $this->persistence->dsql()->expr('(' . implode(' UNION ALL ', $expr) . ') {' . $cnt . '}', $args);
```
mvorisek: should be more like `(() union all ())`

src/Model/Union.php
```diff
@@ -0,0 +1,464 @@
...
+     * fields, e.g. 'balance'=>['sum', 'amount'].
+     *
+     * You can also use Expression instead of array.
+     *
+     * @var array
+     */
+    public $aggregate = [];
+
+    /** @var string Derived table alias */
+    public $table = 'derivedTable';
```
mvorisek: why that default?

tests/Model/Client.php
```diff
@@ -6,6 +6,8 @@
 
 class Client extends User
 {
+    public $table = 'client';
```
mvorisek: now needed for some reason?

tests/Model/Transaction.php
```diff
@@ -0,0 +1,28 @@
...
+    public $nestedPayment;
+
+    public $subtractInvoice;
+
+    public function init(): void
+    {
+        parent::init();
+
+        // first lets define nested models
+        $this->nestedInvoice = $this->addNestedModel(new Invoice(), $this->subtractInvoice ? ['amount' => '-[]'] : []);
```
mvorisek: I strongly belive that nesting/wrapping should be standard property of Model, ie. we should either allow `table` to be a string or child Model.

One way to solve this is to define named models with registry which will produce registration key/string which can be used as a table.

mvorisek (changes requested):

DarkSide666:
> dropping `atk4/report` completely, right?
Yes at least drop support of it. There is nothing useful in there except these two models.

DarkSide666:
> > dropping atk4/report completely, right?
> 
> In its current form - yes. I will replace it with one introducing an UI component "Report".

I think better would be new repository for UI widget because old projects still relies on legacy atk4/report code.

src/Model/Union.php
DarkSide666: Yes. There should not be grouping in union model anymore.
Grouping is group model and union should be only about unions.

If you need both, then you should be able to easily use both model classes in conjunction.

src/Model/Union.php
DarkSide666: There should be logic when you apply conditions on nested models and when on top (_unioned_) model.

Like,
* if field exists in nested model, then apply condition there
* if field is some expression or doesn't exist in nested model, then apply condition on top model itself
Maybe there was something more, but approximately that was idea.

src/Model/Union.php
DarkSide666: NO PARENTHESIS !
I can be mistaken which exactly, but one of SQL engines didn't support `(A) union (B)`, should be `A union B` without parenhesis.

src/Model/Union.php
DarkSide666: Table name alias is mandatory for such union model constructs.

Like
```
select *
from (
    select * from A
    union all
    select * from B
) derivedTable <-- this is mandatory and should be something quite unique, but not random (should be static to not mess with DB engine cache)
```

tests/Model/Client.php
DarkSide666: model table name is required i guess...

tests/Model/Transaction.php
DarkSide666: `$model->table = $another_model` would be nice to support. Also dsql already should support that.
But I guess that's not a functionality of this specific PR.

DarkSide666:
I will definitely would like to closely review this PR before merging.

tests/Model/Client.php
georgehristov: For the testing it is now necssary

docs/unions.rst
```diff
@@ -0,0 +1,120 @@
...
+
+   $unionPaymentInvoice = new \atk4\data\Model\Union();
+   
+   $nestedPayment = $unionPaymentInvoice->addNestedModel(new Invoice());
+   $nestedInvoice = $unionPaymentInvoice->addNestedModel(new Payment());
+
+Next, assuming that both models have common fields "name" and "amount", `$unionPaymentInvoice` fields can be set::
+
+   $unionPaymentInvoice->addField('name');
+   $unionPaymentInvoice->addFiled('amount', ['type'=>'money']);
```
DarkSide666, line 44: typo

docs/unions.rst
```diff
@@ -0,0 +1,120 @@
...
+(see tests/ModelUnionTest.php, tests/Client.php, tests/Payment.php and tests/Invoice.php files).
+
+Union Model Definition
+----------------------
+
+Normally a model is associated with a single table. Union model can have multiple nested models defined and it fetches 
+results from that. As a result, Union model will have no "id" field. Below is an example of inline definition of Union model.
+The Union model can be separated in a designated class and nested model added within the init() method body of the new class::
+
+   $unionPaymentInvoice = new \atk4\data\Model\Union();
```
DarkSide666: for simplicity you can call it `$unionDocument` maybe?

docs/unions.rst
```diff
@@ -0,0 +1,120 @@
...
+Then data can be queried::
+
+   $unionPaymentInvoice->export();
+
+Union Model Fields
+------------------
+
+Below is an example of 3 different ways to define fields for the Union model::
+
+   // Will link the "name" field will all the nested models.
```
DarkSide666, line 55: _with_ typo

docs/unions.rst
```diff
@@ -0,0 +1,120 @@
...
+In most cases optimization on the query and Union model is not necessary as it will be done automatically.
+
+Grouping Results
+----------------
+
+Union model has also a built-in grouping support::
+
+   $unionPaymentInvoice->groupBy('client_id', ['amount'=>'sum']);
+
+When specifying a grouping field and it is associated with nested models then grouping will be enabled on every nested model.
```
DarkSide666: Ah, right. I guess this was why we added grouping in union model class too back then ...

src/Model/Union.php
mvorisek: > If you need both, then you should be able to easily use both model classes in conjunction.

exactly, this is not possible when 2 classes are extended

src/Model/Union.php
mvorisek: makes sense

src/Model/Union.php
mvorisek: fine

src/Model/Union.php
mvorisek: to allow good Model nesting/wrapping, we need a native support for it in standard Model

tests/Model/Client.php
mvorisek: fine

tests/Model/Transaction.php
mvorisek: but what is the purpose of this PR then, if we should introduce the support in completely different way? ;-)

src-schema/PhpunitTestCase.php
```diff
@@ -31,6 +31,10 @@ protected function setUp(): void
     {
         parent::setUp();
 
+        if ($this->db !== null) {
```
mvorisek, line 34: make sure this is never merged, we want to create a new connection for every test case

and it does not solve the connections issue at all...

georgehristov: For sure.

It looks like for some reason on Postgre the connection is not dropped after `$this-db = null` in `tearDown`.

mvorisek: that is what I said in Discord - you introduced circular references somewhere and refcounting is short, ie. the only way to kick destructors is GC

src-schema/PhpunitTestCase.php
```diff
@@ -73,6 +73,8 @@ public function stopQuery(): void
 
     protected function tearDown(): void
     {
+        $this->db->connection->connection()->close();
```
mvorisek, line 76: make sure this is also never merged in

georgehristov: This explicit diconnection solved the issue.

There is no circular reference, simply up to now the number of tests was below the Postres limit.

mvorisek: I already explain on Discord, that this is no true, otherwise burn testing will fail.

mvorisek: try replacing this with `gc_collect_cycles()` by yourself :)

tests/ModelUnionTest.php
```diff
@@ -42,6 +42,15 @@ protected function setUp(): void
         $this->subtractInvoiceTransaction = $this->createSubtractInvoiceTransaction($this->db);
     }
 
+    protected function tearDown(): void
+    {
+        $this->client = null;
```
mvorisek, line 47: 👍

src/Model/Aggregate.php
```diff
@@ -34,6 +35,9 @@
  *
  * You can also pass seed (for example field type) when aggregating:
  * $aggregate->groupBy(['first','last'], ['salary' => ['sum([])', 'type'=>'money']];
+ * 
+ * @property $persistence Persistence\Sql
```
mvorisek, line 39: so for Model we can not assument SQL, but for these extending classes yes?

src/Model/Aggregate.php
```diff
@@ -0,0 +1,338 @@
...
+ *
+ * The base model must not be Union model or another Aggregate model, however it's possible to use Aggregate model as nestedModel inside Union model.
+ * Union model implements identical grouping rule on its own.
+ *
+ * You can also pass seed (for example field type) when aggregating:
+ * $aggregate->groupBy(['first','last'], ['salary' => ['sum([])', 'type'=>'money']];
+ *
+ * @property $persistence Persistence\Sql
+ *
+ * @method Persistence\Sql expr
```
mvorisek, line 41: do we really have Model::expr method?

georgehristov (commented):
.

src/Model/Aggregate.php
georgehristov: In Models under Persistence\Sql - yes

src/Model/Aggregate.php
```diff
@@ -34,6 +35,10 @@
  *
  * You can also pass seed (for example field type) when aggregating:
  * $aggregate->groupBy(['first','last'], ['salary' => ['sum([])', 'type'=>'money']];
+ *
+ * @property \atk4\data\Persistence\Sql $persistence
+ *
+ * @method Expression expr($expr, array $args = []) forwards to Persistence\Sql::expr using $this as model
```
mvorisek, line 41: please see a lot of issues with Model::expr, there are not visible as I supressed them temporary to keep CI green

I think we should:
- stop using dynamic methods, at least for the commonly used names
- in Model, implemented them like:
```
function expr(params) {
    if ($this->persistence->supportsExpr()) {
        return $this->persistence->expr($this, params);
    }

    throw new Exception('Persistence does not support this operation');
}
```

I belive this it the only way to make static analyser happy.

georgehristov: For sure we can thinkof something but not in this PR

georgehristov:
@mvorisek this PR is ready for review.
Native support of Union / Aggregate in Model is not a good idea as in any case we are introducing completely different model with different fields

mvorisek:
> 
> 
> @mvorisek this PR is ready for review.
> Native support of Union / Aggregate in Model is not a good idea as in any case we are introducing completely different model with different fields

That is correct. Instead of table (join/tables), we should allow model and thus maintain types etc. to model Model further.

This is quite important for me, as I think we should always model all tables 1:1 with basic Models and then model everything else based on these base models. And easily also aggregate/union them.

mvorisek:
no go as implemented now, my feedback for better/faster clarification https://discord.com/channels/657649422280425523/664534590160896006/793086660526735380

DarkSide666:
> This is quite important for me, as I think we should always model all tables 1:1 with basic Models and then model everything else based on these base models. And easily also aggregate/union them.

In atk4/data Model is not always supposed to be 1:1 related to specific DB table. That's the power of models. They can contain multiple joined DB tables etc. Also Union and aggregation can and probably should be included in single Model.
For example, union models are often used when generating some reports. In such case there are no strictly pre-defined model, but it can be tweaked on run-time before executing query.
I think old union implementation was quite good. There was some issues in case of applying conditions - apply on query after union or apply on each unioned sub-select first. Result always is the same, but it can very much change performance of query. So if at all possible, then conditions need to be first applied on sub-selects and only then do union. That way temporary datasets are smaller and faster.

georgehristov:
Presently I have integrated the original implementation in Atk4/Data with fixes on applying scope.
Let's target to merge this soon and then upgrade from there.

DarkSide666:
> Presently I have integrated the original implementation in Atk4/Data with fixes on applying scope.
> Let's target to merge this soon and then upgrade from there.

If that's the case then I guess it's fine. Will try to check sometime later. Quite busy year-end :(

mvorisek:
> In atk4/data Model is not always supposed to be 1:1 related to specific DB table.

That is not an advantage, that is only how things used to be :). There is zero advantage defining types in Models multiple times. We should define once - also for DB creation/migration - and then Model based on these typed DB models (now impossible, as Models can not be nested, and my words are to support this first)

> Let's target to merge this soon and then upgrade from there.

I never supports this idea when changes are big, as maintaining/changing 1800+ LOC is very time consuming.

georgehristov:
@mvorisek The models are nested using the present functionality, you do not have to define the aggregate model again. You use existing model, designate existing fields to be used and define aggregate fields.

mvorisek:
@georgehristov please substract https://github.com/atk4/data/pull/817 from this PR so it contains only the new `Model\Union` model feature

docs/unions.rst
mvorisek: TODO update docs before merge, "will be enabled on every nested model" impl/idea was wrong for non-linear aggregate functions and it is not longer present

docs/unions.md
```diff
@@ -0,0 +1,119 @@
...
+## Union Model Definition
+
+Normally a model is associated with a single table. Union model can have multiple nested models defined and it fetches
+results from that. As a result, Union model will have no "id" field. Below is an example of inline definition of Union model.
+The Union model can be separated in a designated class and nested model added within the init() method body of the new class:
+
+```
+$unionPaymentInvoice = new \Atk4\Data\Model\Union();
+
+$nestedPayment = $unionPaymentInvoice->addNestedModel(new Invoice());
```
DarkSide666, line 41: variable names do not match added models

docs/unions.md
```diff
@@ -0,0 +1,119 @@
...
+(see tests/ModelUnionTest.php, tests/Model/Client.php, tests/Model/Payment.php and tests/Model/Invoice.php files).
+
+## Union Model Definition
+
+Normally a model is associated with a single table. Union model can have multiple nested models defined and it fetches
+results from that. As a result, Union model will have no "id" field. Below is an example of inline definition of Union model.
+The Union model can be separated in a designated class and nested model added within the init() method body of the new class:
+
+```
+$unionPaymentInvoice = new \Atk4\Data\Model\Union();
```
DarkSide666, line 39: should be `\Atk4\Data\Model\UnionModel()`

docs/unions.md
```diff
@@ -0,0 +1,119 @@
...
+{ref}`Expressions` and {ref}`Joins` are working just as they would on any other model.
+
+## Field Mapping
+
+Sometimes the field that is defined in the UnionModel model may be named differently inside nested models.
+E.g. Invoice has field "description" and payment has field "note".
+When defining a nested model a field map array needs to be specified:
+
+```
+$nestedPayment = $unionPaymentInvoice->addNestedModel(new Invoice());
```
DarkSide666, line 84: mismatched variable names

docs/unions.md
```diff
@@ -0,0 +1,119 @@
...
+$nestedInvoice = $unionPaymentInvoice->addNestedModel(new Payment(), ['description' => '[note]']);
+$unionPaymentInvoice->addField('description');
+```
+
+The key of the field map array must match the UnionModel field. The value is an expression. (See {php:meth}`Model::addExpression`).
+This format can also be used to reverse sign on amounts. When we are creating "Transactions", then invoices would be
+subtracted from the amount, while payments will be added:
+
+```
+$nestedPayment = $mUnion->addNestedModel(new Invoice(), ['amount' => '-[amount]']);
```
DarkSide666, line 94: mismatched variable names

docs/unions.md
```diff
@@ -0,0 +1,119 @@
...
+$nestedInvoice = $mUnion->addNestedModel(new Payment(), ['description' => '[note]']);
+$unionPaymentInvoice->addField('description');
+```
+
+Should more flexibility be needed, more expressions (or fields) can be added directly to nested models:
+
+```
+$nestedPayment = $unionPaymentInvoice->addNestedModel(new Invoice(), ['amount' => '-[amount]']);
```
DarkSide666, line 102: mismatched variable names

docs/unions.md
```diff
@@ -0,0 +1,119 @@
...
+$nestedInvoice = $unionPaymentInvoice->addNestedModel(new Payment(), ['description' => '[note]']);
+
+$nestedPayment->addExpression('type', ['expr' => '\'payment\'']);
+$nestedInvoice->addExpression('type', ['expr' => '\'invoice\'']);
+$unionPaymentInvoice->addField('type');
+```
+
+A new field "type" has been added that will be defined as a static constant.
```
DarkSide666, line 110: What is "static constant" here? Minimal example would be good.

DarkSide666 (changes requested):
I didn't review codebase, but in documentation there are some errors to fix.

docs/unions.md
mvorisek: 👍

docs/unions.md
mvorisek: feel free to checkout the PR branch and push directly

mvorisek (changes requested):
prevent accidental merge

impl. is still probably problematic and tests should be cleaned up

src/Model/UnionModel.php
```diff
@@ -0,0 +1,228 @@
...
+        return $query;
+    }
+
+    /**
+     * @param list<Persistence\Sql\Query> $subqueries
+     */
+    private function createUnionQuery(array $subqueries): Persistence\Sql\Query
+    {
+        $unionQuery = $this->getPersistence()->dsql();
+        $unionQuery->mode = 'union_all';
```
mvorisek, line 152: subqueries must support order by, limit or double wrap then to workaround

tests/ModelUnionTest.php
```diff
@@ -0,0 +1,411 @@
...
+        $this->assertSameSql('-`amount`', $transaction->expr('[]', [$transaction->getFieldExpr($transaction->nestedInvoice, 'amount', '-[]')])->render()[0]);
+        $this->assertSameSql('-NULL', $transaction->expr('[]', [$transaction->getFieldExpr($transaction->nestedInvoice, 'blah', '-[]')])->render()[0]);
+    }
+
+    public function testCreateSubqueryBasic(): void
+    {
+        $transaction = $this->createTransaction();
+
+        $this->assertSameSql(
+            'select `name` `name` from `invoice` UNION ALL select `name` `name` from `payment`',
```
mvorisek, line 71:
```
 1) Atk4\Data\Tests\Persistence\Sql\MaterializedArrayActionTest::testRenderMultipleRows
Atk4\Data\Persistence\Sql\ExecuteException: An exception occurred while executing a query: ORA-01790: expression must have same datatype as corresponding expression
  query: 'select
      true "bool",
      -9223372036854775808 "int",
      -1.5E - 50 "float",
      '' "string"
    from
      "DUAL"
    union all
    select
      NULL,
      9223372036854775807,
      1.0123456789123E + 300,
      ' foo
    '
    from
      "DUAL"'
```

The `ORA-01790` Oracle error is caused when one select selects `null` from untyped table.

PR #679, opened 2020-07-27
Separate Model reference methods in a trait

georgehristov:
First cosmetic step in moving Reference NS to Model\Reference


mvorisek:
To be consistent with user actions, I think the trait is better to na named as `HasReferencesTrait`.

Is this trait to be used somewhere else, or purely to split one class into multiple files?

mvorisek (approved):

PR #680, opened 2020-07-28
Move Join to Model\Join

georgehristov:
BC break (minor due to internal use of the classes only):
- `instanceof \atk4\data\Join::class` should be replaced with `instanceof \atk4\data\Model\Join::class`
- `instanceof \atk4\data\Join\Array_::class` should be replaced with `instanceof \atk4\data\Persistence\Array_\Join::class`
- `instanceof \atk4\data\Join\Sql::class` should be replaced with `instanceof \atk4\data\Persistence\Sql\Join::class`


mvorisek:
Why Join needs to be Join instead of Model and what si the difference between Union Model?

georgehristov:
> Why Join needs to be Join instead of Model

Because of separation of Model and Persistence. Each persistence handles the JOIN in a child class

> what si the difference between Join and Union Model?

Same like in SQL:
https://stackoverflow.com/questions/905379/what-is-the-difference-between-join-and-union

mvorisek:
answered in https://github.com/atk4/data/issues/682

if this is rename only and @DarkSide666 thinks it should be moved lets move it, I can not comment on something that should be not presented at all ;-)

DarkSide666:
@georgehristov Is this just moving namespaces and moving functionality out from Model class to trait or is there something more to pay attention on?
If it's just moving NS, then it looks fine with me to move it under respective persistence.

georgehristov:
@DarkSide666 nothing more than reordering and separating into Trait

DarkSide666 (approved):

PR #683, opened 2020-07-29
Introduce traversing ref links and aggregates in Model::get

georgehristov:
- cleanup and align Model References further f8a56a7, 2c381345
- fix an issue on recursion on self-referencing models when fields added in init method e13afc8c7d, 2dd281a4
- call `HasOneSql::addField` from `HasOneSql::addTitle` and avoid duplicate code 2dd281a4
- introduce support for resolving traversing links in `Reference::getRef` 1d72c68568
- fix issue with improper loading/saving of referenced models 2d9d817f
- introduce link traversing and aggregates in `Model::get` 27d2b8b599
- update tests to include new functionality 9ab574ab2

Major bug fixes and new functionality introduced:
```
$user->load(1);
$user->get('Company/Orders/amount'); // 'amount' default value
$user->get('Company/Orders/sum(amount)')); // returns sum of all order amounts for the company of user 1
$user->get('Company/Orders/count(*)'); // returns number of orders for the company of user 1
$user->get('Company/count(*)'); // returns number of companies of user 1
```

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -539,7 +539,9 @@ class Field implements Expressionable
      */
     public function getCaption(): string
     {
-        return $this->caption ?? $this->ui['caption'] ?? $this->readableCaption($this->short_name);
+        $caption = $this->caption ?? $this->ui['caption'] ?? $this->readableCaption($this->short_name);
+
+        return $caption instanceof \Closure ? $caption($this) : $caption;
     }
 
     // }}}
--- a/src/Model.php
+++ b/src/Model.php
@@ -813,13 +813,11 @@ class Model implements \IteratorAggregate
             }
 
             // aggregate function
-            if (str_contains($fieldName, '(')) {
-                $args = [];
-                if (preg_match('/^([\w ]+)\(([\w \*]*)/', $fieldName, $args)) {
-                    array_shift($args);
+            $args = [];
+            if (preg_match('~^(\w+)\(([\w*]*)\)$~', $fieldName, $args)) {
+                array_shift($args);
 
-                    return $this->action('fx', $args)->getOne();
-                }
+                return $this->action('fx', $args)->getOne();
             }
         }
 
--- a/src/Model.php
+++ b/src/Model.php
@@ -798,6 +798,11 @@ class Model implements \IteratorAggregate
 
         // if field is not existing maybe it is a traversing link or aggregate function
         if (!$this->hasField($fieldName)) {
+            // alias for count aggregate, consistent with Model scopes
+            if ($fieldName === '#') {
+                $fieldName = 'count(*)';
+            }
+
             // traverses relationships and returns value from a referenced model, e.g contact/company/name
             if (str_contains($fieldName, '/')) {
                 $refs = explode('/', $fieldName);
--- a/src/Model.php
+++ b/src/Model.php
@@ -776,33 +776,59 @@ class Model implements \IteratorAggregate
      *
      * @return mixed
      */
-    public function get(string $field = null)
+    public function get(string $fieldName = null)
     {
-        if ($field === null) {
+        if ($fieldName === null) {
             // Collect list of eligible fields
             $data = [];
             if ($this->only_fields) {
                 // collect data for actual fields
-                foreach ($this->only_fields as $field) {
-                    $data[$field] = $this->get($field);
+                foreach ($this->only_fields as $fieldName) {
+                    $data[$fieldName] = $this->get($fieldName);
                 }
             } else {
                 // get all fields
-                foreach ($this->getFields() as $field => $f) {
-                    $data[$field] = $this->get($field);
+                foreach ($this->getFields() as $fieldName => $field) {
+                    $data[$fieldName] = $this->get($fieldName);
                 }
             }
 
             return $data;
         }
 
-        $this->checkOnlyFieldsField($field);
+        // if field is not existing maybe it is a traversing link or aggregate function
+        if (!$this->hasField($fieldName)) {
+            // traverses relationships and returns value from a referenced model, e.g contact/company/name
+            if (str_contains($fieldName, '/')) {
+                $refs = explode('/', $fieldName);
 
-        if (array_key_exists($field, $this->data)) {
-            return $this->data[$field];
+                $fieldName = array_pop($refs);
+
+                return $this->ref(implode('/', $refs))->get($fieldName);
+            }
+
+            // aggregate function
+            if (str_contains($fieldName, '(')) {
+                $args = [];
+                if (preg_match('/^([\w ]+)\(([\w \*]*)/', $fieldName, $args)) {
+                    array_shift($args);
+
+                    return $this->action('fx', $args)->getOne();
+                }
+            }
         }
 
-        return $this->getField($field)->default;
+        $this->checkOnlyFieldsField($fieldName);
+
+        if ($fieldName === $this->id_field) {
+            return $this->id;
+        }
+
+        if (array_key_exists($fieldName, $this->data)) {
+            return $this->data[$fieldName];
+        }
+
+        return $this->getField($fieldName)->default;
     }
 
     /**
--- a/src/Model/ReferencesTrait.php
+++ b/src/Model/ReferencesTrait.php
@@ -185,6 +185,15 @@ trait ReferencesTrait
      */
     public function getRef($link): Reference
     {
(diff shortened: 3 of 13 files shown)


src/Reference/HasOneSql.php
```diff
@@ -20,9 +20,10 @@ class HasOneSql extends HasOne
      *
      * Returns Expression in case you want to do something else with it.
      *
-     * @param string|Field|array $ourFieldName or [$field, ..defaults]
+     * @param string|Field|array $ourFieldName   or [$field, ..defaults]
+     * @param string|\Closure    $theirFieldName
```
mvorisek, line 24: when we need Closure? It should be probably resolved before passing to this method

src/Model.php
```diff
@@ -776,33 +776,64 @@ public function setMulti(array $fields)
...
+
+                $fieldName = array_pop($refs);
+
+                return $this->ref(implode('/', $refs))->get($fieldName);
+            }
+
+            // aggregate function
+            if (str_contains($fieldName, '(')) {
+                $args = [];
+                if (preg_match('/^([\w ]+)\(([\w \*]*)/', $fieldName, $args)) {
```
mvorisek: why parent if then?

src/Model.php
```diff
@@ -776,33 +776,64 @@ public function setMulti(array $fields)
...
-        $this->checkOnlyFieldsField($field);
+        // if field is not existing maybe it is a traversing link or aggregate function
+        if (!$this->hasField($fieldName)) {
+            // alias for count aggregate, consistent with Model scopes
+            if ($fieldName === '#') {
+                $fieldName = 'count(*)';
+            }
+
+            // traverses relationships and returns value from a referenced model, e.g contact/company/name
+            if (str_contains($fieldName, '/')) {
```
mvorisek, line 807: can be deduplicate this, it seems on a several places now

src/Model.php
```diff
@@ -776,33 +776,64 @@ public function setMulti(array $fields)
...
 
             return $data;
         }
 
-        $this->checkOnlyFieldsField($field);
+        // if field is not existing maybe it is a traversing link or aggregate function
+        if (!$this->hasField($fieldName)) {
+            // alias for count aggregate, consistent with Model scopes
+            if ($fieldName === '#') {
+                $fieldName = 'count(*)';
```
mvorisek, line 803: can we drop `#` then completely?

src/Reference/ContainsOne.php
```diff
@@ -79,25 +79,13 @@ public function init(): void
      *
      * @param Model $model Referenced model
      *
-     * @return Persistence|false
+     * @return Persistence
```
mvorisek: type can be moved to code

src/Model/ReferencesTrait.php
```diff
@@ -185,6 +185,15 @@ public function refLink($link, $defaults = []): self
      */
     public function getRef($link): Reference
     {
+        // support for relationship traversing link, e.g contact/company/customers
```
mvorisek, line 188: using string is quite antipattern - what is the typical usecase?

In the future, we should drop static strings in advance of refactorable concept - this is the same as declaring classes with string directly.

src/Field.php
```diff
@@ -539,7 +539,9 @@ public function isHidden(): bool
      */
     public function getCaption(): string
     {
-        return $this->caption ?? $this->ui['caption'] ?? $this->readableCaption($this->short_name);
+        $caption = $this->caption ?? $this->ui['caption'] ?? $this->readableCaption($this->short_name);
+
+        return $caption instanceof \Closure ? $caption($this) : $caption;
```
mvorisek, line 544: strong usecase for caption as Closure?

src/Reference/HasOneSql.php
georgehristov: We need the closure in this method in order to resolve `$theirModel` later.
The issue comes when a model is referencing its own class and there is usage of `addTitle` or `addField` in the init method which results in endless loop

src/Model.php
georgehristov: Looks cleaner

src/Model.php
georgehristov: `Reference::isTraversingLink`?

src/Model.php
georgehristov: We can do anything we want 😄 
But it is good to have one special short symbol for frequent use:
`$country->action('count')->getOne();`
can be written now
`$country->get('#');`

src/Model/ReferencesTrait.php
georgehristov: It is a shorthand for
`$user->ref('contact')->ref('company')->ref('customers')`
You can still use it

src/Field.php
georgehristov: Same like this [comment](https://github.com/atk4/data/pull/683#discussion_r462558176)

src/Model.php
mvorisek: no, do not test is possibly matching and then a better match :)

also, use regex more like `~^(\w+)\(([\w*]*)\)$~` (or `~^(\w+)\((.*)\)$~`, I do not know what should be allowed)

src/Reference/HasOneSql.php
mvorisek: this needs to be solved definitelly different

src/Model.php
mvorisek: I mean the whole code, not only str_contains...

src/Model.php
mvorisek: you have now introduced `$country->get('count(*)');`, right?

georgehristov: Yep

src/Reference/HasOneSql.php
georgehristov: Ideas and code are welcome 😃

src/Model.php
mvorisek: I looked now more into the code - `get(something else than field name)` is brutal antipattern - it seems super handy, but in reality it will hit db again and again - I am quite sure to not add this - `Model::get()` should return only loaded data, it is up too you to define them first.

the same probably even with getRef, like when you will use `contact/x` and `contact/y`, the `contact` ref will be created twice, right? (we want the user to be aware of this and help him with the lib design to assign `contact` ref to variable first)

src/Reference/HasOneSql.php
mvorisek: the same issue is very common in Model init with circulars refs

this can be addressed in many ways (traverse owners and get what you want if already declared, stack trace, model registry, ...)

but definitelly not criple model design :)

src/Model/ReferencesTrait.php
mvorisek: This is much better and easier to understand (you learn 10 elementary methods and you understand them, instead of shorthand, shorthand, ... Also because https://github.com/atk4/data/pull/683#discussion_r462571533

src/Reference/HasOneSql.php
mvorisek: (In tests) it is solely related with addTitle/withTitle which is called from - from Model init :)

this is common issue not only related with title...

- Wdyt about limitting Model::init to only 5 loops? But what it should return when the limit is reached?
- Or detect single loop in addTitle/withTitle? (should be possible to break it using not setting Persistence, thus not calling init)
- We can detect that using stack trace...
- We can easily return somewhere in the loop model/ref from previous loop, but I think that is not possible, as owner can be then not set correctly, right?
- or very easily do not use addTitle/withTitle in model init like you can not use cyclic ref for other columns, but there should be a better solution...

src/Model/ReferencesTrait.php
georgehristov: Yes, nobody is preventing you from doing so

src/Field.php
DarkSide666: I don't see anything bad about this except that you also have to add Closure in $caption property typehint.
But also I don't see good use-case for it. Closures are useful for dynamically changeable properties, but caption doesn't look like _dynamically changing_ one.

mvorisek (changes requested):
nicely documented, let's cherry pick the fixes / cleanup in individual PRs

once done/merged, recreate this PR with the new/modified functionality

same with #690

mvorisek:
I belive with hintable magic this has zero advantage

PR #684, opened 2020-07-29
Introduce aggregates in ref traversing scope conditions 

georgehristov:
```
// countries of a user that has tickets price summing up more than 50
$country->addCondition('Users/Tickets/sum(price)', '>', 50);
```


tests/ScopeTest.php
```diff
@@ -286,6 +287,17 @@ public function testConditionOnReferencedRecords()
             $this->assertTrue(in_array($c->get('code'), ['FR'], true));
         }
 
+        $country = clone $this->country;
+
+        // countries of a user that has tickets price summing up more than 90
```
mvorisek: of?

and 50 :)

georgehristov: fixed

src/Model/Scope/Condition.php
```diff
@@ -195,8 +195,14 @@ public function toQueryArguments(): array
...
                     foreach (array_reverse($refModels) as $refModel) {
+                        // # is alias for count(*)
                         if ($field === '#') {
-                            $field = $value ? $refModel->action('count') : $refModel->action('exists');
+                            $field = 'count(*)';
+                        }
+
+                        // support for conditions on aggregates
+                        if (preg_match('~^(\w+)\(([\w *]*)$~', $field, $matches)) {
+                            $field = $value ? $refModel->action('fx', array_slice($matches, 1)) : $refModel->action('exists');
```
mvorisek: simplified, no need to test for `(` - this will/should result in invalid column in standard addCondition

also, I do not think fallback to `exists` is correct with fx now (like `SUM(all rows with 0 value)` will not work)

add test for this

otherwise, I am ok with this feature in conditions as it can save a lot of work now

src/Model/Scope/Condition.php
```diff
@@ -201,8 +201,24 @@ public function toQueryArguments(): array
                         }
 
                         // support for conditions on aggregates
-                        if (preg_match('~^(\w+)\(([\w *]*)$~', $field, $matches)) {
-                            $field = $value ? $refModel->action('fx', array_slice($matches, 1)) : $refModel->action('exists');
+                        if (is_string($field) && str_contains($field, '(')) {
```
mvorisek: fix force push!

src/Persistence/Sql.php
```diff
@@ -632,7 +632,13 @@ public function action(Model $model, $type, $args = [])
 
                 [$fx, $field] = $args;
 
-                $field = is_string($field) ? $model->getField($field) : $field;
+                if (is_string($field)) {
```
mvorisek: usecase? I do not think we want that... we can easily allow sql injection

src/Model/Scope/Condition.php
georgehristov: So we either converto count here or we use the expression conversion you commented below

mvorisek: definitely here, we want even want to remove this completely very soon :)

here I was/am ok with this - as preparsed with regex

src/Model/Scope/Condition.php
georgehristov: Now all is adjusted

src/Model/Scope/Condition.php
georgehristov: Now this adjusted

src/Persistence/Sql.php
georgehristov: removed

src/Model/Scope/Condition.php
mvorisek: https://github.com/atk4/data/pull/684/commits/41f3cb186fbe06e2938ed64ba6f661477bcc6818 is still overwrited - badly staged

georgehristov: Your regex does not work + we need to check if $field is a string

mvorisek: probably missing `\)` before `$`

send examples of bad inputs then

georgehristov: The tests are failing here because sum(price) is not detected as matching the regex

https://github.com/atk4/data/blob/5dee73a7436ebccf7ea329efd680d5b8efca7c7e/tests/ScopeTest.php#L288-L291

mvorisek: yes - https://3v4l.org/4b0Yh - missing `\)`

mvorisek: `is_string($field)` tested twice - remove here

src/Model/Scope/Condition.php
```diff
@@ -195,8 +195,22 @@ public function toQueryArguments(): array
...
+
+                            continue;
+                        }
+
+                        $matches = [];
+                        // support for conditions on aggregates
+                        if (is_string($field) && preg_match('~^(\w+)\(([\w *]*)\)$~', $field, $matches)) {
+                            $field = $refModel->action('fx', array_slice($matches, 1));
+
+                            if (!$value) {
```
mvorisek: isn't missing value handled by dsql?

is this 1:1 sql equivalent? What about -1? that is why I think code like this everywhere is really bad...

src/Model/Scope/Condition.php
georgehristov: Check of `is_string($field)` is necessary as we modify `$field` within the loop

mvorisek: > 
> 
> Check of `is_string($field)` is necessary as we modify `$field` within the loop

where? previously you was using str_contains so I do not think so...

georgehristov: Lines 200, 208 or 216 change `$field` within the loop

https://github.com/atk4/data/blob/ec694af90dbd70fa0a977e5f8c33346f34a00bda/src/Model/Scope/Condition.php#L197-L220

mvorisek: ok - 216 - 200 and 208 never

src/Model/Scope/Condition.php
georgehristov: Well it is handled by sql...
condition is `... and (select sum("price") from "ticket" ...)`

Will remove the fallback in php

mvorisek: :) - your test should test that cond is not applied for summing existing rows with zero value

georgehristov: All updated.
I think Space X can now safely use atk4/data for crew dragon - out of the box 😆

src/Model/Scope/Condition.php
```diff
@@ -195,8 +195,17 @@ public function toQueryArguments(): array
...
-                        if ($field === '#') {
-                            $field = $value ? $refModel->action('count') : $refModel->action('exists');
+                        // # is alias for count(*)
+                        if ($field === '#' || $field === 'count(*)') {
+                            $field = $refModel->action($value ? 'count' : 'exists');
+
+                            continue;
+                        }
+
+                        $matches = [];
```
mvorisek: not needed

src/Model/Scope/Condition.php
georgehristov: preg_match expect subject to be string. The check is necessary. Marking this as resolved

src/Model/Scope/Condition.php
georgehristov: Removed

src/Model/Scope/Condition.php
```diff
@@ -195,8 +195,16 @@ public function toQueryArguments(): array
                     }
 
                     foreach (array_reverse($refModels) as $refModel) {
-                        if ($field === '#') {
-                            $field = $value ? $refModel->action('count') : $refModel->action('exists');
+                        // # is alias for count(*)
+                        if ($field === '#' || $field === 'count(*)') {
+                            $field = $refModel->action($value ? 'count' : 'exists');
```
mvorisek, line 200: If operator is `<`?

mvorisek: then we should probably optimize common exprs can we evaluated to exists

better (much better for performance) solution would be if we can generated queries like:
```
select count(*) as c from (
    select <standard query> limit X (<our value> + 1)
) t having c <our operator> <cond value>
```

mvorisek (changes requested):
we want to allow unrefactorable strings as much as possible and this PR requires them

please post a code of the best alrernative using `->ref()` currently to really see if the benefit/saved extra code justifies the negatives

mvorisek:
I belive with hintable magic this has zero advantage

PR #685, opened 2020-07-30
Introduce getQueryArguments in Field

georgehristov:
Let Field object handle conversion to query arguments.
Default functionality remains of Field typecasting the condition value using persistence.

The method can be extended in child field classes to handle different logic, e.g
a new class extending Field for saving multiple values can be set up to save them in persistence like an encoded string `__value1__value2__`.
Then a `new Condition('name', 'value1')` can be converted to query `'name' LIKE '\_\_value1\_\_'` by the Field object enabling conditions on single values

TODO: the question remains what to do with default behavior if the value is an array:
- typecast each element (in case operator is '=')?
- how to handle in case of other operators?

Functionality is important for enhancing atk4/login to handle user multiple roles!

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -7,6 +7,7 @@ namespace atk4\data;
 use atk4\core\DiContainerTrait;
 use atk4\core\ReadableCaptionTrait;
 use atk4\core\TrackableTrait;
+use atk4\data\Model\Scope;
 use atk4\dsql\Expression;
 use atk4\dsql\Expressionable;
 
@@ -513,19 +514,19 @@ class Field implements Expressionable
     /**
      * Returns arguments to be used for query on this field based on the condition.
      *
-     * @param string|null $operator one of Model\Scope\Condition operators
+     * @param string|null $operator one of Scope\Condition operators
      * @param mixed       $value    the condition value to be handled
      *
      * @todo: value is array
      */
-    public function getQueryArguments(Model\Scope\Condition $condition, $operator, $value): array
+    public function getQueryArguments($operator, $value): array
     {
         if ($persistence = $this->owner->persistence) {
             $skipValueTypecast = [
-                Model\Scope\Condition::OPERATOR_LIKE,
-                Model\Scope\Condition::OPERATOR_NOT_LIKE,
-                Model\Scope\Condition::OPERATOR_REGEXP,
-                Model\Scope\Condition::OPERATOR_NOT_REGEXP,
+                Scope\Condition::OPERATOR_LIKE,
+                Scope\Condition::OPERATOR_NOT_LIKE,
+                Scope\Condition::OPERATOR_REGEXP,
+                Scope\Condition::OPERATOR_NOT_REGEXP,
             ];
 
             if (!in_array($operator, $skipValueTypecast, true)) {
--- a/src/Field.php
+++ b/src/Field.php
@@ -521,17 +521,15 @@ class Field implements Expressionable
      */
     public function getQueryArguments($operator, $value): array
     {
-        if ($persistence = $this->owner->persistence) {
-            $skipValueTypecast = [
-                Scope\Condition::OPERATOR_LIKE,
-                Scope\Condition::OPERATOR_NOT_LIKE,
-                Scope\Condition::OPERATOR_REGEXP,
-                Scope\Condition::OPERATOR_NOT_REGEXP,
-            ];
+        $skipValueTypecast = [
+            Scope\Condition::OPERATOR_LIKE,
+            Scope\Condition::OPERATOR_NOT_LIKE,
+            Scope\Condition::OPERATOR_REGEXP,
+            Scope\Condition::OPERATOR_NOT_REGEXP,
+        ];
 
-            if (!in_array($operator, $skipValueTypecast, true)) {
-                $value = $persistence->typecastSaveField($this, $value);
-            }
+        if (!in_array($operator, $skipValueTypecast, true)) {
+            $value = $this->owner->persistence->typecastSaveField($this, $value);
         }
 
         return [$this, $operator, $value];
--- a/src/Field.php
+++ b/src/Field.php
@@ -516,8 +516,6 @@ class Field implements Expressionable
      *
      * @param string|null $operator one of Scope\Condition operators
      * @param mixed       $value    the condition value to be handled
-     *
-     * @todo: value is array
      */
     public function getQueryArguments($operator, $value): array
     {
@@ -529,7 +527,13 @@ class Field implements Expressionable
         ];
 
         if (!in_array($operator, $skipValueTypecast, true)) {
-            $value = $this->owner->persistence->typecastSaveField($this, $value);
+            if (is_array($value)) {
+                foreach ($value as &$option) {
+                    $option = $this->owner->persistence->typecastSaveField($this, $option);
+                }
+            } else {
+                $value = $this->owner->persistence->typecastSaveField($this, $value);
+            }
         }
 
         return [$this, $operator, $value];
--- a/src/Field.php
+++ b/src/Field.php
@@ -528,9 +528,9 @@ class Field implements Expressionable
 
         if (!in_array($operator, $skipValueTypecast, true)) {
             if (is_array($value)) {
-                foreach ($value as &$option) {
-                    $option = $this->owner->persistence->typecastSaveField($this, $option);
-                }
+                $value = array_map(function ($option) {
+                    return $this->owner->persistence->typecastSaveField($this, $option);
+                }, $value);
             } else {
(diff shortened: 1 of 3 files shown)


src/Field.php
```diff
@@ -508,6 +508,33 @@ public function useAlias(): bool
...
     // }}}
 
+    // {{{ Scope condition
+
+    /**
+     * Returns arguments to be used for query on this field based on the condition.
+     *
+     * @todo: value is array
+     */
+    public function getQueryArguments(Model\Scope\Condition $condition, $operator, $value): array
```
mvorisek: - import Scope
- only `Model\Scope\Condition $condition`, right? (it does contain operator and val)
- wdyt about moving quering from `Field` to `Model\QueryHelper` and every Model will have it's own instance - I belive, Field should stay passive/"descriptive only" as much as possible, Model the same
- and Scope classes the same too, so good move 👍

georgehristov: - Field should move under Model NS (up for discussion) so this will sort out naturally
- No, because the operator passed here has been modified already by the Condition object. 
To avoid confusion we can skip passing `$condition` altogether but it is good to have it I believe.
- will think about it
- 👍

mvorisek: >     * Field should move under Model NS (up for discussion) so this will sort out naturally

then we will remove use, very simple ;-)

>     * No, because the operator passed here has been modified already by the Condition object.
>       To avoid confusion we can skip passing `$condition` altogether but it is good to have it I believe.

Hmm, I think we should create new Condition object on calling side and pass it to here

georgehristov: `$condition` as an argument is not necessary and makes confusion so removed

src/Field.php
```diff
@@ -508,6 +509,36 @@ public function useAlias(): bool
...
 
+    // {{{ Scope condition
+
+    /**
+     * Returns arguments to be used for query on this field based on the condition.
+     *
+     * @param string|null $operator one of Scope\Condition operators
+     * @param mixed       $value    the condition value to be handled
+     *
+     * @todo: value is array
```
mvorisek: `$operator, $value` prototype makes sense as we are on field already...

why do we need to support an array?

src/Field.php
```diff
@@ -508,6 +509,36 @@ public function useAlias(): bool
...
+     */
+    public function getQueryArguments($operator, $value): array
+    {
+        if ($persistence = $this->owner->persistence) {
```
mvorisek: if persistence is not set, we should probably throw, is there anything against?

using `$this->owner->checkPersistence()`

mvorisek (approved):
solve the feedback as best as you can, then 🚀

src/Field.php
georgehristov: Is it mandatory to have the persistence?
In case of no persistence we simply do not cast the value and return 'as is'

mvorisek: when we prepare something to query, I do not see a reason why to prepare/handle usecase when we will need the persistence for quering a few ops after

maybe we do not even need to check here, just expect ;-)

src/Field.php
georgehristov: We do support array but we pass it onto dsql to handle it converting to `value in (a, b, c)`
The problem is if we have an array as value in Condition we need to typecast each value for correct results in dsql query
This is clearly the solution when we have '=' as operator.
But what to do if we have `new Condition($field, '>', [a, b, c])`?

Maybe validate value to expect array only when operator is =?

src/Field.php
georgehristov: Yes, I would say the best is to expect `$this->model->persistence` is here and directly use it.
`checkPersistence` is more for checking if a specific method exists in the persistence which we do not need here

src/Field.php
mvorisek: `> arr` may seems like it will expand to multiple ANDs, but I think we should NOT support it

so like you have `$skipValueTypecast` we should have `$valueCanBeArray` with `=` and `!=` operators and typecast individually. I belive, that we should still support only single level array and throw if more than that or operator does not support array.

src/Field.php
mvorisek: > Yes, I would say the best is to expect `$this->model->persistence` is here and directly use it.

yes, I think the same

src/Field.php
georgehristov: We are thinking in same direction - validate operator-value combination when constructing the condition
- accept array as value only when = and != operators
- accept only single level arrays

mvorisek: > validate operator-value combination when constructing the condition

1 + 2x 👍

georgehristov: Done. Let me know if some comments so i can then merge

src/Field.php
```diff
@@ -516,21 +516,23 @@ public function useAlias(): bool
...
+        $skipValueTypecast = [
+            Scope\Condition::OPERATOR_LIKE,
+            Scope\Condition::OPERATOR_NOT_LIKE,
+            Scope\Condition::OPERATOR_REGEXP,
+            Scope\Condition::OPERATOR_NOT_REGEXP,
+        ];
+
+        if (!in_array($operator, $skipValueTypecast, true)) {
+            if (is_array($value)) {
+                foreach ($value as &$option) {
```
mvorisek: use array map instead of ref

tests/ScopeTest.php
```diff
@@ -441,4 +441,18 @@ public function testDestroyEmpty()
 
         $this->assertEmpty($scope->toWords($user));
     }
+    
+    public function testInvalid1()
+    {
+        $this->expectException(Exception::class);
+
```
mvorisek: 2x no space - expectException should be as much as close the to tested expr

ideally fix everywhere

src/Model/Scope/Condition.php
```diff
@@ -128,6 +128,24 @@ public function __construct($key, $operator = null, $value = null)
...
         }
+
+        if (is_array($value)) {
+            if (array_filter($value, 'is_array')) {
+                throw (new Exception('Multi-dimensional array as condition value is not supported'))
+                    ->addMoreInfo('value', $value);
+            }
+
+            if (!in_array($this->operator, [
+                self::OPERATOR_EQUALS,
```
mvorisek, line 139: what is the difference between EQUALS and IN operators now?

georgehristov: Equivalent in case of array.

mvorisek: hmm, are we used to use array with `=`?

Am I correct, that `null = null` will match now, but `null = [null]` will be now converted later to `null in (null)`, thus not work as expected?

if so. we should not alow array with `=` and definitelly `in` should accept only single level array

georgehristov: > Am I correct, that `null = null` will match now, but `null = [null]` will be now converted later to null in (null), thus not work as expected?

This is quite edge case. Result of `null = null` is not what you are implying I believe
https://stackoverflow.com/questions/1843451/why-does-null-null-evaluate-to-false-in-sql-server

> `in` should accept only single level array

It is the case already (after this PR merged)

But we will keep equivalence of `=` and `in`. `in` also works with non-array value same as `=`

Merging

mvorisek: > > Am I correct, that `null = null` will match now, but `null = [null]` will be now converted later to null in (null), thus not work as expected?
> 
> This is quite edge case. Result of `null = null` is not what you are implying I believe
> https://stackoverflow.com/questions/1843451/why-does-null-null-evaluate-to-false-in-sql-server

no, no, how WE support it, as far as I know we convert `= null` to `is null`, so `= [null]` MUST NEVER BEHAVE DIFFERENTLY!

> > `in` should accept only single level array
> 
> Done

really? It seems we do not validate/reject `in non-array`

> But we will keep equivalence of `=` and `in`. `in` also works with non-array value same as `=`

now ok, but `= arr` and `in non-arr` must throw, of course also for negative ops

georgehristov: > we convert = null to is null, so = [null] MUST NEVER BEHAVE DIFFERENTLY!

This is handled by DSQL so if there is a problem it should be addressed there.
If any of the array values is null then convert it to `OR field IS NULL` instead of including it in the list

mvorisek: yes - but it will not happed (that easily) if

> now ok, but `= arr` and `in non-arr` must throw, of course also for negative ops

we should address this in Condition constructor to make the validation complete, right?

georgehristov: > = arr and in non-arr must throw

It handles it equivalently now and it is okay. Basically now `=` === `in`

mvorisek: that is bad, we should probably drop support for `=` with array and drop support for `in` with non-single-level-array

PR #686, opened 2020-07-30
Fix query on hasOne referenced model

georgehristov:
Cherrypick from #683

### DIFF
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -634,6 +634,14 @@ class Sql extends Persistence
 
                 [$fx, $field] = $args;
 
+                if (is_string($field)) {
+                    if ($model->hasField($field)) {
+                        $field = $model->getField($field);
+                    } else {
+                        $field = new Expression($field);
+                    }
+                }
+
                 $this->initQueryConditions($model, $query);
                 $model->hook(self::HOOK_INIT_SELECT_QUERY, [$query, $type]);
 
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -367,7 +367,7 @@ class Sql extends Persistence
 
             // simple condition
             if ($condition instanceof Model\Scope\Condition) {
-                $query = $query->where(...$condition->toQueryArguments());
+                $query->where(...$condition->toQueryArguments());
             }
 
             // nested conditions
@@ -378,7 +378,7 @@ class Sql extends Persistence
                     $this->initQueryConditions($model, $expression, $nestedCondition);
                 }
 
-                $query = $query->where($expression);
+                $query->where($expression);
             }
         }
     }
@@ -632,8 +632,8 @@ class Sql extends Persistence
                         ->addMoreInfo('action', $type);
                 }
 
-                $fx = $args[0];
-                $field = is_string($args[1]) ? $model->getField($args[1]) : $args[1];
+                [$fx, $field] = $args;
+
                 $this->initQueryConditions($model, $query);
                 $model->hook(self::HOOK_INIT_SELECT_QUERY, [$query, $type]);
 
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -634,13 +634,7 @@ class Sql extends Persistence
 
                 [$fx, $field] = $args;
 
-                if (is_string($field)) {
-                    if ($model->hasField($field)) {
-                        $field = $model->getField($field);
-                    } else {
-                        $field = new Expression($field);
-                    }
-                }
+                $field = is_string($field) ? $model->getField($field) : $field;
 
                 $this->initQueryConditions($model, $query);
                 $model->hook(self::HOOK_INIT_SELECT_QUERY, [$query, $type]);
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -620,6 +620,10 @@ class Sql extends Persistence
                 $this->initQueryConditions($model, $query);
                 $this->setLimitOrder($model, $query);
 
+                if ($model->loaded()) {
+                    $query->where($model->id_field, $model->id);
+                }
+
                 return $query;
             case 'fx':
             case 'fx0':
--- a/src/Reference/HasOne.php
+++ b/src/Reference/HasOne.php
@@ -231,7 +231,9 @@ class HasOne extends Reference
         $theirModel->onHook(Model::HOOK_AFTER_SAVE, function ($theirModel) {
             $theirValue = $this->their_field ? $theirModel->get($this->their_field) : $theirModel->id;
 
-            $this->getOurField()->set($theirValue)->owner->save();
+            if ($this->getOurFieldValue() !== $theirValue) {
+                $this->getOurField()->set($theirValue)->owner->save();
+            }
 
             $theirModel->reload();
         });
--- a/src/Reference/HasOne.php
+++ b/src/Reference/HasOne.php
@@ -225,10 +225,15 @@ class HasOne extends Reference
             }
         }
 
+        // their model will be reloaded after saving our model to reflect changes in referenced fields
+        $theirModel->reload_after_save = false;
+
         $theirModel->onHook(Model::HOOK_AFTER_SAVE, function ($theirModel) {
             $theirValue = $this->their_field ? $theirModel->get($this->their_field) : $theirModel->id;
 
-            $this->getOurField()->set($theirValue);
+            $this->getOurField()->set($theirValue)->owner->save();
+
+            $theirModel->reload();
         });
 
         return $theirModel;
--- a/src/Reference/HasOne.php
+++ b/src/Reference/HasOne.php
@@ -196,7 +196,7 @@ class HasOne extends Reference
     {
         $this->getOurModel()->persistence_data['use_table_prefixes'] = true;
 
-        return $this->getOurModel()->getField($this->our_field);
+        return $this->getOurField();
     }
 
     /**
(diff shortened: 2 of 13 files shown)


src/Persistence/Sql.php
```diff
@@ -628,8 +632,16 @@ public function action(Model $model, $type, $args = [])
...
 
-                $fx = $args[0];
-                $field = is_string($args[1]) ? $model->getField($args[1]) : $args[1];
+                [$fx, $field] = $args;
+
+                if (is_string($field)) {
+                    if ($model->hasField($field)) {
+                        $field = $model->getField($field);
+                    } else {
+                        $field = new Expression($field);
```
mvorisek: real usecases? we probably need to support args as well

tests/ReferenceTest.php
```diff
@@ -71,6 +95,27 @@ public function testModelProperty()
...
+        $company->addField('name');
+        $user->hasOne('Company', [$company, 'our_field' => 'company_id', 'their_field' => 'id']);
+
+        $order = new Model($this->db, 'order');
+        $order->addField('company_id');
+        $order->addField('amount', ['default' => 20]);
+
+        $company->hasMany('Orders', [$order]);
+
+        $this->assertEquals(20, $user->ref('Company')->ref('Orders')->get('amount')); // 'amount' default value
```
DarkSide666: Useless test - it will always return default value

What you should do I think is:
1. move this test to ReferenceSqlTest class and revert these changes from ReferenceTest
2. instead of get(amount) use count(...->export()) or export([amount]) or sum(amount) or something like that to actually test which records will be selected
3. Also add tests for unloaded and loaded models
```
$user->unload()->ref('Company')->ref('Orders')->export(['amount']);
$user->load(1)->ref('Company')->ref('Orders')->export(['amount']);
```

mvorisek: where is the discussed usecase?

this test seems quite complex but your usecase was about one very simple thing

this test has only one assertion for a default val

src/Persistence/Sql.php
```diff
@@ -620,6 +620,10 @@ public function action(Model $model, $type, $args = [])
                 $this->initQueryConditions($model, $query);
                 $this->setLimitOrder($model, $query);
 
+                if ($model->loaded()) {
+                    $query->where($model->id_field, $model->id);
```
mvorisek, line 624: action materializes/render, so this makes sense (like `action('update')` uses the current ID)

but `action('select')` does not :( (so action should probably ignore loaded ID - how is/should `action()` behaviour be defined?)

before merge, confirm, that all uses of `action('field')` expect restricted field - does it make any sense to be used somewhere without the restriction to id?

src/Reference/HasOne.php
```diff
@@ -225,10 +225,15 @@ public function ref($defaults = []): Model
...
+        $theirModel->reload_after_save = false;
+
         $theirModel->onHook(Model::HOOK_AFTER_SAVE, function ($theirModel) {
-            $ourValue = $this->their_field ? $theirModel->get($this->their_field) : $theirModel->id;
+            $theirValue = $this->their_field ? $theirModel->get($this->their_field) : $theirModel->id;
+
+            $this->getOurField()->set($theirValue)->owner->save();
 
-            $this->getOurModel()->set($this->our_field, $ourValue);
+            $theirModel->reload();
```
DarkSide666, line 238: I think this is wrong.
Imagine we have `$user->hasOne(City::class)->addTitle()`, so it's `user.city_id = city.id` in SQL.

Now I change name of users city from Riga to London and save it.
What happens then? There is no need to save ourModel if theirField was not the one changed (dirty) in theirModel.

Why we need to reload city model (theirModel)? What maybe should be reloaded is users (ourModel) in case of some linked (title_field) or calculated fields.

Does it matter if ourModel is already loaded or not?

tests/ReferenceTest.php
georgehristov: tests expanded now from the #683

src/Reference/HasOneSql.php
```diff
@@ -54,12 +54,15 @@ function (Model $ourModel) use ($theirFieldName) {
                     return $ourModel->refLink($this->link)->action('field', [$theirFieldName])->reset('order');
                 },
             ],
-            $defaults
+            $defaults,
```
DarkSide666, line 57: Am I right  that there are no functional changes in HasOneSql class? Only some prettifying.

DarkSide666 (changes requested):
See comments. Also not sure what exactly is fixed and where.

src/Persistence/Sql.php
DarkSide666: There definitely are usecases for Expression.

tests/ReferenceTest.php
DarkSide666: Good. Now also need to test same things with unloaded user model.
It should return all orders of all users.

src/Persistence/Sql.php
mvorisek: but it need args!

tests/ReferenceTest.php
georgehristov: Done

src/Persistence/Sql.php
georgehristov: Reverted to original functionality

mvorisek: what is currently the best alternative /wo the planned optional trait?

tests/ReferenceSqlTest.php
```diff
@@ -350,6 +350,56 @@ public function testOtherAggregates()
...
+        $user->load(1);
+
+        $this->assertSame([
+            ['id' => '1', 'company_id' => '1', 'description' => 'Vinny Company Order 1', 'amount' => '50'],
+            ['id' => '3', 'company_id' => '1', 'description' => 'Vinny Company Order 2', 'amount' => '15'],
+        ], $user->ref('Company')->ref('Orders')->export());
+
+        $user->unload();
+
+        $this->assertSame([
```
mvorisek: so @DarkSide666 was right with his guess yesterday? :)

tests/ReferenceSqlTest.php
```diff
@@ -350,6 +350,56 @@ public function testOtherAggregates()
         $this->assertNull($l->get('chicken5'));
     }
 
+    public function testReferenceHasOneTraversing()
```
mvorisek, line 353: and hasMany?

src/Reference/HasOne.php
georgehristov: I made 2 changes here:
- delay the reloading of $theirModel to after setting $ourFieldValue to $theirFieldValue.
This is necesary as the `reload` cannot find the new city record (query on city is limited to 'only cities whose user.city_id = new city.id'. But there are no such users as the user.city_id has not been set to city.id yet. So we first set user.city_id=city.id and then reload city.
- save $ourModel when $ourFieldValue updated (original code was setting it but not saving). Now I added also check to only save when changed

tests/ReferenceSqlTest.php
georgehristov: There are tests for hasMany already

tests/ReferenceSqlTest.php
georgehristov: No, I was right. See the updated tests.

tests/ReferenceTest.php
georgehristov: All updated

tests/ReferenceSqlTest.php
mvorisek: I see, is cloning on traversing documented?

src/Reference/HasOneSql.php
georgehristov: Only one major change in `HasOneSql::ref`. The below code is now executed irrespective if ourModel loaded or not (originally was only when not loaded)
```
$ourFieldExpression = $ourModel->action('field', [$ourField]);

return $theirModel->addCondition($theirField, $ourFieldExpression);
```

tests/ReferenceSqlTest.php
mvorisek: documented here - https://github.com/atk4/data/blame/701532620fc597835002c2f7cfc58626e032bc26/docs/references.rst#L82

one central clone, that is good, but:
- the method should be named more like `getTheirModelClone()` or `createTheirModelClone`
- is it wanted that when model is created by Closure, the model can be not cloned?

georgehristov: The theirModel is generated every time you call `getTheirModel` so the name is correct.
The wording in the phpdoc is wrong. It is not a clone but a fresh object

mvorisek: yes, clone or factory... can be a clone, not a fresh object

georgehristov: Cannot be a clone as conditions are assigned to it depending on whether ourModel is loaded or not.

It is good as it is now. I have only solved the issue with referencing.

No need to waste time now on discussion abt method names, etc

src/Persistence/Sql.php
DarkSide666: Not 100% sure about this. Maybe there is some use-case where this is supposed to generate
`SELECT field FROM table` without ID condition even if model is already loaded.

src/Reference/HasOne.php
DarkSide666: Ok, probably that's correct.

DarkSide666 (approved):

src/Persistence/Sql.php
mvorisek: That is why I messages you about these changes - I think this is ok, as user can easily unload (or even clone + unload) and then call this method...

georgehristov: No issues have been observed so far in testing, we will see in realtime use but this is true - model unload should be the solution if such usage exists

PR #688, opened 2020-07-31
Refactor Model::action to Model::toQuery

mvorisek:
- doc might need an update (search `(?<!\w)action(?!\w)`)
- `function action` is not renamed in persistences (called only in one place - Model::toQuery)
- Array persistence does not use query :) - how to solve it?

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -507,7 +507,7 @@ class Model implements \IteratorAggregate
                 // or it can be simple string = field name
                 $name = $field;
                 $field = [];
-            } elseif (is_array($field) && isset($field[0]) && is_string($field[0])) {
+            } elseif (is_array($field) && is_string($field[0] ?? null)) {
                 // or field name can be passed as first element of seed array (old behaviour)
                 $name = array_shift($field);
             } else {
@@ -1379,7 +1379,6 @@ class Model implements \IteratorAggregate
     /**
      * Load record by condition.
      *
-     * @param mixed $field_name
      * @param mixed $value
      *
      * @return $this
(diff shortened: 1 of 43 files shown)


src/Model.php
```diff
@@ -1881,7 +1881,7 @@ public function atomic($f, Persistence $persistence = null)
      *
      * @return Query
      */
-    public function action($mode, $args = [])
+    public function toQuery($mode, $args = [])
```
georgehristov, line 1883: We should keep `action` forwarding to `toQuery` with a deprecation notice

georgehristov:
Please rebase it on [fix/has-one-ref](https://github.com/atk4/data/tree/fix/has-one-ref) and include deprecation notice on the `Model::action` method

georgehristov:
> Array persistence does not use query :) - how to solve it?

Implement `Persistence\QueryInterface` and require `Persistence::action` to return this interface.

It will have the methods in `Query` and we apply this to `Action/Iterator` (it needs to be moved out of there to `Persistence/Array_/Query`)
And we will have 
- `Persistence\Array_\Query implements Persistence\QueryInterface`.
- `Persistence\Sql\Query extends dsql\Query implements Persistence\QueryInterface`. Will be empty class :)

@mvorisek will you take it up?

src/Model.php
DarkSide666: Yes this is a must because `action()` method is used in projects quite widely.
Also this PR probably need BC-break label.

mvorisek: labeled

mvorisek:
closed in favor of https://github.com/atk4/data/pull/690

PR #690, opened 2020-08-02
Abstract Query implementation from persistence specific ones

georgehristov:
The concept unifies the persistence query objects which must extend `Persistence\AbstractQuery`
In the case of Persistence\Sql\Query the calls are forwarded to underlying DSQL object.

This abstracts the link to the actual query engine and unifies the use of `Persistence\Query` object in atk4/data irrelevant of persistence type.

The `Model::load` functionality has been moved to the Model object strictly leaving Persistence object to care only about raw data CRUD.

The Persistence CRUD methods have been moved to the generic Persistence class which relies on storage specific Query objects to perform the interaction with the data storage engine (Sql, Array (memory), Csv).

Array and Csv Query engines extend `Persistence\IteratorQuery`. 
- Array_ persistence is using `\ArrayIterator` 
- Csv persistence is using `\SplFileObject` which is also an iterator

If the concept is accepted it can be further worked on:
- introduce `Model\QueryTrait` to introduce direct access to query methods (insert, select, delete, etc) and use in the model object
- introduce optional `Model\QueryAggregatesTrait` to introduce direct access to aggregate methods (sum, avg, etc) and use in the model object
- optimize code

BC break:
- `Model::action` renamed to `Model::toQuery`
- `Persistence::action` renamed to `Persistence::query`


src/Persistence/Sql/Query.php
```diff
@@ -0,0 +1,339 @@
...
+ * @method Query limit($limit, $offset = 0)
+ * @method Query order($fields)
+ * @method Query getDebugQuery()
+ * @method Query render()
+ * @method Query mode()
+ */
+class Query implements Persistence\QueryInterface, Expressionable
+{
+    /** @var Model */
+    protected $model;
```
mvorisek: it seems we do not need a model here

src/Persistence/Sql/Query.php
```diff
@@ -0,0 +1,339 @@
...
+                }
+
+                $this->limit($limit, $offset ?? 0);
+            }
+        }
+
+        return $this->dsql;
+    }
+
+//     public function __call($method, $args)
```
mvorisek: you was using this probably in debug, yes, we should probably not left it here - any specific impl. should be queried thru ->dsql->method()

src/Persistence/Sql/Query.php
```diff
@@ -0,0 +1,339 @@
...
+use atk4\data\Field;
+use atk4\data\FieldSqlExpression;
+use atk4\data\Model;
+use atk4\data\Persistence;
+use atk4\dsql\Expression;
+use atk4\dsql\Expressionable;
+use atk4\dsql\Query as DsqlQuery;
+
+/**
+ * @method Query get()
```
mvorisek: anotations not needed, use phpdoc above implemented methods directly

based on Liskov substitution principle, I think we can anotate return type directly (thus even no need for phpdoc)

src/Persistence/Sql/Query.php
georgehristov: In the query we definitely need the Model

src/Persistence/Sql/Query.php
georgehristov: This converts Model::order array to dsql

src/Persistence/Sql/Query.php
georgehristov: Note `\atk4\data\Persistence\Query\Sql` does not extend `\atk4\dsql\Query`, it just forwards the calls to it

mvorisek: but it impl. `QueryInterface`, which seems to mix query and result methods.

src/Persistence.php
```diff
@@ -18,6 +18,19 @@ class Persistence
     use \atk4\core\NameTrait;
     use \atk4\core\DiContainerTrait;
 
+    /** @const string */
+    public const HOOK_INIT_SELECT_QUERY = Persistence\AbstractQuery::HOOK_INIT_SELECT;
```
mvorisek: seems like copied - if moved, deprecate with comment on the source side!

src/Persistence/AbstractQuery.php
```diff
@@ -0,0 +1,271 @@
...
+        } catch (Exception $e) {
+            throw (new Exception('Unable to execute query', 0, $e))
+                ->addMoreInfo('query', $this->getDebug())
+                ->addMoreInfo('message', $e->getMessage());
+        }
+
+        return $result;
+    }
+
+    public function execute(): iterable
```
mvorisek: execute of like delete is iterable as well?

src/Persistence/Array_/Query.php
```diff
@@ -2,48 +2,188 @@
...
+
+        return $row;
+    }
+
+    /**
+     * Return one value from one row of data.
+     *
+     * @return mixed
+     */
+    public function getOne()
```
mvorisek: like in dsql, it should have the string `?string` return type

src/Persistence/AbstractQuery.php
```diff
@@ -0,0 +1,271 @@
...
+
+abstract class AbstractQuery implements \IteratorAggregate
+{
+    public const MODE_SELECT = 'SELECT';
+    public const MODE_UPDATE = 'UPDATE';
+    public const MODE_INSERT = 'INSERT';
+    public const MODE_DELETE = 'DELETE';
+
+    /** @const string */
+    public const HOOK_INIT_SELECT = self::class . '@initSelect';
```
mvorisek: also copied from somewhere?

src/Model.php
```diff
@@ -1886,14 +1885,44 @@ public function action($mode, $args = [])
     /**
      * @param string $mode
      * @param array  $args
-     *
-     * @return Query
      */
-    public function toQuery($mode, $args = [])
+    public function toQuery($mode, $args = []): Persistence\AbstractQuery
```
mvorisek: if this should be used a lot, we should redefined to `...$args` - or do we use/want to use assoc `$args` for something else than alias?

src/Persistence/AbstractQuery.php
georgehristov: This I removed as it is wrong. Only results of SELECT are iterable

src/Persistence.php
georgehristov: phpdocs are still pending. QUERY constants were moved to AbstractQuery. This is for BC and will be commented accordingly

src/Model.php
georgehristov: My idea is to eventually directly use the query method on the model variable:
`$invoices->where('vendor_name', 'Test')->get();`

The model scope is automatically applied and the query methods are used to further adjust the results.

src/Persistence.php
mvorisek: very easy to forget, always comment before even you paste it!

src/Persistence/Array_/Query.php
georgehristov: `?string` clashes with current setup

https://github.com/atk4/data/blob/e64b392d057243510e62be797ff405bc598ff06a/tests/PersistentArrayTest.php#L333

mvorisek: Hmm, quite important, please look at this, there is something wrong, we need same result level across all persistences.

src/Model.php
mvorisek: wdym, your example does not use Query at all.

src/Persistence/Array_/Query.php
georgehristov: Fully agree results of the query should be uniform across the persistences.
But I would split this in 2 stages:
1. finlize and merge the reworked classes in fully BC way
2. apply limitiations as deemed necessary which may introduce BC-breaks. It will be easy once 1. is in place.

src/Model.php
georgehristov: Imagine below in Model with anotations to have hints in IDE. The methods we would like performing additional statements (like wrap `delete` in a transaction, hooks, etc) we can keep as method in Model.

```
    public function __call($method, $args)
    {
        return $this->toQuery($method, $args);
    }
```

georgehristov: Above would not work because of dynamic methods so we can create a QueryTrait with the query methods

mvorisek: I think we shifted a lot from original comment and you mix so many things at once
- (current/Model) toQuery methods are known before, there is not need to use dynamic methods
- annotations are used usually for extra info but still with impl. defined - the only exception are lazy properties, which can be not defined (only phpdoced) or unset at create time (but this is a little fragile as constructor can be skipped with refl)

src/Persistence/Sql/Query.php
```diff
@@ -336,20 +272,18 @@ protected static function fillWhere(DsqlQuery $query, Model\Scope\AbstractScope
...
-        try {
-            return $this->select()->execute();
-        } catch (\PDOException $e) {
-            throw (new Exception('Unable to execute iteration query', 0, $e))
-                ->addMoreInfo('query', $this->getDebugQuery())
-                ->addMoreInfo('message', $e->getMessage())
-                ->addMoreInfo('model', $this->model)
-                ->addMoreInfo('scope', $this->model->scope()->toWords());
-        }
+        return $this->tryExecute();
```
mvorisek: try sounds like we do not throw at all, really? intended?

georgehristov: It is exactly the opposite 😄 
wrapping the exception and adding query info to it
any suggections about naming?

src/Persistence/Sql.php
```diff
@@ -434,9 +436,7 @@ public function tryLoad(Model $model, $id): ?array
      */
     public function load(Model $model, $id): array
     {
-        $data = $this->tryLoad($model, $id);
-
-        if (!$data) {
+        if (!$data = $this->tryLoad($model, $id)) {
```
mvorisek: please, do not assign in separate lines, also do not assign more variables per one line

src/Persistence/Array_/Query.php
```diff
@@ -287,7 +233,7 @@ public function aggregate($fx, $field, $alias = null, $coalesce = false)
      *
      * @return $this
      */
-    protected function initWhere()
+    protected function initWhere(): void
```
mvorisek: phpdoc still return $this

src/Persistence/Sql/Query.php
georgehristov: Maybe I just include the try wrap in the execute method so it always adds the query debug info

src/Model.php
```diff
@@ -1888,9 +1888,41 @@ public function action($mode, $args = [])
      */
     public function toQuery($mode, $args = []): Persistence\AbstractQuery
     {
-        $this->checkPersistence('action');
+        $this->checkPersistence('query');
 
-        return $this->persistence->action($this, $mode, $args);
+        // start backward compatibility -->
```
mvorisek, line 1860: new way?

we should also deprecate, we can not provide BC and grow our code forever

src/Persistence/Sql/Query.php
mvorisek: `execute` :) it does not try anything, it only nest exception with better info - also, `Unable` is wrong, unable means you was not able to so mething instead of something that was able to exec failed

georgehristov: yes, figured out execute only is best and implemented it. Exception message also changed

src/Persistence/Sql.php
georgehristov: Reverted

src/Persistence/Sql.php
```diff
@@ -436,7 +436,9 @@ public function tryLoad(Model $model, $id): ?array
      */
     public function load(Model $model, $id): array
     {
-        if (!$data = $this->tryLoad($model, $id)) {
+        $data = $this->tryLoad($model, $id);
+
```
mvorisek: and no space ;-) - we load and "assert" immediatelly the loaded val...

src/Model.php
georgehristov: It is in the original PR description above but it is of course up for discussion.
And `Model::toQuery` will not be removed.

I suggest now I wrap up first stage here.
I will include all phpdocs and work on the comments on the code so far then we merge.
The `Model\QueryTrait` and `Model\QueryAggregatesTrait` will be easy to implement as either optional or standard functionality

src/Persistence/Array_/Query.php
```diff
@@ -2,65 +2,202 @@
...
+
+    /**
+     * Calculates SUM|AVG|MIN|MAX aggragate values for $field.
      *
-     * @return \atk4\data\Action\Iterator
+     * @param string $fieldName
+     * @param string $alias
      */
-    public function aggregate($fx, $field, $coalesce = false)
+    protected function initAggregate(string $functionName, $field, string $alias = null, bool $coalesce = false): void
```
mvorisek: 👍 for $fx -> $functionName

src/Persistence/Array_/Query.php
georgehristov: updated

src/Persistence/Sql.php
georgehristov: t is not so readable when no space. I keep it like it was originaly

src/Model.php
georgehristov: Up for discussion - if and when to deprecate

src/Persistence/Sql.php
mvorisek: it is, sequentially readable and logically grouped

src/Persistence/Sql.php
```diff
@@ -456,39 +439,25 @@ public function tryLoadAny(Model $model): ?array
...
-                    ->addMoreInfo('id_field', $model->id_field)
-                    ->addMoreInfo('data', $data);
-            }
+            $this->loadModelIdFromData($model, $data);
         }
 
         return $data;
     }
 
+    protected function loadModelIdFromData(Model $model, $data): void
```
mvorisek: `setModelIdFromData`

src/Persistence/AbstractQuery.php
```diff
@@ -328,16 +438,30 @@ protected function executeQueryWithDebug(\Closure $fx)
         }
     }
 
+    public function getIterator(): iterable
+    {
+        return $this->execute();
+    }
+
+    /**
+     * Return the model the query runs on.
```
mvorisek: comments like this does not provide any value, they only distract or even confuse if formulated not-up-too date - remove

src/Persistence/AbstractQuery.php
```diff
@@ -192,13 +254,19 @@ protected function setMode($mode): self
...
+     */
     public function where($fieldName, $operator = null, $value = null): self
     {
         $this->scope->addCondition(...func_get_args());
 
         return $this;
     }
 
+    /**
+     * Limit scope to only records with $id (leaves model scope intact).
```
mvorisek: `leaves model scope intact` is property of Query, thus it should be not duplicated in the comments

then comment is unneeded :)

src/Persistence/AbstractQuery.php
```diff
@@ -53,11 +57,19 @@ public function __construct(Model $model)
         $this->limit = $model->limit;
     }
 
+    /**
+     * Find and return data from record with $id or NULL if none found.
```
mvorisek: this comment is on edge, but programmers will imply this from the method signature - drop as well

src/Persistence/AbstractQuery.php
```diff
@@ -53,11 +57,19 @@ public function __construct(Model $model)
...
+     */
     public function find($id): ?array
     {
         return $this->whereId($id)->getRow();
     }
 
+    /**
+     * Setup query as selecting list of $fields or all if $field = NULL.
```
mvorisek, line 55: the only valueable/not implied comment part is `Setup query` only - this is bad design

src/Persistence/AbstractQuery.php
```diff
@@ -111,8 +146,14 @@ public function delete($id = null): self
         return $this;
     }
 
+    /**
+     * Initiate the delete operation in the child class.
```
mvorisek: zero value vs. method name

src/Persistence/Sql/Query.php
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    public function getDebug(): array
+    {
+        return array_merge([
+            'sql' => $this->dsql->getDebugQuery(),
+        ], parent::getDebug());
+    }
+
+    public function __call($method, $args)
```
mvorisek, line 280: Better to declare the 3 methods manually and abstract the impl completely

georgehristov: Too many of the underlying DSQL methods are used
I can expand the annotation

mvorisek: if like 5 now keep it for now - but do we really want to offer all of these in the data? that is the main question...

in Query itself, we can use ->dsql->

georgehristov: Later we can consider deprecating this but let's keep this direct forward for now.

src/Persistence.php
```diff
@@ -18,6 +18,7 @@ class Persistence
     use \atk4\core\NameTrait;
     use \atk4\core\DiContainerTrait;
 
+    // backward compatibility - will be removed in jul-2021 -->
```
mvorisek: Max 1/2 year, we probably want even less (as deprecated code distracts in dev)

I propose to switch to 3 month deprecation period or new minor (d.x.d) ver, so we can deprecate with comments like "will be remove in the next version 2.3"

georgehristov: I am still doubtful about the hooks.

Maybe we keep them in Persistence as they were?
Or keep the forwarded like now (do not deprecate)

src/Persistence/Sql/Query.php
mvorisek: more about better abstraction, code otherwise clean, fine with this now

src/Persistence.php
mvorisek: they are related with Sql only, thus deprecating them from Persistence is correct

let's set the depr deriod to 1/2 year

I also mailed Dark to check if we can agree on the "new minor version" depr period for future changes

src/Persistence/AbstractQuery.php
```diff
@@ -0,0 +1,389 @@
...
+     * Get array of records matching the query.
+     */
+    public function get(): array
+    {
+        return $this->executeQueryWithDebug(function () {
+            return $this->doGet();
+        });
+    }
+
+    abstract protected function doGet(): array;
```
mvorisek, line 307: I would prefer `_` prefix instead even if not advised by PSR - it is then much clearer what it does.

georgehristov: I don't like the `_` prefix as then we have several methods with same name `get`, `_get`, `__get`.
Not very descriptive.
The `do` prefix is more clear and is already used in UserActions for instance.

mvorisek: not critical to me... would be good if you can get @DarkSide666 to review this PR too

src/Persistence.php
georgehristov: done

src/Persistence/IteratorQuery.php
```diff
@@ -0,0 +1,393 @@
...
+use atk4\data\Exception;
+use atk4\data\Field;
+use atk4\data\Model;
+use atk4\data\Model\Scope\Condition;
+use atk4\data\Persistence;
+
+/**
+ * Class to perform queries on Array_ persistence.
+ */
+class IteratorQuery extends AbstractQuery
```
mvorisek, line 16: ArrayIteratorQuery?

georgehristov:
Now it is better 😄 
diff

mvorisek:
> Now it is better 😄
> diff

Very good, I hope we can trash more! :D

About review: still waiting on @DarkSide666, I will review it then as we already did some iterations together.

src/Persistence/Array_.php
```diff
@@ -31,15 +31,9 @@ public function __construct(array $data = [])
...
     protected $lastInsertIds = [];
 
-    /**
-     * @deprecated TODO temporary for these:
-     *             - https://github.com/atk4/data/blob/90ab68ac063b8fc2c72dcd66115f1bd3f70a3a92/src/Reference/ContainsOne.php#L119
-     *             - https://github.com/atk4/data/blob/90ab68ac063b8fc2c72dcd66115f1bd3f70a3a92/src/Reference/ContainsMany.php#L66
-     *             remove once fixed/no longer needed
-     */
-    public function getRawDataByTable(string $table): array
+    public function getRawDataIterator($table): \Iterator
```
mvorisek: 👍 for removing `getRawDataByTable()`

but isn't the `getRawDataIterator` basically the same? do we want to offer raw data at all?

src/Persistence/IteratorQuery.php
georgehristov: No because it is used for various iterators for persistences.
For `Persistence\Array_` we pass `\ArrayIterator`
For `Persistence\Csv` we pass `\SplFileObject` which is also an iterator

src/Reference/ContainsOne.php
```diff
@@ -98,8 +98,8 @@ public function ref($defaults = []): Model
         // set some hooks for ref_model
         foreach ([Model::HOOK_AFTER_SAVE, Model::HOOK_AFTER_DELETE] as $spot) {
             $theirModel->onHook($spot, function ($theirModel) {
-                $row = $theirModel->persistence->getRawDataByTable($this->table_alias);
-                $row = $row ? array_shift($row) : null; // get first and only one record from array persistence
+                $rows = iterator_to_array($theirModel->persistence->getRawDataIterator($this->table_alias));
+                $row = $rows ? array_shift($rows) : null; // get first and only one record from array persistence
```
mvorisek: is this correct when more than 1 result is returned?

when something is 1:1, only (zero/null or) one result should be returned only, otherwise throw, right?

src/Persistence/Array_.php
georgehristov: Question here is if we introduce an Interface for the persistence with 3 methods
`getRawDataIterator`
`setRawData`
`unsetRawData`

Now used in both `Array_` and `Csv`

mvorisek: we should never work on raw data, as the structure can be completely different, even if "Array" persistence and we was almost there...

georgehristov: Only Query objects work on raw data. For sql we use the link to SQL server, for other persistences the Persistence object provides access to raw data.
If you figure out a way to limit the access nicely then we can implement.

src/Reference/ContainsOne.php
georgehristov: I modified this now to how it should be - using Query

src/Persistence/IteratorQuery.php
```diff
@@ -311,7 +313,7 @@ protected function evaluateIf($v1, $operator, $v2): bool
     {
         switch (strtoupper((string) $operator)) {
             case Condition::OPERATOR_EQUALS:
-                $result = is_array($v2) ? $this->evaluateIf($v1, Condition::OPERATOR_IN, $v2) : $v1 == $v2;
+                $result = is_array($v2) ? $this->evaluateIf($v1, Condition::OPERATOR_IN, $v2) : $v1 === $v2;
```
mvorisek, line 323: very good, all repos except this uses strict types already, what is left to use strict types also for this repo?

georgehristov: One of the remaining points in this repo is to streamline the Field typecasting. It has to do with the strict types.

src/Persistence.php
```diff
@@ -18,6 +20,21 @@ class Persistence
     use \atk4\core\NameTrait;
     use \atk4\core\DiContainerTrait;
 
+    // backward compatibility - will be removed in dec-2020 -->
```
mvorisek: this is new, why it should be removed later?

georgehristov: We have hook constants in Persistence forwarded to hook constants in AbstractQuery.

We can keep it like it is without deprecation, I am for it but up for discussion.

src/Model.php
```diff
@@ -1862,21 +1803,60 @@ public function atomic(\Closure $fx, Persistence $persistence = null)
...
+                ->addMoreInfo('type', $mode);
+        }
+
+        return $this->persistence->query($this)->{$mode}(...$args);
+    }
+
+    /**
+     * Get query object to perform query on raw persistence data.
+     */
+    public function query(): Persistence\AbstractQuery
```
mvorisek: the issue is that it does not return query, but instead it creates a new one, thus toQuery :)

src/Persistence/AbstractQuery.php
DarkSide666: `doGet` naming looks fine.

src/Model.php
```diff
@@ -1944,3 +1924,24 @@ public function __debugInfo(): array
 
     // }}}
 }
+
+class RecordNotFoundException extends Exception
```
DarkSide666: This should probably be defined in separate file.

src/Model.php
georgehristov: `Model::query` changed to `Model::toQuery`

src/Model.php
mvorisek: yes

georgehristov: There is one more class `ArrayCallbackIterator` in the `IteratorQuery` file.
Move out?

mvorisek: yes, always, max. 1 class per file (at least because autoload will not work otherwise)

tests/ScopeTest.php
```diff
@@ -298,7 +298,7 @@ public function testConditionOnReferencedRecords()
             $user->addCondition('Tickets/user/country_id/Users/country_id/Users/name', '!=', null); // should be always true
         }
 
-        $this->assertEquals(2, $user->toQuery('count')->getOne());
+        $this->assertEquals(2, $user->toQuery()->count()->getOne());
```
mvorisek: This may look better at the first sight, but:
- can `$user->toQuery()` be reused? I think no.
- what is the purpose of "uninitialized/unbuilded" query object (`$user->toQuery()`)?

georgehristov: Purpose is to access the raw query / data.
In case of `count` I would say we can introduce a method `Model::count` directly

src/Model.php
```diff
@@ -1684,6 +1684,14 @@ public function export(array $fields = null, $key_field = null, $typecast_data =
         return $res;
     }
 
+    /**
+     * Number of records in current model scope
+     */
+    public function getCount(): int
```
mvorisek, line 1722: It is not get... I think we should use something like `toQuery(Query::MODE_COUNT)->getOne()`. And then maybe `Model::countRecords`, but I think we do not use this terminology in Model now...

src/Model.php
georgehristov: The idea is that these classes are used withing the autoloaded class or its children.

But if it is more clear then I will separate.

romaninsh (changes requested):
i like where it's heading but please add documentation.

georgehristov:
@romaninsh for sure but I would like to settle on the code first to avoid writing and rewriting.
Update on docs is compulsory

georgehristov:
@romaninsh @DarkSide666 @mvorisek 

One design issue to solve is whether to use a single pair of hooks for Query execution `HOOK_BEFORE_EXECUTE` and `HOOK_AFTER_EXECUTE` or maintain operation specific pairs.
The query object is passed on to the hook closure as an argument so you can get the operation by `$query->getMode()`.
The disadvantage is that such hooks are called on each operation.

https://github.com/atk4/data/blob/2d47db6537668fecb0b8fd87880d461b441ef6db/src/Persistence/AbstractQuery.php#L279-L296

romaninsh:
@georgehristov will looked quickly through, but will look again.

src/Persistence/AbstractQuery.php
```diff
@@ -284,8 +284,6 @@ public function execute()
 
 //          $this->model->hook(self::HOOK_BEFORE_EXECUTE, [$this]);
 
-            $this->hookOnModel('HOOK_BEFORE_' . $this->getMode(), [$this]);
```
mvorisek: HOOK_BEFORE_EXECUTE is commented, but it seems this hook should be outer based on HOOK_AFTER_EXECUTE...

georgehristov: Question is to keep `HOOK_BEFORE_INSERT`, `HOOK_BEFORE_UPDATE`, etc or replace all with single `HOOK_BEFORE_EXECUTE`

tests/ScopeTest.php
mvorisek: I highly prefer to not enlarge Model, we should to exactly the oposite

src/Model.php
mvorisek: No assumptions when they are loaded, never, I think it is also enforced by some PSR.

src/Persistence/AbstractQuery.php
mvorisek: I would say no strongly - how would you otherwise hook only for update for example?

georgehristov: You can distinguish using `$query->getMode()` in the closure

mvorisek: I analysed `HOOK_(BEFORE)_(delete|update|insert)` matches now:

- `getMode()` in hook itself is ok for me, but user usually needs to hook for one type of operation only (for ex. I use it for Doctrine, but it can be easily replaced), like -2/10 helpful
- you probably want it because:
image
right? would one hook for these actually help? I do not think so, <-3, 2>/10 helpful
- 3x in Model:
image
no sure how you want to solve this then... <-8,2>/10 helpful

georgehristov: Thanks for analysis. Will stick to the separate hooks then, no problem

But actually the hooks you found are not the same :)
We are talking about `Query::HOOK_BEFORE_UPDATE` for instance.
It is different to `Model::HOOK_BEFORE_UPDATE`

mvorisek: yes, checking the result is probably the strongest argument + https://github.com/atk4/core/pull/246 (as with single hook the probability the hook is already executing is much higher)

georgehristov: We will keep hook constants defined in Persistence as now.

src/Persistence/AbstractQuery.php
```diff
@@ -281,32 +264,22 @@ public function execute()
...
         });
     }
 
     abstract protected function doExecute();
 
-    protected function hookOnModel($name, $args = []): void
+    protected function hookOnModel($when, $args = []): void
     {
-        $hookSpotConst = self::class . '::' . $name;
+        $hookSpotConst = get_class($this->persistence) . '::HOOK_' . $when . '_' . $this->getMode() . '_QUERY';
```
mvorisek: better to query constant value here directly (use something like `constant(Persistance::class . '::HOOK_' . ...)`)

`string $stage` seems better name vs. `$when`

src/Persistence/AbstractQuery.php
```diff
@@ -281,32 +264,22 @@ public function execute()
...
         if (defined($hookSpotConst)) {
```
mvorisek, line 292: do not check if defined

src/Persistence/AbstractQuery.php
georgehristov: > no, you must query the constant value! (to not rely on any fixed value

This is what I am doing, see 2 rows below. The idea of using `get_class($this->persistence)` is that some persistences can introduce more query modes or hooks

src/Persistence/AbstractQuery.php
georgehristov: And if not defined?

src/Persistence/AbstractQuery.php
mvorisek: > This is what I am doing, see 2 rows below.

I see, already updated my comment... The issue is when you pass bad arg/mode, it will remain undetected...

src/Persistence/AbstractQuery.php
mvorisek: see comment above

src/Persistence/AbstractQuery.php
georgehristov: The only way I can think of is declare all constants for all modes (BEFORE_SELECT, AFTER_SELECT and AFTER_DELETE missing now)

mvorisek: Yes, I think this is the best approach

georgehristov: or use only 2 `BEFORE_EXECUTE_QUERY` and `AFTER_EXECUTE_QUERY`.

Code will be much cleaner then :)

These hooks are used in very special cases anyway

mvorisek: you are right, we should use this approach, but dispatch like before (ie. multiple times for ex. in https://user-images.githubusercontent.com/2228672/90612356-37159b00-e208-11ea-9bb9-66d80f5c7074.png) - if we pass mode as the 1st value, then we can check for the first value like before

https://github.com/atk4/core/pull/246 might be harder later, but that is good if solved "properly" :)

I like this approach with `toQuery(string $mode)`, so we can return initialized Query immediatelly. We should have constant for every MODE_ and like other constants, value should start with class name to be sure we use the constant instead of fixed value.

docs/sql.rst
```diff
@@ -513,7 +513,7 @@ Technically you can also specify expression as a $table property of your model::
         function init(): void {
             parent::init();
 
-            $this->init = $this->expr("call get_report_data()");
+            $this->table = $this->expr('call get_report_data()');
```
mvorisek, line 516: table can be an expr?

docs/sql.rst
```diff
@@ -219,80 +219,80 @@ field expressions will be automatically substituted. Here is long / short format
...
 
-Actions
+Queries
 =======
 
-The most basic action you can use with SQL persistence is 'select'::
+The most basic query mode you can use with SQL persistence is 'select'::
 
-    $action = $model->action('select');
+    $query = $model->toQuery()->select();
```
mvorisek, line 227: returning `$model->toQuery()` is very dangerous as it can not be reused for another query/mode

we should keep `$model->toQuery(MODE::...)`

src/Persistence/AbstractQuery.php
```diff
@@ -0,0 +1,369 @@
...
+
+namespace atk4\data\Persistence;
+
+use atk4\data\Exception;
+use atk4\data\Model;
+use atk4\data\Persistence;
+
+abstract class AbstractQuery implements \IteratorAggregate
+{
+    public const MODE_SELECT = 'SELECT';
```
mvorisek: I prefer to keep LC, this is not SQL.

src/Persistence/IteratorQuery.php
```diff
@@ -0,0 +1,395 @@
...
+
+    public function getDebug(): array
+    {
+        return array_merge([
+            'fields' => $this->fields,
+        ], parent::getDebug());
+    }
+}
+
+class ArrayCallbackIterator extends \IteratorIterator
```
mvorisek: move

docs/sql.rst
georgehristov: I have not tested it now but this is in original documentation and I don't see why not as it is handled by dsql template in this case. So [table] in the template is replaced with the expression

src/Persistence/IteratorQuery.php
georgehristov: moved

src/Persistence/AbstractQuery.php
georgehristov: done

docs/sql.rst
georgehristov: Included an exception if mode change attempted.
I prefer keeping the way to set the mode using methods.
If no mode specifically set then default is 'select'

mvorisek:
to be solved here: https://github.com/atk4/ui/pull/1529#discussion_r517374420

src/Persistence/IteratorQuery.php
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    protected function initField($fieldName, string $alias = null): void
+    {
+        if (!$fieldName) {
+            throw new Exception('Field query requires field name');
+        }
+
+        // get first record
+        if ($row = $this->persistence->query($this->model)->select([$fieldName])->getRow()) {
```
mvorisek, line 150: I wonder why this https://github.com/atk4/data/pull/861 bug passed test in this refactoring...

mvorisek:
Hi @georgehristov, the ship sailed already too much to be able to resolve the conflicts easily. Can you please rebase/redo this PR on top of the current develop branch or should this PR be closed?

mvorisek:
@georgehristov I am gonna to close this PR soon, if there is any value outside the rename of "action" to "query", I would happily integrate these improvements as a separate PR.

mvorisek:
Closing due inactivity. The persistence changes might be welcomed, but should be submit without the action->query refactoring, any big refactoring must be submit as a separate PR.

PR #691, opened 2020-08-05
Refactor Model::getFields

georgehristov:
- Introduce constants to declare Model field filter presets
- Adds `Model::FIELD_FILTER_PERSIST` filter preset
- Adds `Model::FIELD_FILTER_ONLY_FIELDS` filter preset
- uses new presets to remove duplicate code
- Provide the possibility to filter fields by Closure

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -73,6 +73,17 @@ class Model implements \IteratorAggregate
     /** @const string Executed when self::onlyFields() method is called. */
     public const HOOK_ONLY_FIELDS = self::class . '@onlyFields';
 
+    /** @const string */
+    public const FIELD_FILTER_SYSTEM = 'system';
+    /** @const string */
+    public const FIELD_FILTER_NOT_SYSTEM = 'not system';
+    /** @const string */
+    public const FIELD_FILTER_VISIBLE = 'visible';
+    /** @const string */
+    public const FIELD_FILTER_EDITABLE = 'editable';
+    /** @const string */
+    public const FIELD_FILTER_PERSIST = 'persist';
+
     // {{{ Properties of the class
 
     /**
@@ -621,19 +632,19 @@ class Model implements \IteratorAggregate
         }
 
         $filterPresets = [
-            'system' => function (Field $field) {
+            self::FIELD_FILTER_SYSTEM => function (Field $field) {
                 return $this->isOnlyFieldsField($field->short_name) && $field->system;
             },
-            'not system' => function (Field $field) {
+            self::FIELD_FILTER_NOT_SYSTEM => function (Field $field) {
                 return $this->isOnlyFieldsField($field->short_name) && !$field->system;
             },
-            'editable' => function (Field $field) {
+            self::FIELD_FILTER_EDITABLE => function (Field $field) {
                 return $this->isOnlyFieldsField($field->short_name) && $field->isEditable();
             },
-            'visible' => function (Field $field) {
+            self::FIELD_FILTER_VISIBLE => function (Field $field) {
                 return $this->isOnlyFieldsField($field->short_name) && $field->isVisible();
             },
-            'persisting' => function (Field $field) {
+            self::FIELD_FILTER_PERSIST => function (Field $field) {
                 if ($field->never_persist) {
                     return false;
                 }
@@ -1654,7 +1665,7 @@ class Model implements \IteratorAggregate
 
         // prepare array with field names
         if ($fields === null) {
-            $fields = array_keys($this->getFields('persisting'));
+            $fields = array_keys($this->getFields(self::FIELD_FILTER_PERSIST));
         }
 
         // add key_field to array if it's not there
--- a/src/Model.php
+++ b/src/Model.php
@@ -577,17 +577,20 @@ class Model implements \IteratorAggregate
         return $this;
     }
 
-    private function checkOnlyFieldsField(string $field)
+    private function checkOnlyFieldsField(string $fieldName)
     {
-        $this->getField($field); // test if field exists
+        $field = $this->getField($fieldName); // test if field exists
 
-        if ($this->only_fields) {
-            if (!in_array($field, $this->only_fields, true) && !$this->getField($field)->system) {
-                throw (new Exception('Attempt to use field outside of those set by onlyFields'))
-                    ->addMoreInfo('field', $field)
-                    ->addMoreInfo('only_fields', $this->only_fields);
-            }
+        if (!$this->isOnlyFieldsField($fieldName) && !$field->system) {
+            throw (new Exception('Attempt to use field outside of those set by onlyFields'))
+                ->addMoreInfo('field', $fieldName)
+                ->addMoreInfo('only_fields', $this->only_fields);
         }
+    }
+
+    public function isOnlyFieldsField(string $fieldName)
+    {
+        return !$this->only_fields || in_array($fieldName, $this->only_fields, true);
     }
 
     /**
@@ -619,16 +622,16 @@ class Model implements \IteratorAggregate
 
         $filterPresets = [
             'system' => function (Field $field) {
-                return $this->isOnlyField($field->short_name) && $field->system;
+                return $this->isOnlyFieldsField($field->short_name) && $field->system;
             },
             'not system' => function (Field $field) {
-                return $this->isOnlyField($field->short_name) && !$field->system;
+                return $this->isOnlyFieldsField($field->short_name) && !$field->system;
             },
             'editable' => function (Field $field) {
-                return $this->isOnlyField($field->short_name) && $field->isEditable();
+                return $this->isOnlyFieldsField($field->short_name) && $field->isEditable();
             },
             'visible' => function (Field $field) {
-                return $this->isOnlyField($field->short_name) && $field->isVisible();
+                return $this->isOnlyFieldsField($field->short_name) && $field->isVisible();
             },
             'persisting' => function (Field $field) {
                 if ($field->never_persist) {
@@ -636,7 +639,7 @@ class Model implements \IteratorAggregate
                 }
 
                 if (!$field->system) {
-                    return $this->isOnlyField($field->short_name);
+                    return $this->isOnlyFieldsField($field->short_name);
                 }
 
                 return true;
@@ -661,11 +664,6 @@ class Model implements \IteratorAggregate
 
             return false;
         }, ARRAY_FILTER_USE_BOTH);
-    }
-
-    public function isOnlyField($fieldName)
-    {
-        return !$this->only_fields || in_array($fieldName, $this->only_fields, true);
     }
 
     /**
--- a/src/Model.php
+++ b/src/Model.php
@@ -592,7 +592,7 @@ class Model implements \IteratorAggregate
 
     private function assertOnlyFieldsField(string $fieldName)
     {
-        $field = $this->getField($fieldName); // test if field exists
+        $field = $this->getField($fieldName);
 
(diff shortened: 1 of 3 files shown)


src/Model.php
```diff
@@ -613,26 +616,49 @@ public function getFields($filter = null): array
...
-                return false;
-            }
-            foreach ($filter as $f) {
-                if (
-                    ($f === 'system' && $field->system)
-                    || ($f === 'not system' && !$field->system)
-                    || ($f === 'editable' && $field->isEditable())
-                    || ($f === 'visible' && $field->isVisible())
-                ) {
+        $filterPresets = [
```
mvorisek: good, but declare inside with if directly - declaring array again and again is too costly for this core thing

mvorisek:
> Provides possibility to filter fields by Closure

I do not get this - we should get all and filter then, what is the advantage of your approach?

georgehristov:
This is also true. Main point was to introduce the 'persisting' preset to avoid duplicate code.
And streamline the code a bit with the ifs and buts.
Passing a Closure as filter is a side-effect

src/Model.php
```diff
@@ -613,26 +616,49 @@ public function getFields($filter = null): array
...
+            'not system' => function (Field $field) {
+                return $this->isOnlyFieldsField($field->short_name) && !$field->system;
+            },
+            'editable' => function (Field $field) {
+                return $this->isOnlyFieldsField($field->short_name) && $field->isEditable();
+            },
+            'visible' => function (Field $field) {
+                return $this->isOnlyFieldsField($field->short_name) && $field->isVisible();
+            },
+            'persisting' => function (Field $field) {
```
mvorisek: `persist` is better name

and move to FIELD_FILTER_ constants

georgehristov: Yes, I was also thinking about declaring in constants, will do

src/Model.php
mvorisek: @georgehristov ping

georgehristov: Not sure what you mean.
We accept string, array of strings (also closures)
And we return matching with OR

mvorisek: In current design you create `$filterPresets` array with callbacks every time `getFields()` is called.

I think we should no create array like this every time and check like before directly.

src/Model.php
```diff
@@ -820,7 +820,8 @@ public function get(string $field = null)
                 $data[$field->short_name] = $this->get($field->short_name);
             }
 
-            return $data;
+            // return $data keys sorted in the order of Model::$only_fields
+            return array_merge($this->only_fields ? array_flip($this->only_fields): [], $data);
```
mvorisek: array_merge? why to sort at all?

georgehristov: Just for equivalence. Not sure if `only_fields` was not used to set the order in `get()` result.
Tests pass without the sorting so can be removed

src/Model.php
georgehristov: Better now?

src/Model.php
mvorisek: yes, return simply `$data`

src/Model.php
```diff
@@ -774,22 +803,15 @@ public function get(string $field = null)
...
-                // collect data for actual fields
-                foreach ($this->only_fields as $field) {
-                    $data[$field] = $this->get($field);
-                }
-            } else {
-                // get all fields
-                foreach ($this->getFields() as $field => $f) {
-                    $data[$field] = $this->get($field);
-                }
+            foreach ($this->getFields(self::FIELD_FILTER_ONLY_FIELDS) as $field) {
```
mvorisek, line 806: `FIELD_FILTER_ONLY_FIELDS` should be removed, and `isOnlyFieldsField` should be used directly

src/Model.php
```diff
@@ -609,37 +625,50 @@ public function isDirty(string $field): bool
...
-                    throw (new Exception('Filter is not supported'))
-                        ->addMoreInfo('filter', $f);
                 }
             }
 
             return false;
         }, ARRAY_FILTER_USE_BOTH);
     }
 
+    protected function fieldMatchesFilter(Field $field, string $filter): bool
```
mvorisek, line 647: quite mixed concept of testing only field with some property

src/Model.php
```diff
@@ -577,25 +590,28 @@ public function allFields()
         return $this;
     }
 
-    private function checkOnlyFieldsField(string $field)
+    private function assertOnlyFieldsField(string $fieldName)
     {
-        $this->getField($field); // test if field exists
+        $field = $this->getField($fieldName); // test if field exists
```
mvorisek: if assigned, remove ` // test if field exists` then

src/Model.php
```diff
@@ -609,37 +625,50 @@ public function isDirty(string $field): bool
...
-                if (
-                    ($f === 'system' && $field->system)
-                    || ($f === 'not system' && !$field->system)
-                    || ($f === 'editable' && $field->isEditable())
-                    || ($f === 'visible' && $field->isVisible())
-                ) {
+        return array_filter($this->fields, function (Field $field, $name) use ($filters) {
+            foreach ($filters as $filter) {
+                if ($this->fieldMatchesFilter($field, $filter)) {
                     return true;
```
mvorisek, line 645: this imply OR - do we want that and support array at all? I think that is illogical, we should probably drop array support, wdyt?

georgehristov: This is how it was, no changes. I suggest to keep it like it is.

src/Model.php
```diff
@@ -609,37 +625,50 @@ public function isDirty(string $field): bool
      *
      * @return Field[]
      */
-    public function getFields($filter = null): array
+    public function getFields($filters = null): array
     {
-        if ($filter === null) {
+        if ($filters === null) {
             return $this->fields;
```
mvorisek: I do not like this mixed concept.

if null, then all
if for ex. `FIELD_FILTER_VISIBLE`, then not visible from all, but also with ONLY FIELDS filter

can we do something about this?

also, it seems that the new constants are not applied to other code

src/Model.php
```diff
@@ -577,25 +590,28 @@ public function allFields()
         return $this;
     }
 
-    private function checkOnlyFieldsField(string $field)
+    private function assertOnlyFieldsField(string $fieldName)
```
georgehristov, line 593: We should probably change to `assertAccessibleField` or similar as naming now does not match the logic in other places.
onlyFieldsField is strictly a field in `only_fields` irrelevant if it is `system` or not

src/Model.php
mvorisek: I highly prefer to address the topic completely when opened.

georgehristov: I see no problem with results being OR

src/Model.php
georgehristov: I agree it is mixed and results are not intuitive
`FIELD_FILTER_VISIBLE`
`FIELD_FILTER_VISIBLE_ONLY_FIELDS`, etc?

mvorisek: it will be not BC anyway, wdyt about additional `$filterOnlyFields` param? then probably no `FIELD_FILTER_ONLY_FIELDS` and `FIELD_FILTER_PERSIST`

georgehristov: I played with that as well but we are in BC-break territory then.
if we set `$filterOnlyFields = true` by default (which is now the case for filter presets) it will not work correctly when `$filters = null`

mvorisek: yes - but is this wanted?

georgehristov: Logically not, historically - yes :)
Logically $filterOnlyFields would be false and all fields returned.
Then gradually filtered out

src/Model.php
mvorisek: but current/expected use case?

src/Model.php
```diff
@@ -625,15 +625,21 @@ public function isDirty(string $field): bool
      *
      * @return Field[]
      */
-    public function getFields($filters = null): array
+    public function getFields($filters = null, $onlyFields = null): array
```
mvorisek: `bool $onlyFields`?

georgehristov: Changed

src/Model.php
georgehristov: Resolved

src/Model.php
mvorisek: but why to null by default?

src/Model.php
```diff
@@ -609,37 +625,56 @@ public function isDirty(string $field): bool
...
      */
-    public function getFields($filter = null): array
+    public function getFields($filters = null, bool $onlyFields = null): array
     {
-        if ($filter === null) {
-            return $this->fields;
-        } elseif (is_string($filter)) {
-            $filter = [$filter];
+        if ($filters === null) {
+            return $onlyFields ? $this->getFields(self::FIELD_FILTER_ONLY_FIELDS) : $this->fields;
```
mvorisek, line 631: no need to recurse, just set $filter...

src/Model.php
```diff
@@ -609,37 +625,56 @@ public function isDirty(string $field): bool
...
+        } elseif (!is_array($filters)) {
+            $filters = [$filters];
         }
 
-        return array_filter($this->fields, function (Field $field, $name) use ($filter) {
-            // do not return fields outside of "only_fields" scope
-            if ($this->only_fields && !in_array($name, $this->only_fields, true)) {
+        $onlyFields = $onlyFields ?? true;
```
mvorisek, line 636: remove this then

src/Model.php
```diff
@@ -609,37 +625,56 @@ public function isDirty(string $field): bool
...
+        switch ($filter) {
+            case self::FIELD_FILTER_SYSTEM:
+                return $field->system;
+            case self::FIELD_FILTER_NOT_SYSTEM:
+                return !$field->system;
+            case self::FIELD_FILTER_EDITABLE:
+                return $field->isEditable();
+            case self::FIELD_FILTER_VISIBLE:
+                return $field->isVisible();
+            case self::FIELD_FILTER_ONLY_FIELDS:
```
mvorisek, line 664: we should not need `FIELD_FILTER_ONLY_FIELDS` then

src/Model.php
georgehristov: To keep BC

`$model->getFields()` returns all fields historically
`$model->getFields('system')` returns system fields in onlyFields

So if `$filter` argument set `$onlyFields` defaults to `true`

georgehristov:
I have another idea:
Presently filters are in first argument as array and considered OR.

We can introduce multiple arguments which will be filters merged with AND.

So `$model->getFields('system', 'only_fields')` will return fields system AND only fields
So `$model->getFields(['system', 'only_fields'])` will return fields system OR only fields

It will not be BC though I believe.

mvorisek:
> 
> I have another idea:
> Presently filters are in first argument as array and considered OR.

I like minimalistic APIs. We extend Model more and more for BC... And the questions is what this more and more code provides? Does it really help and reduce the maintanence requirements? If yes, go ahead!

mvorisek (changes requested):
I like to deduplicate the code.

However, I am not sure if we should mix very unrelated attributes together and offer partial "not" constants

romaninsh (commented):
Good improvement.

Do address feedback from @mvorisek

mvorisek:
TODO rebase on develop, it is based on https://github.com/atk4/data/pull/690, only the last 10 commits are relevant (others are from the mentioned PR) - https://github.com/atk4/data/compare/be3cb50a37...d6307b30c3

mvorisek:
Some changes might be welcomed, but need more discussion.

PR #696, opened 2020-08-26
add loadAny() to Persistence\Array_

PhilippGrashoff:
merge after https://github.com/atk4/data/pull/690 or close (if https://github.com/atk4/data/pull/690 adds `loadAny`)

This method has somehow been missing from Persistence\Array_. This PR adds it and adds a few tests.

However, I see there should be some cleaning up done in PersistenceArrayTest.php in another PR:
Some functions in there are general Persistence tests and have nothing to do with Persistence\Array_.
My proposal is to move PersistenceArrayTest and PersistenceSQLTest to a new directory tests/Persistence/ to meet the structure in /src dir.


georgehristov:
@PhilippGrashoff thanks but we are working on #690 which would resolve this issue

src/Persistence/Array_.php
```diff
@@ -119,11 +119,22 @@ public function tryLoad(Model $model, $id, string $table = null): ?array
         return $this->typecastLoadRow($model, $this->data[$table][$id]);
     }
 
+    /**
+     * Tries to load first available record and return data record.
+     */
+    public function loadAny(Model $model, string $table = null): ?array
+    {
+        $row = $this->tryLoadAny($model, $table);
+        if (!$row) {
```
mvorisek, line 128: you should use strict type here

src/Persistence/Array_.php
```diff
@@ -119,11 +119,22 @@ public function tryLoad(Model $model, $id, string $table = null): ?array
...
+            throw new Exception('No matching records were found!', 404);
```
mvorisek, line 129: does this exception match the SQL persistence incl. the code?

src/Persistence/Array_.php
PhilippGrashoff: Copy-pasted from SQL persistence. You are right, `$row === null` is more explicit.

src/Persistence/Array_.php
PhilippGrashoff: closely, the sql exeption adds more infos:
```

            throw (new Exception('No matching records were found', 404))
                ->addMoreInfo('model', $model)
                ->addMoreInfo('scope', $model->scope()->toWords());
        }
```

To me it seems there is quite some duplicate code between Array_ and SQL Persistences. Maybe we can sort this in #690?

mvorisek: with https://github.com/atk4/data/issues/699 we can drop Array completely...

src/Persistence/Array_.php
mvorisek: put there `$row === null`

mvorisek:
closing as branch seems forcepushed to some other PR

PhilippGrashoff:
@mvorisek yes, I somehow messed up my fork. Opened a new PR with the same content.

PR #697, opened 2020-08-27
Convert Model to Entity (prevent reload with a different record)

mvorisek:
merge after https://github.com/atk4/ui/pull/1429

Initial step to convert Model to Entity

## motivation ##

Currently, Model is used as a definition (scope, fields, ...) and also as a data holder. This design pattern is called active pattern.

This is nice concept to save system resources (less instances), but we can not use/assing/pass Model as truly data, as it can change or be reloaded to a completely different record.

## change ##

To address the motivation we need to ensure that an instance is never resused for an another record. This design pattern is called data mapper.

To allow smooth transition, we keep "active methods" (from current design, methods like `save()`) on Model, but once Model is loaded, we throw if an attempt to reload it to a different ID is made.

## in simple english ##

My idea is that Model class can stay as is for a while, it will continue working as data holder incl. active methods, but once a record is loaded, the model can not be reused to load an another record (that is the point of "entity" - unique instance bound to a one record).

With quite simple change on Model we can enforce the entity behaviour on which we can start to rely on.

User code does not actually need a big change now - see https://github.com/atk4/ui/pull/1429/files which fixes CRUD completely.

With this Model design change, we can integrate with another frameworks.

### If you define Model and then load it no more than once ###

no change needed

### If you define Model and then load/foreach it using iterator ###

you need to use the iterator values to obtain the data, example:

before:
```
foreach ($model as $ignore) {
    var_dump($model->id);
}
```

after:
```
foreach ($model as $entity) {
    var_dump($entity->id); // $model data are unchanged when using iterators
}
```

### If you define Model and then load it multiple times with different/unstable conditions ###

you need to clone Model before load like `$entity = (clone $model)->loadAny();` and then use `$entity` object to read/modify the data

### DIFF
--- a/tests/JoinSqlTest.php
+++ b/tests/JoinSqlTest.php
@@ -431,21 +431,21 @@ class JoinSqlTest extends \atk4\schema\PhpunitTestCase
         $j_country = $j_contact->join('country');
         $j_country->addField('country_name', ['actual' => 'name']);
 
-        $m_u->load(10);
-        $m_u->delete();
+        $m_u2 = (clone $m_u)->load(10);
+        $m_u2->delete();
 
-        $m_u->loadBy('country_name', 'US');
-        $this->assertEquals(30, $m_u->id);
-        $m_u->set('country_name', 'USA');
-        $m_u->save();
+        $m_u2 = (clone $m_u)->loadBy('country_name', 'US');
+        $this->assertEquals(30, $m_u2->id);
+        $m_u2->set('country_name', 'USA');
+        $m_u2->save();
 
-        $m_u->tryLoad(40);
-        $this->assertFalse($m_u->loaded());
+        $m_u2 = (clone $m_u)->tryLoad(40);
+        $this->assertFalse($m_u2->loaded());
 
-        $this->assertSame($m_u->getField('country_id')->join, $m_u->getField('contact_phone')->join);
+        $this->assertSame($m_u2->getField('country_id')->join, $m_u2->getField('contact_phone')->join);
 
-        $m_u->unload();
-        $m_u->save(['name' => 'new', 'contact_phone' => '+000', 'country_name' => 'LV']);
+        $m_u2->unload();
+        (clone $m_u)->save(['name' => 'new', 'contact_phone' => '+000', 'country_name' => 'LV']);
 
         $this->assertEquals(
             [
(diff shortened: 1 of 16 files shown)


mvorisek:
image

georgehristov:
> you need to clone Model before load like $entity = (clone $model)->loadAny(); and then use $entity object to read/modify the data

why not performing the cloning in loadAny already and returning the cloned loaded model?

mvorisek:
> why not performing the cloning in loadAny already and returning the cloned loaded model?

This is planned, there are so many uses of code line ˙$m->load; $m something...˙ So this PR still support it, as long as you do not reload loaded model with a different record.

mvorisek:
> why not performing the cloning in loadAny already and returning the cloned loaded model?

This is planned, there are so many uses of code like ˙$m->load; $m something...˙ So this PR still support it, as long as you do not reload loaded model with a different record.

tests/JoinSqlTest.php
```diff
@@ -431,21 +431,21 @@ public function testDoubleJoin()
         $j_country = $j_contact->join('country');
         $j_country->addField('country_name', ['actual' => 'name']);
 
-        $m_u->load(10);
-        $m_u->delete();
+        $m_u2 = (clone $m_u)->load(10);
```
mvorisek: last 3 tests are failing - @georgehristov, you are the data hero, do you know where the issue is and how to fix the tests?

romaninsh:
Is there a good way to spot incorrect usage in the existing (large) codebase? - like an exception when used incorrectly.

romaninsh (approved):
good work but make sure documentation is updated.

mvorisek:
> 
> 
> Is there a good way to spot incorrect usage in the existing (large) codebase? - like an exception when used incorrectly.

check incorrect foreach usage with regex `foreach.*(model|\$m).*\$(ignore|junk)`

if an entity is reused for a different record, an exception is thrown (and covered by unit test)

rickyosser:
Hi,

I tried to follow the instructions above for the changes and still get 
Screenshot from 2020-09-04 11-04-35

The error is pointing to the line with foreach.

I'm doing:
`foreach($m as $entity) {`
`   $name = $entity->get("name");`
`}`
Before this change the code kind of worked! :)
/Rickard

mvorisek:
@rickyosser `$m` was previously probably already loaded

PR #700, opened 2020-09-01
Revert join test fix after join/reference clone is fixed

mvorisek:
introduced/discovered in https://github.com/atk4/data/pull/697

I think, there is a bug in Join/Reference when model is cloned. I belive, that clone should be equivalent and should work exactly the same.


mvorisek:
@georgehristov this is the problematic usecase

georgehristov:
Let me look into it now

georgehristov:
I looked into `JoinSqlTest::testDoubleJoin` now. I did so before as well.
Honestly I believe the test is failing now but the results in DB are now correct and not correct before.

contact ID 100 - expected deleted but it should not be deleted
country ID 1 - expected deleted but it should not be deleted

@mvorisek What is your opinion?

georgehristov:
Join relies on its `$owner` to be loaded but you load a clone of the join model.
That is why it cannot determine the master field value.

georgehristov:
This is a quickfix but issue stems from the `Model::onHook` initialization done in `Join::init`. 
Hooks should be re-bound to the cloned model

mvorisek:
> Join relies on its `$owner` to be loaded but you load a clone of the join model.
> That is why it cannot determine the master field value.

that is correct, but what about bounded and later similarry cloned hooks? like in:
image

I belive clone of this must fail or at least produce unwanted result

mvorisek:
@georgehristov the fix does not fix after rebase onto develop/#711

mvorisek:
@georgehristov the fix does not fix after rebase onto develop/#711 , can you adjust it, so we can close this PR?

georgehristov:
I am working on this today

src/Model/JoinLinkTrait.php
```diff
@@ -0,0 +1,32 @@
...
+namespace atk4\data\Model;
+
+trait JoinLinkTrait
+{
+    /**
+     * The short name of the join link.
+     *
+     * @var string
+     */
+    protected $join;
```
mvorisek: rename to `joinName`

georgehristov:
So the conclusion now is: never reference model elements as objects (in properties, etc) but use the `short_name` instead. That is how you get the actual object associated with the cloned model.

src/Model/JoinLinkTrait.php
```diff
@@ -0,0 +1,32 @@
...
+
+    public function getJoin(): Join
+    {
+        return $this->getOwner()->getElement($this->join);
+    }
+
+    public function hasJoin(): bool
+    {
+        return $this->join && $this->getOwner()->hasElement($this->join);
```
mvorisek:
```suggestion
        return $this->join !== null;
```

getter should fail if name is set but Join does not exist!

mvorisek:
> 
> 
> So the conclusion now is: never reference model elements as objects (in properties, etc) but use the `short_name` instead. That is how you get the actual object associated with the cloned model.

yes, absolutely agree to reference using scalar value if added somewhere else reachable from the current object, that is also how hooks are designed now

mvorisek:
@georgehristov please also see `revert_after_join_clone_fixed__cherrypicked` vs. this PR branch - which one is correct? (the first one contains also the original fix)

src/Model/JoinLinkTrait.php
georgehristov: Renaming would be BC break

georgehristov:
> please also see revert_after_join_clone_fixed__cherrypicked vs. this PR branch - which one is correct? (the first one contains also the original fix)

Both branches are absolutely identical in end result code so we keep this one.

mvorisek:
> Both branches are absolutely identical in end result code so we keep this one.

image

@georgehristov they are not

georgehristov:
identical <> same :)

This PR we should use

mvorisek:
> identical <> same :)

yes, but the issue is/was in another word

code <> functionality :)

I see now what it should do, so yes, the "end functionality is equivalent"

FYI `revert_after_join_clone_fixed__cherrypicked` deleted

mvorisek:
@georgehristov tests pass, but I think the fix is not complete:

- https://github.com/atk4/data/blob/4e3b91d551ccc4131656fe7c732260625fb8685a/src/Reference/HasOne.php#L173 - in Field, we have name/string, so we can not assign an object!

- https://github.com/atk4/data/blob/4e3b91d551ccc4131656fe7c732260625fb8685a/src/Model/Join.php#L262 - we can not set `join` to any object - which object classes/ifaces does that method expects?

- https://github.com/atk4/data/blob/4e3b91d551ccc4131656fe7c732260625fb8685a/src/Model/Join.php#L333 - the seed is for `Reference\HasOne`, ie. we can not assign a string here

- all other `join` prop uses in atk4/data seems ok

src/Model/JoinLinkTrait.php
mvorisek: I renamed it

marking with BC break label, breaking anyway, as previously we assigned an object instead of string

georgehristov (approved):

src/Model/Join.php
```diff
@@ -250,22 +250,6 @@ public function addFields($fields = [])
         return $this;
     }
 
-    /**
-     * Adds any object to owner model.
-     */
-    public function add(object $object, array $defaults = []): object
```
mvorisek, line 262: @georgehristov it turned out, that this method is completely unused, so I removed it.

mvorisek:
@georgehristov, thanks!

PR #706, opened 2020-09-02
Fix PostgreSQL CI errors

### LINKED ISSUE #705: Fix PostgreSQL tests
a lot of tests are failing

mvorisek:
fixes https://github.com/atk4/data/issues/705

PostgreSQL SEQUENCE is not incremented if ID is explicitly set.
```
CREATE SEQUENCE seq;

create table target(
   id integer,
   name varchar(100),
   CONSTRAINT pk PRIMARY KEY (id)
);

INSERT INTO target (id, name) values (seq.nextval, 'test');
INSERT INTO target (id, name) values (3, 'test');
INSERT INTO target (id, name) values (seq.nextval, 'test');
INSERT INTO target (id, name) values (2, 'test');
INSERT INTO target (id, name) values (seq.nextval, 'test');
```

Using PostgreSQL SERIAL is NOT a solution, as it only a sugar for SEQUENCE.
```
create table target(
   id SERIAL PRIMARY KEY,
   name varchar(100)
);

INSERT INTO target (name) values ('test');
INSERT INTO target (id, name) values (3, 'test');
INSERT INTO target (name) values ('test');
INSERT INTO target (id, name) values (2, 'test');
INSERT INTO target (name) values ('test');
```

Oracle should support AI like:
```
create table target(
   id INTEGER GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1) NOT NULL,
   name varchar(100),
   CONSTRAINT pk PRIMARY KEY (id)
);

INSERT INTO target (name) values ('test');
INSERT INTO target (id, name) values (3, 'test');
INSERT INTO target (name) values ('test');
INSERT INTO target (id, name) values (2, 'test');
INSERT INTO target (name) values ('test');
```

but I am not sure if that has the same issue or not (I have no access to Oracle DB, related with https://github.com/atk4/dsql/issues/241 ).

### DIFF
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -783,9 +783,10 @@ class Sql extends Persistence
     {
         $insert = $model->action('insert');
 
-        // don't set id field at all if it's NULL
-        if ($model->id_field && array_key_exists($model->id_field, $data) && $data[$model->id_field] === null) {
+        if ($model->id_field && !isset($data[$model->id_field])) {
             unset($data[$model->id_field]);
+
+            $this->syncIdSequence($model);
         }
 
         $insert->set($this->typecastSaveRow($model, $data));
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -802,9 +802,11 @@ class Sql extends Persistence
                 ->addMoreInfo('scope', $model->scope()->toWords());
         }
 
+        $id = $model->persistence->lastInsertId($model);
+
         $model->hook(self::HOOK_AFTER_INSERT_QUERY, [$insert, $st]);
 
-        return $model->persistence->lastInsertId($model);
+        return $id;
     }
 
     /**
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -802,7 +802,11 @@ class Sql extends Persistence
                 ->addMoreInfo('scope', $model->scope()->toWords());
         }
 
-        $id = $model->persistence->lastInsertId($model);
+        if ($model->id_field && isset($data[$model->id_field])) {
+            $id = $data[$model->id_field];
+        } else {
+            $id = $model->persistence->lastInsertId($model);
+        }
 
         $model->hook(self::HOOK_AFTER_INSERT_QUERY, [$insert, $st]);
 
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -807,7 +807,7 @@ class Sql extends Persistence
 
             $this->syncIdSequence($model);
         } else {
-            $id = $model->persistence->lastInsertId($model);
+            $id = $this->lastInsertId($model);
         }
 
         $model->hook(self::HOOK_AFTER_INSERT_QUERY, [$insert, $st]);
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -977,9 +977,8 @@ class Sql extends Persistence
         // PostgreSQL sequence must be manually synchronized if a row with explicit ID was inserted
         if ($this->connection instanceof \atk4\dsql\Postgresql\Connection) {
             $this->connection->expr(
-                // @TODO soft-escape
-                'select setval(pg_get_serial_sequence([], []), coalesce(max(' . $model->id_field . '), 0) + 1, false) from "' . $model->table . '"',
-                [$model->table, $model->id_field]
+                'select setval([], coalesce(max({}), 0) + 1, false) from {}',
+                [$this->getIdSequenceName($model), $model->id_field, $model->table]
             )->execute();
         }
     }
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -776,10 +776,8 @@ class Sql extends Persistence
 
     /**
      * Inserts record in database and returns new record ID.
-     *
-     * @return mixed
      */
-    public function insert(Model $model, array $data)
+    public function insert(Model $model, array $data): string
     {
         $insert = $model->action('insert');
 
@@ -804,7 +802,7 @@ class Sql extends Persistence
         }
 
         if ($model->id_field && isset($data[$model->id_field])) {
-            $id = $data[$model->id_field];
+            $id = (string) $data[$model->id_field];
 
             $this->syncIdSequence($model);
         } else {
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -804,6 +804,8 @@ class Sql extends Persistence
 
         if ($model->id_field && isset($data[$model->id_field])) {
             $id = $data[$model->id_field];
+
+            $this->syncIdSequence($model);
         } else {
             $id = $model->persistence->lastInsertId($model);
         }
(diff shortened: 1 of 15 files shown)


mvorisek:
@georgehristov do you think we can fix the PostreSQL errors?

src/Persistence/Sql.php
```diff
@@ -945,21 +954,35 @@ public function getFieldSqlExpression(Field $field, Expression $expression)
...
+    {
+        return $this->connection->lastInsertId($this->getIdSequenceName($model));
+    }
+
+    protected function syncIdSequence(Model $model): void
+    {
+        // PostgreSQL sequence must be manually synchronized if a row with explicit ID was inserted
+        if ($this->connection instanceof \atk4\dsql\Postgresql\Connection) {
+            $this->connection->expr(
+                // @TODO soft-escape
```
mvorisek, line 982: todo before merge, please advise

src/Persistence/Sql.php
```diff
@@ -945,21 +952,34 @@ public function getFieldSqlExpression(Field $field, Expression $expression)
...
 
-        // PostgreSQL PDO always requires sequence name in lastInsertId method as parameter
-        // So let's use its default one if no specific is set
-        if ($this->connection instanceof \atk4\dsql\Postgresql\Connection && $seq === null) {
-            $seq = $model->table . '_' . $model->id_field . '_seq';
+        if ($sequenceName === null) {
+            // PostgreSQL uses sequence internally for PK autoincrement,
+            // use default name if not set explicitly
+            if ($this->connection instanceof \atk4\dsql\Postgresql\Connection) {
+                $sequenceName = $model->table . '_' . $model->id_field . '_seq';
```
georgehristov, line 963: I would suggest creating a method `Connection::getIdSequenceName` and then add the Postrgresql specific code there.

So this method body will be:
`return $model->sequence ?: $this->connection->getIdSequenceName();`

We should write implementation specific code in the specific class

src/Persistence/Sql.php
```diff
@@ -945,21 +952,34 @@ public function getFieldSqlExpression(Field $field, Expression $expression)
...
+
+    public function lastInsertId(Model $model): string
+    {
+        return $this->connection->lastInsertId($this->getIdSequenceName($model));
+    }
+
+    protected function syncIdSequence(Model $model): void
+    {
+        // PostgreSQL sequence must be manually synchronized if a row with explicit ID was inserted
+        if ($this->connection instanceof \atk4\dsql\Postgresql\Connection) {
```
georgehristov, line 978: Then this would also go to DSQL

src/Persistence/Sql.php
mvorisek: more testing with MSSQL and Oracle must be done to see "how much vendor specific" it is

mvorisek:
comments should be solved later in https://github.com/atk4/dsql/pull/243

src/Persistence/Sql.php
DarkSide666: As far as I know this default sequence name is only for PostgreSQL. Not sure about MSSQL, but for now I guess it's fine to leave it as it is now.

src/Persistence/Sql.php
```diff
@@ -776,16 +776,15 @@ public function loadAny(Model $model): array
 
     /**
      * Inserts record in database and returns new record ID.
-     *
-     * @return mixed
      */
-    public function insert(Model $model, array $data)
+    public function insert(Model $model, array $data): string
```
DarkSide666, line 780: Why return type is string? It can be integer too isn't it?

src/Persistence/Sql.php
```diff
@@ -945,21 +952,34 @@ public function getFieldSqlExpression(Field $field, Expression $expression)
...
+            if ($this->connection instanceof \atk4\dsql\Postgresql\Connection) {
+                $sequenceName = $model->table . '_' . $model->id_field . '_seq';
+            }
         }
 
-        return $this->connection->lastInsertId($seq);
+        return $sequenceName;
+    }
+
+    public function lastInsertId(Model $model): string
```
DarkSide666, line 970: Method description (comment) is lost :) and is it really always a string (not integer)?
Can't it also be `null` in case nothing is inserted yet?

src/Persistence/Sql.php
```diff
@@ -945,21 +952,34 @@ public function getFieldSqlExpression(Field $field, Expression $expression)
...
+        return $this->connection->lastInsertId($this->getIdSequenceName($model));
+    }
+
+    protected function syncIdSequence(Model $model): void
+    {
+        // PostgreSQL sequence must be manually synchronized if a row with explicit ID was inserted
+        if ($this->connection instanceof \atk4\dsql\Postgresql\Connection) {
+            $this->connection->expr(
+                'select setval([], coalesce(max({}), 0) + 1, false) from {}',
+                [$this->getIdSequenceName($model), $model->id_field, $model->table]
```
DarkSide666, line 981: This could be quite inefficient to select `max(id)` on each insert if table is very big. Isn't it?

Also not sure how this helps if called in case I already set id=999 manually while inserting.
For example, I have:
* in table records 1,2,3
* sequence currval is at 2
* then insert new record and also set specific id for it id=9;

Then if i'm not mistaken, it will also call this method and set sequence currval to `3+1=4` and later on will insert record with id=9. And at that point sequence currval is not consistent with inserted data (currval after insert should be 9).

src/Persistence/Sql.php
mvorisek: This was here prior this PR :), later, if we can pass tests when moved to only one vendor, fine

DarkSide666 (changes requested):

src/Persistence/Sql.php
mvorisek: it must throw if nothing was inserted yet

src/Persistence/Sql.php
mvorisek: really? ID SHOULD have always index (and ALWAYS implied if ID is PK (=unique index))

mvorisek: if you insert 9, then next ID must be 10, where it the issue? (this is how MySQL AI behave)

src/Persistence/Sql.php
DarkSide666: Is there a test for that?

src/Persistence/Sql.php
mvorisek: image

src/Persistence/Sql.php
DarkSide666: Yes next ID will be fine, but sequence.currval *after* inserting 9 but before inserting next record will be wrong = 4.
Sequence.currval right after inserting 9 should also switch to 9 isn't it?

src/Persistence/Sql.php
mvorisek: At least PostgreSQL will, this is quite vendor specific (some vendors may return 0), we should trust DSQL here as we use it here

DarkSide666: OK fine.

DarkSide666 (approved):
ok

PR #707, opened 2020-09-04
Fix Field::compare and Model::get()

mvorisek:


src/Model.php
```diff
@@ -806,24 +806,18 @@ public function get(string $field = null)
...
+                $data[$field] = $this->get($field);
             }
 
             return $data;
         }
 
         $this->checkOnlyFieldsField($field);
 
-        if (array_key_exists($field, $this->data)) {
+        if ($field === $this->id_field) {
```
mvorisek, line 818: I hit this several times - what is the purpose of `id` field instead of `data['id']`?

currently, they can very easily go out of sync as `id` is not magic property and `data` is public and used also multiple times in `Model` outside `get`/`set` functions

src/Field.php
```diff
@@ -489,7 +489,12 @@ public function setNull(): self
      */
     public function compare($value): bool
     {
-        return $this->get() == $value;
+        if ($this->owner->persistence === null) {
```
mvorisek, line 492: what do to with this if? It will fail if normalize returns an object which may not have `__toString()` implemented (like `\DateTime`)

src/Field.php
```diff
@@ -489,7 +489,12 @@ public function setNull(): self
...
+            return (string) $this->normalize($this->get()) === (string) $this->normalize($value);
+        }
+
+        return (string) $this->owner->persistence->typecastSaveRow($this->owner, [$this->short_name => $this->get()])[$this->short_name]
```
mvorisek, line 496: this will fail if `get()` does not use `->id` (see https://github.com/atk4/data/pull/707/files#r483735690 )

mvorisek:
solved in https://github.com/atk4/data/pull/709

PR #708, opened 2020-09-04
Drop ID property in advance of standard row data, provide BC using magic

mvorisek:
Labeling as a bug, as previously `get(id_field)` was very commonly have a different value than the actual id property.

Now, id property is implemented as magical one and data can never go out of sync.

Later, I think we should drop it in favor of getId() and setId(val) as we do not provide property access for other fields.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -117,7 +117,7 @@ However if you change the ID for record that was loaded, then your database
 record will also have its ID changed. Here is example::
 
     $m->load(123);
-    $m->set($m->id_field, 321);
+    $m->setId(321);
     $m->save();
 
 After this your database won't have a record with ID 123 anymore.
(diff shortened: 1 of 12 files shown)


georgehristov:
How about introducing getId and setId(val) with this PR?

mvorisek:
> 
> 
> How about introducing getId and setId(val) with this PR?

I target now for BC replacement of id prop with a magical one. It can be merged immediatelly :)

Then we may create another PR, the changes are completely unrelated except it is for the same domain - what will be gained when implemented in this PR?

mvorisek:
> 
> 
> How about introducing getId and setId(val) with this PR?

I target now for BC replacement of id prop with a magical one. It can be merged immediatelly :)

If you mean only to introduce these methods without dropping id property, then good idea, I will do so.

mvorisek:
@georgehristov there are two WIP commits - https://github.com/atk4/data/pull/708/commits

do you know how to fix them?

georgehristov (approved):
LGTM

PR #709, opened 2020-09-04
On-the-fly transform of ID in array persistance from key to result data

mvorisek:


src/Persistence/Csv.php
```diff
@@ -258,7 +258,7 @@ public function tryLoadAny(Model $model): ?array
         }
 
         $data = $this->typecastLoadRow($model, $data);
-        $data['id'] = $this->line;
+        $data[$model->id_field] = $this->line;
```
mvorisek, line 261: this was a bug I belive

georgehristov:
This is solved in #690 but I will merge this to it then

mvorisek:
@georgehristov any concerns left, can we merge it?

src/Persistence/Csv.php
PhilippGrashoff: Good catch!

mvorisek: @PhilippGrashoff can you approve this PR?

src/Field.php
```diff
@@ -489,7 +489,12 @@ public function setNull(): self
      */
     public function compare($value): bool
     {
-        return $this->get() == $value;
+        if ($this->owner->persistence === null) {
+            return (string) $this->normalize($this->get()) === (string) $this->normalize($value);
```
PhilippGrashoff, line 493: Why the explicit (string) typecast? Shouldn't normalize() return same data type if $value equals $this->get() anyway?

src/Model.php
```diff
@@ -806,16 +806,8 @@ public function get(string $field = null)
...
-                // collect data for actual fields
-                foreach ($this->only_fields as $field) {
-                    $data[$field] = $this->get($field);
-                }
-            } else {
-                // get all fields
-                foreach ($this->getFields() as $field => $f) {
-                    $data[$field] = $this->get($field);
-                }
+            foreach ($this->only_fields ?: array_keys($this->getFields()) as $field) {
```
PhilippGrashoff, line 809: Nice shortening ;)

src/Model.php
```diff
@@ -1220,7 +1212,7 @@ public function asModel(string $class, array $options = []): self
         $m = $this->newInstance($class, $options);
 
         foreach ($this->data as $field => $value) {
-            if ($value !== null && $value !== $this->getField($field)->default) {
+            if ($value !== null && $value !== $this->getField($field)->default && $field !== $this->id_field) {
```
PhilippGrashoff, line 1215: Why dont you want id_field copied? I think we need to define what asModel() does:
1) Should it work like this (I don't like the code):
```
$person = (new SomePerson($persistence))->load(44);
$admin = $person->asModel(Admin::class);
$admin->set('something', 'someValue'');
$admin->save(); //Will overwrite record with ID 44
```

2) Or should it create a copy of the record like:
```
$person = (new SomePerson($persistence))->load(44);
$admin = $person->asModel(Admin::class);
$admin->set('something', 'someValue'');
$admin->save(); //Will add new record to Persistence
```

Personally, I don't see much use for 1), so for me code change is ok. What do others think?

PhilippGrashoff (approved):
LGTM

src/Model.php
mvorisek: ID is normally not globally unique and it is assigned usually by on insert "by the target table", thus we do not copy ID by default

src/Field.php
mvorisek: normalize does not cast if type is not defined, might be worth later to update...

src/Model.php
PhilippGrashoff: Ah, I get your thought. I did not think about the use case that the models are saved into different tables. You are right, very sensible!

src/Persistence/Array_.php
```diff
@@ -37,9 +37,51 @@ public function __construct(array $data = [])
      *             - https://github.com/atk4/data/blob/90ab68ac063b8fc2c72dcd66115f1bd3f70a3a92/src/Reference/ContainsMany.php#L66
      *             remove once fixed/no longer needed
      */
-    public function getRawDataByTable(string $table): array
+    public function getRawDataByTable(Model $model, string $table): array
```
georgehristov, line 40: I would only remove second argument and use $model->table instead

src/Reference/ContainsMany.php
```diff
@@ -37,7 +37,7 @@ public function ref(array $defaults = []): Model
         // set some hooks for ref_model
         foreach ([Model::HOOK_AFTER_SAVE, Model::HOOK_AFTER_DELETE] as $spot) {
             $theirModel->onHook($spot, function ($theirModel) {
-                $rows = $theirModel->persistence->getRawDataByTable($this->table_alias);
+                $rows = $theirModel->persistence->getRawDataByTable($this->getTheirModel(), $this->table_alias);
```
georgehristov: We should use $theirModel directly

src/Reference/ContainsOne.php
```diff
@@ -97,7 +97,7 @@ public function ref(array $defaults = []): Model
         // set some hooks for ref_model
         foreach ([Model::HOOK_AFTER_SAVE, Model::HOOK_AFTER_DELETE] as $spot) {
             $theirModel->onHook($spot, function ($theirModel) {
-                $row = $theirModel->persistence->getRawDataByTable($this->table_alias);
+                $row = $theirModel->persistence->getRawDataByTable($this->getTheirModel(), $this->table_alias);
```
georgehristov: We should use $theirModel directly

georgehristov (approved):

src/Persistence/Array_.php
mvorisek: image

@georgehristov please see, the table names are different, `Reference::$table_alias` is different than `(Their)Model::$table`, this was one of the initial issue with this PR

src/Reference/ContainsOne.php
mvorisek: yes, we should, 👍

src/Reference/ContainsMany.php
mvorisek: yes, we should, 👍

PR #710, opened 2020-09-05
Normalize array persistence seed data

mvorisek:


src/Persistence/Array_.php
```diff
@@ -31,6 +29,52 @@ public function __construct(array $data = [])
...
+            foreach ($rows as $id => $row) {
+                $this->saveRow($model, $row, $id);
+            }
+        }
+
+        // for array persistence join which accept table directly (without model initialization)
+        foreach ($model->getFields() as $field) {
+            if ($field->hasJoin()) {
+                $join = $field->getJoin();
+                $joinTable = \Closure::bind(function () use ($join) {
```
DarkSide666, line 67: Why not simply `$joinTable = $join->foreign_table;` ?

src/Persistence/Array_.php
```diff
@@ -31,6 +29,52 @@ public function __construct(array $data = [])
...
+                    return $join->foreign_table;
+                }, null, Array_\Join::class)();
+                if (isset($this->seedData[$joinTable])) {
+                    $dummyJoinModel = new Model($this, ['table' => $joinTable]);
```
DarkSide666, line 71: Why this is needed?

DarkSide666 (approved):

src/Persistence/Array_.php
mvorisek: It is protected and I do not want to make it public because of this internal use.

src/Persistence/Array_.php
mvorisek: To seed join (unmodelled) data correctly.

PR #711, opened 2020-09-06
Fix not rebound hooks after clone

### LINKED ISSUE #212: looks like misspelled ->table
This line possibly accesses non-existent property.

https://github.com/atk4/data/blob/develop/src/Join.php#L328

mvorisek:
merge together with https://github.com/atk4/core/pull/272

BC break - if you clone object with hooks bounded to a different object, it throws correctly (but less strict now because of https://github.com/atk4/core/pull/275 PR)

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -347,7 +347,7 @@ before and just slightly modifying it::
             }
         }
 
-        function softDelete($m) {
+        function softDelete(Model $m) {
             if (!$m->loaded()) {
                 throw (new \atk4\core\Exception('Model must be loaded before soft-deleting'))->addMoreInfo('model', $m);
             }
@@ -428,7 +428,7 @@ inside your model are unique::
             $this->owner->onHook(Model::HOOK_BEFORE_SAVE, \Closure::fromCallable([$this, 'beforeSave']));
         }
 
-        function beforeSave($m)
+        function beforeSave(Model $m)
         {
             foreach ($this->fields as $field) {
                 if ($m->dirty[$field]) {
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -77,11 +77,11 @@ Another scenario which could benefit by type substitution would be::
 ATK Data allow class substitution during load and iteration by breaking "afterLoad"
 hook. Place the following inside Transaction::init()::
 
-    $this->onHook(Model::HOOK_AFTER_LOAD, function ($m) {
-        if (get_class($this) != $m->getClassName()) {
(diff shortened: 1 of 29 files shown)


docs/advanced.rst
```diff
@@ -652,23 +652,23 @@ Here is how to add them. First you need to create fields::
 I have declared those fields with never_persist so they will never be used by
 persistence layer to load or save anything. Next I need a beforeSave handler::
 
-    $this->onHook(Model::HOOK_BEFORE_SAVE, function($m) {
-        if($m->_isset($m['client_code') && !$m->_isset($m['client_id')) {
+    $this->onHookShort(Model::HOOK_BEFORE_SAVE, function() {
+        if($this->_isset($this['client_code') && !$this->_isset($this['client_id')) {
```
georgehristov: Should be `$this->_isset('client_code')` I believe

tests/Model/Smbo/Transfer.php
```diff
@@ -23,12 +23,12 @@ protected function init(): void
 
         $this->addField('destination_account_id', ['never_persist' => true]);
 
-        $this->onHook(self::HOOK_BEFORE_SAVE, function ($m) {
+        $this->onHookShort(self::HOOK_BEFORE_SAVE, function () {
             // only for new records and when destination_account_id is set
-            if ($m->get('destination_account_id') && !$m->id) {
+            if ($this->get('destination_account_id') && !$this->id) {
```
georgehristov: Migrate out of $model->id?

mvorisek: wdym?

docs/advanced.rst
mvorisek: yes, old typo, fixed, thanks

georgehristov (commented):
Generally LGTM. I am a bit worried that it gets complicated...
Does it completely solve the issue of hook binding on cloning?

src/Field.php
```diff
@@ -223,6 +223,21 @@ public function __construct(array $defaults = [])
...
     }
 
+    protected function onHookToOwner(string $spot, \Closure $fx, array $args = [], int $priority = 5): int
+    {
+        $name = $this->short_name;
+
+        return $this->owner->onHookDynamic(
+            $spot,
+            static function (Model $owner) use ($name) {
+                return $owner->getField($name);
```
georgehristov, line 233: Why not `return $owner->getField($this->short_name);`

mvorisek: because we will prevent (at least) old `$this` to be GCed, goiing to sleep today ...zzz

DarkSide666 (approved):
Well... OK :)

PR #716, opened 2020-09-08
Drop magic ID property from Model

mvorisek:
## how to migrate ##

fix all your assigments to `id` property:
- from `$model->id = $id;` to `$model->setId($id);`

and all `id` property reads:
- from `$id = $model->id;` to `$model->getId();`


mvorisek:
please advise what to do with this doc:

image

otherwise RTM

georgehristov (approved):
LGTM

georgehristov (approved):

PR #717, opened 2020-09-08
Drop $from_persistence support for Model::load()

mvorisek:
this is some weird hack, no tests, no usage in data, only troubles


georgehristov (approved):
Agree

PR #720, opened 2020-09-10
Stricter Condition validation, remove legacy OR condition support

mvorisek:
buggy now, good impulse to accept more strict input only

### DIFF
--- a/src/Model/Scope/Condition.php
+++ b/src/Model/Scope/Condition.php
@@ -299,11 +299,10 @@ class Condition extends AbstractScope
 
                 if ($field === '#') {
                     $words[] = $this->operator ? 'number of records' : 'any referenced record exists';
-                    $field = null;
                 }
             }
 
-            if ($field !== null) {
+            if ($model->hasField($field)) {
                 $field = $model->getField($field);
             }
         }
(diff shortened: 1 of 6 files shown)


src/Model/Scope/Condition.php
```diff
@@ -200,8 +209,8 @@ public function toQueryArguments(): array
                         } else {
                             $refModel->addCondition($field, $operator, $value);
                             $field = $refModel->action('exists');
-                            $operator = null;
-                            $value = null;
+                            $operator = '>';
```
mvorisek, line 212: needed for MSSQL

tests/ScopeTest.php
```diff
@@ -173,7 +173,7 @@ public function testContitionToWords()
 
         $country = clone $this->country;
 
-        $country->addCondition('Users/#');
+        $country->addCondition('Users/#', '>', 0);
```
mvorisek, line 176: needed for MSSQL, see https://dba.stackexchange.com/questions/275313/can-an-expression-be-used-as-an-expression-as-well-as-a-predicate

mvorisek:
@georgehristov please fix the toWords

georgehristov:
Shall we drop support for legacy OR so fast?

georgehristov (approved):
LGTM

PR #721, opened 2020-09-12
[fix] Condition key can be Expressionable

georgehristov:
We should make Expression implement Expressionable but meanwhile this fix is needed.

### DIFF
--- a/src/Model/Scope/Condition.php
+++ b/src/Model/Scope/Condition.php
@@ -104,7 +104,7 @@ class Condition extends AbstractScope
             throw new Exception('Only Scope can contain another conditions');
         } elseif ($key instanceof Field) { // for BC
             $key = $key->short_name;
-        } elseif (!is_string($key) && !($key instanceof Expression)) {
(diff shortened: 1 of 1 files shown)


src/Model/Scope/Condition.php
```diff
@@ -104,7 +104,7 @@ public function __construct($key, $operator = null, $value = null)
             throw new Exception('Only Scope can contain another conditions');
         } elseif ($key instanceof Field) { // for BC
             $key = $key->short_name;
-        } elseif (!is_string($key) && !($key instanceof Expression)) {
+        } elseif (!is_string($key) && !($key instanceof Expression) && !($key instanceof Expressionable)) {
```
mvorisek, line 107: probably yes, but why `Expression` does not implement `Expressionable` (and does not even provide that one prototyped method)?

where is `Expressionable` used in Scopes?

georgehristov: $key can be Expressionable.
Expressionable is any object that can be converted to expression using getSqlExpression method

mvorisek (approved):
seems 👍

PR #722, opened 2020-09-12
Persistence\Sql::getFieldSqlExpression return expression using field owner expr method

georgehristov:


mvorisek (approved):
somethings smells badly in the design, but typo seems correct

PR #723, opened 2020-09-13
Add Oracle DB support

mvorisek:
merge after https://github.com/atk4/dsql/pull/243 and https://github.com/atk4/schema/pull/82

### DIFF
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -835,12 +835,12 @@ class Sql extends Persistence
     /**
      * Prepare iterator.
      */
-    public function prepareIterator(Model $model): \PDOStatement
+    public function prepareIterator(Model $model): iterable
     {
         try {
             $export = $model->action('select');
 
-            return $export->execute();
+            return $export->getIterator();
         } catch (\PDOException $e) {
             throw (new Exception('Unable to execute iteration query', 0, $e))
                 ->addMoreInfo('query', $export->getDebugQuery())
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -592,11 +592,11 @@ class Sql extends Persistence
                 $this->initQueryConditions($model, $query);
                 $model->hook(self::HOOK_INIT_SELECT_QUERY, [$query, $type]);
 
-                // MSSQL does not support EXISTS everywhere, so wrap in SELECT
-                if ($this->connection instanceof \atk4\dsql\Mssql\Connection) {
-                    return $this->dsql()->expr(
-                        '(select case when exists[] then 1 else 0 end)',
-                        [$query]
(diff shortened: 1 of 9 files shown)


src/Persistence/Sql.php
```diff
@@ -592,11 +592,11 @@ public function action(Model $model, string $type, array $args = [])
                 $this->initQueryConditions($model, $query);
                 $model->hook(self::HOOK_INIT_SELECT_QUERY, [$query, $type]);
 
-                // MSSQL does not support EXISTS everywhere, so wrap in SELECT
-                if ($this->connection instanceof \atk4\dsql\Mssql\Connection) {
-                    return $this->dsql()->expr(
-                        '(select case when exists[] then 1 else 0 end)',
-                        [$query]
+                // MSSQL and Oracle do not support EXISTS everywhere, so wrap in CASE
+                if ($this->connection instanceof \atk4\dsql\Mssql\Connection
```
georgehristov: We should move this to Connection specific class code, e.g
```
$this->dsql()->exists($query);
```
And we have specific `\atk4\dsql\Oracle\Expression::exists` method

src/Persistence/Sql.php
```diff
@@ -971,6 +971,19 @@ private function getIdSequenceName(Model $model): ?string
 
     public function lastInsertId(Model $model): string
     {
+        // TODO: Oracle does not support lastInsertId(), only for testing
+        // as this does not support concurrent inserts
+        if ($this->connection instanceof \atk4\dsql\Oracle\Connection) {
```
georgehristov, line 968: Same here - we should move this to sql Connection/Expression classes

georgehristov (approved):

src/Persistence/Sql.php
mvorisek: In current design, Expression is not too much too be extended :( as used a lot directly and already further extended.

however, yes, it should be moved, related probably with #690

src/Persistence/Sql.php
mvorisek: this can be moved easier, but we will have to still provide a lot of data about the Model (but not Model, as we are in dsql there)

georgehristov:
I know and I was thinking about it.
How about arguments $table, $idField, $defaultSequenceName?

src/Persistence/Sql.php
georgehristov: I implemented it similar way in #690 for now but added a @todo for moving it

mvorisek:
> 
> 
> I know and I was thinking about it.
> How about arguments $table, $idField, $defaultSequenceName?

first 2 parameters are absolutely needed thus legit

the 3rd one - probably `$sequenceName` or completely move to dsql if possible

PR #724, opened 2020-09-14
Rollback because dirty should be available in after save hook callback.

DarkSide666:


mvorisek (changes requested):
no rollback, position is good, or explain deeply, why the original hook position is better

I think, we should not call hook in "middle of some low-level operation"

DarkSide666:
Agree

PR #726, opened 2020-09-15
Revert to official Docker image

mvorisek:
merge after https://github.com/atk4/data/pull/723 and https://github.com/atk4/image/issues/6


DarkSide666 (approved):
ok

PR #727, opened 2020-09-16
Use DBAL Type for Field type

mvorisek:
If a custom typecasting is needed, easy, simply register own type in DBAL. If a library wants to provide some custom types by default, bootstrap (register on load time) using composer.

I think this is step in a good direction, as we can then leave all typecasting to standard DBAL Types and remove all hacky typing from the code. DBAL Types can also be easily replaced and/or a new/custom Type can be easily added.

- yes, this will solve all typecasting across different Persistences
- this PR is strictly focused on typecasting a managed value (php value) to an unmanaged value (string, scalar in general)
- this PR is BC as much as possible, ie. it keep working without explicit Persistence nor Platform

## BC breaks:

- removed `Field::serialize` property, to use PHP serialization, use `object` type.
- removed `Field::typecast` property, to use a custom typecasting, extend the target DBAL type class and register it or implement a custom persistence
- removed `Field::dateTimeClass/dateTimeZoneClass` properties, see bullet point above
- ~removed `Field::persist_format/persist_timezone` properties, see above, but you probably want to handle it using persistence/UI solely - if there was a shared format/TZ, different format/TZ was not possible to set for the database and UI~
- removed `Model::strict_types` property, was already enabled by default, normalization can still be disabled/broken by `Model::HOOK_NORMALIZE` hook
- `array` type is temporary forbidden to be used, `object` type now serializes data using PHP serialization, to use JSON, use `json` type. For more details, see the following table:

| Serialization purpose | Original Atk4 3.0 (or lower) type  | New/DBAL type equivalent | Notes                                                                |
|-----------------------|------------------------------------|--------------------------|----------------------------------------------------------------------|
| Encode array as JSON  | `array`                            | `json`                   | `array` type now behaves differently, thus temporary forbidden to be used in atk4/data       |
| Encode object as JSON | `object`                           | `json`                   | https://github.com/doctrine/dbal/blob/3.1.2/src/Types/JsonType.php   |
| Serialize php data    | `Field::serialize` flag was needed | `object`                 | https://github.com/doctrine/dbal/blob/3.1.2/src/Types/ObjectType.php |
| Serialize php data    | `Field::serialize` flag was needed | `array`                 | forbidden, see the first table row and use `object` type instead |


src/Field.php
```diff
@@ -497,6 +507,42 @@ public function setNull(): self
         return $this;
     }
 
+    private function getDatabasePlatform(): Platforms\AbstractPlatform
+    {
+        return $this->getOwner()->persistence !== null
```
mvorisek: Should we should typecast with one generic platform - let say Sqlite?

All results will be then same and we can implemented 2 static methods for typecasting back and from like `Field::convertToUnmanagedValue($value)` and `Field::convertToPhpValue($value)`. Which can be easily used also for defining Scope values etc.

Or take care of this inside dsql Expression?

georgehristov: We have to keep the persistence agnostic nature of ATK.
My idea is that ATK specific types are introduced and when typecasting is done field specific or persistence specific then persistence maps this type to its own storage mechanism.

For instance we have ATK `Type\Array`. For Array persistence there will be no typecasting but for Sql for instance it will be typecasted to `Type\Text`.
So for this field the Sql persistence knows it must store a Type\Text

Still rough - working on the concept

mvorisek: `Type\Array` should typecast to string (json, serialize, ...) and it can work everywhere. I think, it should always typecast to unmanaged type, ie. objects should be newly recreated - if you do not like it, you can keep them when typecasting and relink back later ;-).

src/Field.php
```diff
@@ -265,6 +268,13 @@ public function normalize($value)
 
             $f = $this;
 
+//            // POC only - we still need to solve unsigned, not null, see how Doctrine handle Type options
+//            // then all code below should not be needed
+//            $platform = new \Doctrine\DBAL\Platforms\MySqlPlatform(); // should be provided by a Persistence (and DSQL Connection)
+            $type = $this->getTypeObject();
```
mvorisek, line 274: enforce Type presence

src/Field.php
```diff
@@ -497,6 +500,42 @@ public function setNull(): self
...
+    {
+        return $this->getOwner()->persistence !== null
+            ? $this->getOwner()->persistence->getDatabasePlatform()
+            : new Persistence\NosqlPlatform();
+    }
+
+    /**
+     * @param mixed|void $value
+     */
+    public function getUnmanagedValue($value = null): ?string
```
mvorisek: @georgehristov or `getDatabaseValue` to be consistent with DBAL naming?

unmanaged is used more in C, but "database" imply SQL persistence to me...

bootstrap_atk_money_type.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace atk4\data;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Types\Types;
+
+// TODO for types to DBAL migration, remove later, there should be no type "money"
+
+class bootstrap_atk_money_type extends Type
```
mvorisek, line 13: all tests pass with this super simple replacement with `decimal` native DBAL Type

@DarkSide666 @georgehristov should we completely remove `money` (database) type now in favor of native `decimal`?

src/Field.php
georgehristov: For me the naming in order of preference
1. getPersistenceValue
2. getDatabaseValue
...

src/Reference/HasOneSql.php
```diff
@@ -71,7 +71,7 @@ function (Model $ourModel) use ($theirFieldName) {
                 $theirModel = $this->getTheirModel();
 
                 $theirModel->addCondition($theirFieldName, $ourModel->get($ourFieldName));
-                $ourModel->set($this->getOurFieldName(), $theirModel->action('field', [$theirModel->id_field]));
+                $ourModel->set($this->getOurFieldName(), $theirModel->action('field', [$theirModel->id_field])->getOne());
```
mvorisek: @DarkSide666 @georgehristov a bug?

src/Reference/HasOneSql.php
```diff
@@ -212,7 +212,7 @@ function (Model $ourModel) {
                 $theirModel = $this->getTheirModel();
 
                 $theirModel->addCondition($theirModel->title_field, $ourModel->get($fieldName));
-                $ourModel->set($this->getOurFieldName(), $theirModel->action('field', [$theirModel->id_field]));
+                $ourModel->set($this->getOurFieldName(), $theirModel->action('field', [$theirModel->id_field])->getOne());
```
mvorisek, line 215: same here

src/Reference/HasOneSql.php
georgehristov: It is not a bug as it includes the SQL statement

georgehristov (commented):
.

src/Reference/HasOneSql.php
mvorisek: `$this->getOurFieldName()` will always point to a `FieldSqlExpression` Field?

if yes, `FieldSql` should still be RO (read only) and it seems it does not/should not support set at all

I think, my fix is wrong as it is missing typecast, but also the orignal code is wrong because of the above (and failing tests).

src/Field.php
mvorisek: 👍 for `getPersistenceValue`

mvorisek: I think I have choosen `unmanaged`/`maanged` because we need two names, eg. for both directions.

If we go with `getPersistenceValue`, `getPHPValue` as named by Doctrine https://github.com/doctrine/dbal/blob/3.1.2/src/Types/Type.php#L87 will not be ideal as persistence can be implemented in PHP as well.

src/Field.php
```diff
@@ -485,6 +321,61 @@ public function setNull(): self
         return $this;
     }
 
+    public /* TODO change before merge private*/ function getDatabasePlatform(): Platforms\AbstractPlatform
```
mvorisek, line 324: TODO

DarkSide666 (approved):
Pre-approving, but would be nice to have `array` type too...

src/Field.php
mvorisek: I dropped these new methods completely, just use persistence with the original names.

src/Field.php
mvorisek: I dropped these new methods completely, just use persistence with the original names.

PR #728, opened 2020-09-16
Introduce Field::getPersistenceName method

georgehristov:


src/Field.php
```diff
@@ -522,8 +522,18 @@ public function compare($value, $value2 = null): bool
...
+        $actualName = $this->getActualName();
+
+        return (string) $this->owner->persistence->typecastSaveRow($this->owner, [$this->short_name => $value])[$actualName]
+                === (string) $this->owner->persistence->typecastSaveRow($this->owner, [$this->short_name => $value2])[$actualName];
+    }
+
+    /**
+     * The actual name that is used in persistence.
+     */
+    public function getActualName(): string
```
mvorisek: rename, then we do not need even that dummy comment

src/Persistence.php
```diff
@@ -137,27 +137,27 @@ public function atomic(\Closure $fx)
...
-                $result[$key] = $value;
+            if (!$model->hasField($fieldName)) {
+                $result[$fieldName] = $value;
 
                 continue;
             }
 
             // Look up field object
-            $f = $m->getField($key);
+            $field = $model->getField($fieldName);
```
mvorisek, line 153: yes, `field` is object, `fieldName` is string, would be great to fix in Model etc.

mvorisek (approved):
preaprove

src/Persistence.php
```diff
@@ -137,27 +137,27 @@ public function atomic(\Closure $fx)
...
 
             // Figure out the name of the destination field
-            $field = isset($f->actual) && $f->actual ? $f->actual : $key;
+            $fieldActualName = $field->getActualName();
```
mvorisek: I think we can afford to get the actual name twice, getters are quite cheap

src/Persistence.php
```diff
@@ -196,40 +196,40 @@ public function typecastSaveRow(Model $m, array $row): array
      * may be "aliased" from SQL persistences or mapped depending on persistence
      * driver.
      */
-    public function typecastLoadRow(Model $m, array $row): array
+    public function typecastLoadRow(Model $model, array $row): array
     {
         $result = [];
-        foreach ($row as $key => $value) {
+        foreach ($row as $fieldName => $value) {
```
mvorisek, line 199: is this correct?

I think we get persistenceName and save fieldName

georgehristov: It is correct currently but in my opinion should be as you said.
Persistence returns the loaded data already with keys using short_name (implementing aliasing for Sql)
It is mostly useful in case field is an Sql expression.
I am thinking how to best modify it to be consistent and intuitive ...

mvorisek: see how `typecastSaveRow` method is implemented

georgehristov: I see. It should be same with load.
But internal workings of Sql persistence need to be changed a bit.
I will come up with a solution.
Add commit here or keep this PR only for introducing the method?

mvorisek: both ways are fine with me, if you think, it is breaking a lot external apps, then 2nd PR

PR #734, opened 2020-09-19
Add DBAL platform getter to Persistence

mvorisek:
merge after https://github.com/atk4/dsql/pull/254

for https://github.com/atk4/data/pull/727

@georgehristov please review and merge if ok, RTM from my side, maybe we want to add a test

### DIFF
--- a/src/Persistence/NosqlPlatform.php
+++ b/src/Persistence/NosqlPlatform.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace atk4\data\Persistence;
+
+use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Platforms;
+
+class NosqlPlatform extends Platforms\AbstractPlatform
+{
+    private function createNotSupportedException(): DBALException
+    {
+        return DBALException::notSupported('SQL');
+    }
+
+    public function getName(): string
+    {
+        return 'nosql';
+    }
+
+    protected function initializeDoctrineTypeMappings(): void
+    {
+    }
+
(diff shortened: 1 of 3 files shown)


src/Persistence/NosqlPlatform.php
```diff
@@ -0,0 +1,60 @@
...
+    public function getName(): string
+    {
+        return 'nosql';
+    }
+
+    protected function initializeDoctrineTypeMappings(): void
+    {
+    }
+
+    protected function _getCommonIntegerTypeDeclarationSQL(mixed $columnDef): string
```
mvorisek: implemented abstract methods...

georgehristov:
I will comment also on the Types draft but we should keep the persistence agnostic nature of atk4/data.
If we want to implement DBAL we should do so only in Sql persistence (and any other that DBAL can handle) but not in base persistence code.
Also for data types - we will introduce atk4/data types which in case of Persistence/Sql will be mapped to DBAL types (or introduce Persistence/Doctrine?).
This way we keep open to introduce other persistences (especially usable over APIs e.g GraphQL) where we can map the atk4 data types to respective persistence types.

mvorisek:
@georgehristov I understand what you mean, DBAL may seem for mapping between managed value and SQL.

This is true. However, I think we should view the DBAL Type differently.

We should view the Type conversion methods like "from managed value to unmanaged/scalar one" (and back).

This underlay that the Type is normally used for exactly this in Doctrine ORM.

Viewing Type with this reasoning, it is justified for all Persistences - we always need to typecast from managed to unmanaged value. And to use Type from DBAL, we need Platform, for non-SQL Persistence classes, I implemented no SQL one.

mvorisek:
merge in types PR if needed

PR #735, opened 2020-09-20
Add smoke and burn CI testing

mvorisek:
smote test: make CI testing fast again
burn test: detect memory leaks
speed trap: detect slow tests

detected memory leaking issue seems to be a bug in php: https://bugs.php.net/bug.php?id=80125


mvorisek:
@georgehristov one PostgreSQL test is super slow:
```
You should really speed up these slow tests (>500ms)...
 1. 6300ms to run atk4\data\tests\ScopeTest:testConditionOnReferencedRecords
```
do you know how to fix the SQL to make it fast?

DarkSide666 (approved):
Nice

PR #736, opened 2020-09-22
Fix scheduled/200 iterations burn test

mvorisek:


DarkSide666 (approved):
No idea why this line can fix memory leaking, but OK if that helps :)

mvorisek:
mysql test still failing

PR #737, opened 2020-09-29
Fix GC for CallbackFilterIterator using WeakReference

mvorisek:
we can fix cleanly, not sure when it get fixed in php itself

burn testing issue is related with:
https://bugs.php.net/bug.php?id=33487
and it needs to be solved like this:
https://3v4l.org/Kfj7D
to preresize class data structures

or by simply calling `gc_collect_cycles()` after every burn iteration (but it slows burn testing)

I belive, this php issue is also related with this fix: https://github.com/atk4/data/pull/736/files#diff-d1cd5dea60b08030dd9ea65704f84d0eR119


DarkSide666 (approved):
LGTM

PR #739, opened 2020-10-02
Fix Array persistence with Datetime fields

PhilippGrashoff:
Currently, the added test fails and produces this error message:

```
Error : Object of class DateTime could not be converted to string
 /var/www/html/atk/data/src/Field.php:539
 /var/www/html/atk/data/src/Field.php:542
 /var/www/html/atk/data/src/Model.php:710
 /var/www/html/atk/data/tests/PersistentArrayTest.php:212
 
```

The objective of this PR is to fix this issue. As no fix is provided yet, just a draft.

### DIFF
--- a/tests/PersistentArrayTest.php
+++ b/tests/PersistentArrayTest.php
@@ -232,6 +232,52 @@ class PersistentArrayTest extends AtkPhpunit\TestCase
         $this->expectException(Exception::class);
         $m->tryLoad(1);
     }
+    
+    /**
+     * different to other field types, date, time and datetime fields return an
+     * object, not a scalar value to work with. Make sure Array persistence handles them correctly.
+     */
+    public function testDateTimeFields() {
+        $p = new Persistence\Array_();
+        $m = new Model($p);
+        $m->addField('date', ['type' => 'date']);
+        $m->addField('time', ['type' => 'time']);
+        $m->addField('datetime', ['type' => 'datetime']);
+        $m->save();
+
+        //try setting DateTime Object
+        $dateTime = \DateTime::createFromFormat('Y-m-d H:i:s', '2020-01-01 11:11:11');
+        $m->set('date', $dateTime);
+        $m->set('time', $dateTime);
+        $m->set('datetime', $dateTime);
+        $m->save();
+
+        self::assertInstanceOf(
+            \DateTime::class,
+            $m->get('date')
+        );
+        self::assertInstanceOf(
+            \DateTime::class,
+            $m->get('time')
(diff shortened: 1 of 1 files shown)


tests/PersistentArrayTest.php
```diff
@@ -232,6 +232,52 @@ public function testTryLoadNotSupportedException()
...
+    public function testDateTimeFields() {
+        $p = new Persistence\Array_();
+        $m = new Model($p);
+        $m->addField('date', ['type' => 'date']);
+        $m->addField('time', ['type' => 'time']);
+        $m->addField('datetime', ['type' => 'datetime']);
+        $m->save();
+
+        //try setting DateTime Object
+        $dateTime = \DateTime::createFromFormat('Y-m-d H:i:s', '2020-01-01 11:11:11');
```
mvorisek, line 249: `new \DateTime('2020-01-01 11:11:11')` is enough

tests/PersistentArrayTest.php
```diff
@@ -232,6 +232,52 @@ public function testTryLoadNotSupportedException()
...
+        $m->set('date', $dateTime);
+        $m->set('time', $dateTime);
+        $m->set('datetime', $dateTime);
+        $m->save();
+
+        self::assertInstanceOf(
```
mvorisek, line 255: keep on 1 line, also, probably use `$this->assert` to be consistent

PhilippGrashoff: Does StyleCI command this? I find assertions way more readable this way.

mvorisek: 1 line - keep files short - important in general

mvorisek:
valid usecase - related with https://github.com/atk4/data/pull/727

we should typecase inside Field to an unmanaged object/string and twice (to/from) for normalize

probably also related with #690

UPDATE: #690 does not solve it, array persistence simply does not care about typecasting, which is wrong, as object can not be compared at all (in sense of non strict equality) and typecasting/normalize thru field Type seems to be the only correct approach

mvorisek (changes requested):
We should aim to deduplicate tests and tests things like typecasting for every Persistence from one file.

PhilippGrashoff:
Yes, this sounds reasonable. In general I am not happy adding another typecasting to this persistence. You and Georgi already put a lot of thought in this topic, and as far as I understood you have some good idea how to fix typecasting issues on a different level.

mvorisek:
Is this still valid with https://github.com/atk4/data/pull/804 ?

PR #747, opened 2020-10-14
Drop driverType property

mvorisek:
for https://github.com/atk4/dsql/pull/261

BC break:
(for unknown reasons, there was) removed `Persistence::driverType` property


DarkSide666 (approved):
LGTM

PR #748, opened 2020-10-14
Integrate atk4/schema into atk4/data

mvorisek:
then archive `atk4/schema` repo with comment it has been integrated here - https://github.com/atk4/schema/pull/88

later, we should drop the custom Migration code in favor of much, much better DBAL

full credit to original owners :)


DarkSide666 (approved):
LGTM

PR #749, opened 2020-10-16
remove deprecated code for dec-2020

georgehristov:
Clean up of deprecated classes and BC code


mvorisek (approved):
@georgehristov very good!

PR #751, opened 2020-10-16
Simplify Join persistence getter

mvorisek:
review after https://github.com/atk4/data/pull/750

maybe we do not want to use `$this->persistence` everywhere, would be great if there will be some tests for cross persistence


mvorisek:
not optimal...

PR #753, opened 2020-10-17
Replace original atk4/schema with DBAL SchemaManager

mvorisek:
merge after https://github.com/atk4/dsql/pull/262

DBAL MSSQL issue: https://github.com/doctrine/dbal/pull/4284
and patch for it: https://github.com/doctrine/dbal/pull/4360


src-schema/Migration.php
```diff
@@ -11,745 +11,209 @@
...
-            $this->dropField($field);
-            ++$dropped;
-            ++$changes;
+        try {
+            $this->getSchemaManager()->dropTable($this->getDatabasePlatform()->quoteSingleIdentifier($this->table->getName()));
+        } catch (\Doctrine\DBAL\DBALException $e) {
+        }
+
+        if ($this->getDatabasePlatform() instanceof OraclePlatform) {
+            // drop trigger if exists
```
DarkSide666, line 115: Probably should first drop trigger and only then drop table.

tests-schema/BasicTest.php
```diff
@@ -36,15 +25,8 @@ public function testCreateAndAlter()
             ->field('tm', ['type' => 'time'])
             ->field('dt', ['type' => 'date'])
             ->field('dttm', ['type' => 'datetime'])
-            ->field('dbl', ['type' => 'double'])
```
DarkSide666, line 39: don't we support double, money, enum field types anymore?

tests-schema/ModelTest.php
```diff
@@ -134,7 +127,7 @@ protected function init(): void
         parent::init();
 
         $this->addField('name');
-        $this->addField('password', ['type' => 'password']);
```
DarkSide666, line 137: don't we support field type "password" anymore?

tests/ContainsManyTest.php
```diff
@@ -266,6 +267,10 @@ public function testNestedContainsMany()
         // do we also correctly calculate discounts from nested containsMany?
         $this->assertSame(24.2 * 15 / 100 + 86.25 * 20 / 100, $i->get('discounts_total_sum')); // =20.88
 
+        if ($this->getDatabasePlatform() instanceof OraclePlatform) {
+            $this->markTestIncomplete('TODO - fix CLOB support on Oracle');
```
DarkSide666, line 271: isn't this fixed in other place where you had get_stream call?

DarkSide666 (approved):
LGTM, please check comments

tests-schema/BasicTest.php
mvorisek: instead of `double` use `float`, like in PHP, DBAL stores both as double (when not overrided), see https://github.com/doctrine/dbal/blob/b5e15999e8f626b9a50edef142fffc7be57d0211/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L932

src-schema/Migration.php
mvorisek: fixed

tests/ContainsManyTest.php
mvorisek: CLOB - not working, I did fixed stream to string transform in dsql, but still need more work

tests-schema/ModelTest.php
mvorisek: I commented it out, so we can get to it later once we switch to DBAL types

PR #762, opened 2020-10-24
fix phpdoc of _hasReference

PhilippGrashoff:
As $className is passed directly to fromSeed(), it either has to be an object or an array.


DarkSide666 (approved):

mvorisek:
reverted as invalid...

PR #763, opened 2020-10-24
Fix never thrown PDOException

mvorisek:
merge after https://github.com/atk4/dsql/pull/266


src/Persistence/Sql.php
```diff
@@ -689,7 +690,7 @@ public function tryLoad(Model $model, $id): ?array
                 return null;
             }
             $data = $this->typecastLoadRow($model, $dataRaw);
-        } catch (\PDOException $e) {
+        } catch (DsqlException $e) {
```
mvorisek, line 693: original `PDOException` is never thrown, as always wrapped by DSQL

georgehristov (approved):
LGTM

PR #764, opened 2020-10-26
Revert "fix phpdoc of _hasReference"

mvorisek:
Reverts atk4/data#762 + some new fixes

ping @PhilippGrashoff

I think we can fix there: `array *\$defaults(?! = \[)` (in Model constructor, the `$defaults` with string should probably stay for now)

### DIFF
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -305,12 +305,11 @@ class Join
     /**
      * Creates reference based on a field from the join.
      *
-     * @param string $link
-     * @param array  $defaults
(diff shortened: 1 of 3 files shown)


PhilippGrashoff (approved):
LGTM! Unfortunately I am somehow not a member of this repo so my review does not count in I think :)

mvorisek:
@PhilippGrashoff 99 PRs and we can talk about your official badge :D

georgehristov (approved):

PR #765, opened 2020-10-29
Fix typo

mvorisek:


.github/workflows/unit-tests.yml
```diff
@@ -129,7 +129,7 @@ jobs:
 
       - name: Install PHP dependencies
         run: |
-          if [[ "${{ matrix.type }}" != Phpunit* ]; then composer remove --no-interaction --no-update phpunit/phpunit johnkary/phpunit-speedtrap phpunit/dbunit phpunit/phpcov --dev ; fi
+          if [[ "${{ matrix.type }}" != Phpunit* ]]; then composer remove --no-interaction --no-update phpunit/phpunit johnkary/phpunit-speedtrap phpunit/dbunit phpunit/phpcov --dev ; fi
```
georgehristov: Is the typo not the `[[`?

mvorisek: no, because of `*`, bacause of bash...

georgehristov (approved):
Any idea why tests are failing?

.github/workflows/unit-tests.yml
mvorisek: failing because GH probably uses sh instead of bash, debugging it....

PR #766, opened 2020-10-29
Fix default string/text case sensitivity in Migrator

mvorisek:


src/Model.php
```diff
@@ -1676,9 +1676,11 @@ public function insert(array $row)
      */
     public function import(array $rows)
     {
-        foreach ($rows as $row) {
-            $this->insert($row);
-        }
+        $this->atomic(function () use ($rows) {
```
mvorisek: now import is now only safer, but much faster ;-)

src-schema/Migration.php
```diff
@@ -167,10 +168,16 @@ public function field(string $fieldName, $options = []): self
...
         }
 
-        if ($column->getType() === 'integer' && $refType !== self::REF_TYPE_NONE) {
+        if ($column->getType()->getName() === 'integer' && $refType !== self::REF_TYPE_NONE) {
             $column->setUnsigned(true);
         }
 
+        if (in_array($column->getType()->getName(), ['string', 'text'], true)) {
+            if ($this->getDatabasePlatform() instanceof SqlitePlatform) {
+                $column->setPlatformOption('collation', 'NOCASE');
```
mvorisek, line 177: standard behaviour of MySQL string/text column (but only ASCII support)

mvorisek:
test tested 🚀

src/Model.php
DarkSide666: I wonder why didn't we fix this ages ago. Multiple inserts in single transaction should give noticeable performance improvement.

DarkSide666 (approved):

src/Model.php
mvorisek: they were not in single transaction, but for each import a new transaction was started :D

mvorisek: as major, moved to saparate PR - https://github.com/atk4/data/pull/767

PR #773, opened 2020-10-31
Enable CLOB tests for Oracle, support fixed in dsql

mvorisek:
support fixed in https://github.com/atk4/dsql/pull/273


DarkSide666 (approved):
OK

PR #774, opened 2020-10-31
Fix for dsql - do not compare dsql/Expression on Model::set()

mvorisek:
for https://github.com/atk4/dsql/pull/270

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -700,7 +700,7 @@ class Model implements \IteratorAggregate
         $currentValue = array_key_exists($field, $this->data)
             ? $this->data[$field]
             : (array_key_exists($field, $this->dirty) ? $this->dirty[$field] : $f->default);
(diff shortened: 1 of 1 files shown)


src/Model.php
```diff
@@ -700,7 +700,7 @@ public function set(string $field, $value)
         $currentValue = array_key_exists($field, $this->data)
             ? $this->data[$field]
             : (array_key_exists($field, $this->dirty) ? $this->dirty[$field] : $f->default);
-        if ($f->compare($value, $currentValue)) {
+        if (!$value instanceof \atk4\dsql\Expression && $f->compare($value, $currentValue)) {
```
georgehristov, line 703: How about Expressionable value?

mvorisek: no - see Discord

georgehristov (approved):

PR #777, opened 2020-11-03
Fix Join last inserted ID source

mvorisek:


mvorisek:
@DarkSide666 @georgehristov any idea where the ID is set to NULL and where to fix it?

DarkSide666 (approved):
not sure why it's needed, but ok :)

PR #778, opened 2020-11-03
Fix Oracle insert AI inconsistency

### LINKED ISSUE #714: PostgreSQL - sequence must work on concurent mixed (/w/wo ID) inserts
I think, when ID is not set, we can insert like `INSERT INTO t (id, c, ...) VALUES (xx, c_val, ...)`

where `xx` can be something like `select max(id, seq.nextval)` instead
https://github.com/atk4/data/blob/8f03b553d16f9fc67a3a7880937fe66b940ba007/src/Persistence/Sql.php#L939-L945

note: As PK is unique, this should throw an duplicate PK now, thus not corrupt data.

mvorisek:
fixes https://github.com/atk4/data/issues/714

also fixes AI inconsistency with PostgreSQL, now AI is unified for all DB vendors

related probably with:
- https://github.com/doctrine/orm/commit/a7b9140d2fddcab995b2597be4d589155ff1aa8f
- https://github.com/doctrine/dbal/commit/d2845256d22a0ea2c5e5392aa67f4b95f252d5c4


mvorisek:
@georgehristov I am not exactly sure where the issue come from, can you look at it?

DarkSide666 (approved):
LGTM

PR #780, opened 2020-11-04
Use new Factory methods

mvorisek:
for https://github.com/atk4/core/pull/289


georgehristov (approved):
I am still of the opinion we should change the factory method name but this PR is 👍

mvorisek:
@georgehristov to what name?

georgehristov:
`make`, `build`, etc (if we have to avoid `create`)

mvorisek:
we should aim to not use `Xxx:fromSeed()` to support autohinting

PR #781, opened 2020-11-06
Add test which uses [tags] in SQL expressions

PhilippGrashoff:
This adds a test which uses [tags] in SQL expressions.

### DIFF
--- a/tests/ExpressionSqlTest.php
+++ b/tests/ExpressionSqlTest.php
@@ -26,12 +26,14 @@ class ExpressionSqlTest extends \atk4\schema\PhpunitTestCase
 
     public function testBasic()
     {
-        $this->setDb([
-            'invoice' => [
-                ['total_net' => 10, 'total_vat' => 1.23],
-                ['total_net' => 20, 'total_vat' => 2.46],
-            ],
-        ]);
+        $this->setDb(
+            [
+                'invoice' => [
+                    ['total_net' => 10, 'total_vat' => 1.23],
+                    ['total_net' => 20, 'total_vat' => 2.46],
+                ],
+            ]
+        );
 
         $db = new Persistence\Sql($this->db->connection);
         $i = (new Model($db, 'invoice'))->addFields(['total_net', 'total_vat']);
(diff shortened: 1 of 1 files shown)


georgehristov:
Code should be following I believe (provided Sql persistence, no changes in latest versions):

```
$model->addExpression(
            'full_name_with_salutation',
            $model->expr(
                "CONCAT([salutation], ' ', [full_name])"
            )
        );
```

mvorisek:
@PhilippGrashoff very important - never, never, change CS in specific PRs. If you belive we need to change some CS, propose as a separate PR.

tests/ExpressionSqlTest.php
```diff
@@ -239,4 +252,43 @@ public function testExpressionActionAlias()
         $q = $m->action('fx0', ['sum', 'x']);
         $this->assertEquals([0 => ['sum_x' => 5]], $q->getRows());
     }
+
+    public function testTagsInBracketsCanBeUsedInLiteralSql()
```
mvorisek, line 256: name implies https://github.com/atk4/dsql/pull/264 which I think is unrelated

mvorisek:
please describe in desc what expr are input and what expr should be output - it seems you ask for some dsql feature

PhilippGrashoff:
Thought there was some issue in atk data, but basically lagged the knowledge that Model::expr() is a dynamic method added by Persistence\SQL

mvorisek:
> 
> 
> Thought there was some issue in atk data, but basically lagged the knowledge that Model::expr() is a dynamic method added by Persistence\SQL

does it makes any difference? I am still missing expected input -> output.

PhilippGrashoff:
I basically used tags as in test in my model and they failed, but this was due to not using Model::expr(). I don't see this test as too useful any more as it mainly should test if `[tagsinbrackets]` are used correctly everywhere. Having this additional test I wrote in some class with other non-related Tests does not seem to good any more.

mvorisek:
I mean added expr `first_plus_second` should be usable in another expr. If that is NOT the case, I do not understand the issue.

PhilippGrashoff:
there is no issue, as this works. I thought it wouldnt, but that was due to the fact that I thought Model::expr() was removed and used ->persistence->connection->expr().

mvorisek:
I see, for record:
image

PR #783, opened 2020-11-12
Improve UserAction properties access

ibelar:
Return default caption when not set;
description can use closure;
refactor getConfirmation


DarkSide666 (approved):
LGTM

PR #784, opened 2020-11-19
Fix field SQL identifier render for queries /wo table alias

DarkSide666:
check if needed once https://github.com/atk4/data/pull/938 is fixed, this PR changes does not help

steps to reproduce:

1. open http://localhost/.../atk4/login/demos/admin-roles.php
2. click:
  image
3. add "name" field and click save:
  image
4. error like this is displayed:
  image

### DIFF
--- a/tests/ReferenceSqlTest.php
+++ b/tests/ReferenceSqlTest.php
@@ -224,7 +224,7 @@ class ReferenceSqlTest extends TestCase
         $i->addExpression('total_net', $i->refLink('line')->action('fx', ['sum', 'total_net']));
 
         $this->assertSameSql(
-            'select "invoice"."id", "invoice"."ref_no", (select sum("total_net") from "invoice_line" "_l_6438c669e0d0" where "invoice_id" = "invoice"."id") "total_net" from "invoice"',
+            'select "id", "ref_no", (select sum("total_net") from "invoice_line" "_l_6438c669e0d0" where "invoice_id" = "invoice"."id") "total_net" from "invoice"',
             $i->action('select')->render()
         );
     }
--- a/tests/ReferenceSqlTest.php
+++ b/tests/ReferenceSqlTest.php
@@ -703,7 +703,7 @@ class ReferenceSqlTest extends TestCase
     /**
      * Test update of referenced model.
      */
-    public function testRefModelUpdateAlias()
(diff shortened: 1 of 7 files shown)


georgehristov:
The issue is on this line:
https://github.com/atk4/data/blob/7a39b627c32715a81abfaa94101d994c5c269b18/src/Reference/HasOneSql.php#L191

Model::refLink is used which adds aliases but in DSQL the operations apart from SELECT use template `[table_noalias]`

DarkSide666:
Maybe we should have a way to tell refLink to not add aliases in some specific cases?

mvorisek:
with the latest data (https://github.com/atk4/data/commit/f7805b6ed2f8d3e7345bee30238a4d0a06c59f6b) the test is passing without any change needed

tests/ReferenceSqlTest.php
```diff
@@ -692,4 +692,37 @@ public function testHasOneReferenceType(): void
...
+        $o = (new Model($this->db, ['table' => 'order', 'table_alias' => 'oo']))->addFields(['name']); // table alias here is important otherwise it will use alias "order" which is fine
+
+        $o->hasOne('owner_id', ['model' => $u, 'our_field' => 'owner_id', 'their_field' => 'id'])->withTitle(); // withTitle here is important, in that case it will start adding these aliases
+        $u->hasMany('Orders', ['model' => $o, 'our_field' => 'id', 'their_field' => 'owner_id']);
+
+        $this->assertSame(3, count($u->ref('Orders')->export()));
+
+        $ord = $u/*->load(1)*/->ref('Orders')->load(3);
+        $ord->set('name', 'CCC');
+        // var_dump($ord->action('update')->getDebugQuery()); // update "order" set "name"='CCC'  where "oo"."id" = 3 which is wrong
```
mvorisek, line 723:
```
update
  "order"
set
where
  "oo"."owner_id" in (
    select
      "id"
    from
      "user" "uu"
  )
```

is now dumped/queried

mvorisek (changes requested):
I am still not convinced to fixing this issue for Field and for condition only as proposed by @samsouthardjr fix in #909. However the fix fixes the failing test.

mvorisek:
closing in favor of proper fix in https://github.com/atk4/data/pull/938 incl. test

PR #785, opened 2020-11-27
[fix] Their_field property access

ibelar:
Property their_field is need for properly building ScopeBuilder dropdwon item.

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -80,6 +80,11 @@ class Reference
         $this->link = $link;
     }
 
+    public function getTheirField()
+    {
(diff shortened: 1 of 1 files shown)


DarkSide666 (approved):
LGTM

src/Reference.php
```diff
@@ -80,6 +80,11 @@ public function __construct(string $link)
         $this->link = $link;
     }
 
+    public function getTheirField()
```
mvorisek, line 83: it should have a suffix `Name` and type `: string`

PR #786, opened 2020-11-27
Should use recursive merging

DarkSide666:
Should use recursive merging otherwise ui or any other array property cant be extended only overwritten.

For example, if you already have set some default properties and constructor also sets some, then constructor completely override the default ones.
If default one is `ui => [editable => true]` and you call constructor with `ui => [visible => true]` then you ended up with just `ui => [visible => true]` not `ui => [editable => true, visible => true]` as expected.
That's especially important in case of UI Multiline field which didn't work well because of this.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -210,7 +210,7 @@ class Field implements Expressionable
     {
         foreach ($defaults as $key => $val) {
             if (is_array($val)) {
-                $this->{$key} = array_merge(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
+                $this->{$key} = array_merge_recursive(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
             } else {
                 $this->{$key} = $val;
             }


ibelar (approved):
LGTM

src/Field.php
```diff
@@ -210,7 +210,7 @@ public function __construct(array $defaults = [])
     {
         foreach ($defaults as $key => $val) {
             if (is_array($val)) {
-                $this->{$key} = array_merge(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
+                $this->{$key} = array_merge_recursive(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
```
mvorisek, line 213: this is in a loop already, so it is quite dangerous to merge recursively, which can/would result with mismatched values

mvorisek:
I would say not a bug and this PR is definitely a BC break.

src/Field.php
DarkSide666: In a loop yes, but defaults can be arrays too and these arrays should be merged recursively otherwise it's not possible to extend them - only fully overwrite.

I agree about adding BC break and removing Bug.

mvorisek: yes, but merging elements can result in mismatch, what is the exact usecase we need to deep merge?

PR #787, opened 2020-11-27
Finally fix containsOne/Many UI property overwrites

DarkSide666:


ibelar (approved):
LGTM

src/Reference/ContainsOne.php
```diff
@@ -60,14 +57,18 @@ protected function init(): void
...
-        if (!$ourModel->hasElement($this->our_field)) {
-            $ourModel->addField($this->our_field, [
+        if (!$ourModel->hasElement($ourField)) {
+            $ourModel->addField($ourField, [
                 'type' => $this->type,
                 'reference' => $this,
                 'system' => $this->system,
                 'caption' => $this->caption, // it's ref models caption, but we can use it here for field too
-                'ui' => $this->ui,
+                'ui' => array_merge_recursive([
```
mvorisek, line 68: array_merge_recursive DOES NOT have any affect here -> use `array_merge`

DarkSide666: `ui` is multi-level array property. array_merge will merge only 1st level of array and overwrite all other levels, so it's not possible to extend `ui` with `ui => ['form' => [AnotherClass::class]]` for example without breaking other (maybe default ones) ui properties.

OH crap, you're right. It should be `array_replace_recursive` here - I will fix.

```
$a = ['form' => ['MyForm', 'color'=>'red'], 'table' => ['color'=>'blue']];
$b = ['form' => ['color'=>'green'], 'table'=>['MyTable']];
print_r($c = array_merge_recursive($a,$b)); // will create color as array [red, green] :(
print_r($d = array_replace_recursive($a,$b)); // will correctly replace color
assert($d, ['form' => ['MyForm', 'color'=>'green'], 'table' => ['MyTable', 'color'=>'blue']]);
```
fixed in https://github.com/atk4/data/pull/789

mvorisek: I did not read, but here the 1st arg is NOT nested arrray, thus I do not understand `_recursive`, seems like a typo to me

PR #788, opened 2020-11-27
Improve #785

mvorisek:
Guys, please review better and reject as much as a new code "quickly helful" is general. Every 2nd PR is very shitty.

PS: @ibelar this will probably break something


PhilippGrashoff:
LGTM, method name is more speaking and return type declaration very sensible

ibelar (approved):
Thanks! I will adjust ui pr

PR #789, opened 2020-11-30
correct fix for merging properties

DarkSide666:
It should be array_replace_recursive there.

```
$a = ['form' => ['MyForm', 'color'=>'red'], 'table' => ['color'=>'blue']];
$b = ['form' => ['color'=>'green'], 'table'=>['MyTable']];
print_r($c = array_merge_recursive($a,$b)); // will create color as array [red, green] :(
print_r($d = array_replace_recursive($a,$b)); // will correctly replace color
assert($d, ['form' => ['MyForm', 'color'=>'green'], 'table' => ['MyTable', 'color'=>'blue']]);
```

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -210,7 +210,7 @@ class Field implements Expressionable
     {
         foreach ($defaults as $key => $val) {
             if (is_array($val)) {
-                $this->{$key} = array_merge_recursive(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
+                $this->{$key} = array_replace_recursive(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
             } else {
                 $this->{$key} = $val;
             }
--- a/src/Reference/ContainsOne.php
+++ b/src/Reference/ContainsOne.php
@@ -65,7 +65,7 @@ class ContainsOne extends Reference
                 'reference' => $this,
                 'system' => $this->system,
                 'caption' => $this->caption, // it's ref models caption, but we can use it here for field too
-                'ui' => array_merge_recursive([
+                'ui' => array_replace_recursive([
                     'visible' => false, // not visible in UI Table, Grid and Crud
                     'editable' => true, // but should be editable in UI Form
                 ], $this->ui),
--- a/src/Reference/ContainsOne.php
+++ b/src/Reference/ContainsOne.php
@@ -65,7 +65,7 @@ class ContainsOne extends Reference
                 'reference' => $this,
                 'system' => $this->system,
                 'caption' => $this->caption, // it's ref models caption, but we can use it here for field too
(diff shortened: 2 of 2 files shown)


src/Reference/ContainsOne.php
```diff
@@ -65,7 +65,7 @@ protected function init(): void
                 'reference' => $this,
                 'system' => $this->system,
                 'caption' => $this->caption, // it's ref models caption, but we can use it here for field too
-                'ui' => array_merge_recursive([
+                'ui' => array_replace_recursive([
```
mvorisek: see original PR, I think recursive not needed at all

mvorisek:
better to post as sandpit code - https://3v4l.org/sjJQs

Yes, replace, and use with brutal care!

If I remember there was also some PR in data, check everything.

src/Reference/ContainsOne.php
DarkSide666: See https://3v4l.org/sjJQs for example. That's why we need to merge/replace here recursively.

Form control can have it's own default ui settings (like Multiline control does) and we should be able to extend (not fully overwrite) them. `array_replace_recursive` here does exactly that.

mvorisek: No! here is something totally different, see line 69-70 (that two lines below)

DarkSide666: Yes, by default it's not visible and is editable, but we should be able to overwrite these values without affecting any other values. Well, ok, in this particular case `array_replace` (without recursive) would work too, but ... recursive is more generic.

mvorisek: see, use simple/simply `array_replace` when data are hardcoded!

src/Field.php
```diff
@@ -210,7 +210,7 @@ public function __construct(array $defaults = [])
     {
         foreach ($defaults as $key => $val) {
             if (is_array($val)) {
-                $this->{$key} = array_merge_recursive(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
+                $this->{$key} = array_replace_recursive(is_array($this->{$key} ?? null) ? $this->{$key} : [], $val);
```
DarkSide666, line 213: Usecase:
```
// custom form control class defines
public $ui = ['form' => [CustomField::class, 'size' => 'small', 'color' => 'red']];

// in model i want to use this control with default size (small), but in green color
$this->addField('foo', ['ui' => ['form' => ['color' => 'green']]]);
```

So as result I need to recursively replace ui property not completely overwrite it.
Required result `ui => ['form' => [CustomField::class, 'size' => 'small', 'color' => 'green']]`

src/Reference/ContainsOne.php
DarkSide666: ok will go with `array_merge` then.

PR #790, opened 2020-11-30
Add PHPStan to CI level=1

mvorisek:
rebased onto develop once https://github.com/atk4/data/pull/793 is merged

### DIFF
--- a/tests/FolderTest.php
+++ b/tests/FolderTest.php
@@ -16,7 +16,7 @@ class Folder extends \atk4\data\Model
         $this->addField('name');
 
         $this->hasMany('SubFolder', [new self(), 'their_field' => 'parent_id'])
-            ->addField('count', ['aggregate' => 'count', 'field' => $this->expr('*')]); // TODO PHPStan
+            ->addField('count', ['aggregate' => 'count', 'field' => $this->persistence->expr($this, '*')]);
 
         $this->hasOne('parent_id', new self())
             ->addTitle();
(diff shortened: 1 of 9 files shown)


tests/FolderTest.php
```diff
@@ -16,7 +16,7 @@ protected function init(): void
         $this->addField('name');
 
         $this->hasMany('SubFolder', [new self(), 'their_field' => 'parent_id'])
-            ->addField('count', ['aggregate' => 'count', 'field' => $this->expr('*')]);
+            ->addField('count', ['aggregate' => 'count', 'field' => $this->expr('*')]); // TODO PHPStan
```
mvorisek: `$model->expr()` is not defined, @georgehristov what is a replacement?

DarkSide666: `$model->persistence->expr()`

DarkSide666 (approved):
LGTM except this data dependency on UI
```
  399    Class atk4\ui\Button not found.                                     
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  399    Class atk4\ui\Icon not found.                                       
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  405    Class atk4\ui\Icon not found.                                       
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
 ```

I know it's not the question about this PR, but now we see this problem.
I really think that data should not rely on ui in any way.

I guess this is also reported here https://github.com/atk4/ui/issues/1548

PR #791, opened 2020-12-01
Few fixes for Email field

### LINKED ISSUE #450: Field type 'Email' doesn't work with addFields();
When an 'Email' field type is added within addFields(), it is ignored. This might be a problem for other new fields as well but I haven't tested.

```
// Works
$model->addField('email', ['Email']);

// Doesn't work
$model->addFields([['email', ['Email']]]);
```

### LINKED ISSUE #471: Field type 'Email' stores NULL as empty string
When a NULL value is passed to a field of type 'Email', it seems to be stored as an empty string instead of NULL value.

DarkSide666:
fix #471 
fix #450

### DIFF
--- a/src/Field/Email.php
+++ b/src/Field/Email.php
@@ -49,6 +49,10 @@ class Email extends Field
      */
     public function normalize($value)
     {
+        if ($value === null) {
+            return $value;
+        }
+
         // split value by any number of separator characters
         $emails = preg_split('/[' . implode('', array_map('preg_quote', $this->separator)) . ']+/', $value, -1, PREG_SPLIT_NO_EMPTY);
 
--- a/tests/FieldTypesTest.php
+++ b/tests/FieldTypesTest.php
@@ -41,7 +41,7 @@ class FieldTypesTest extends \atk4\schema\PhpunitTestCase
         $m->save();
         $this->assertSame('foo@example.com', $m->get('email'));
 
-        // padding, spaceing etc removed
+        // padding, spacing etc removed
         $m->set('email', " \t " . 'foo@example.com ' . " \n ");
         $m->save();
         $this->assertSame('foo@example.com', $m->get('email'));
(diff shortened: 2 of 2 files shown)


src/Field/Email.php
```diff
@@ -49,6 +49,10 @@ class Email extends Field
      */
     public function normalize($value)
```
mvorisek, line 50:
```suggestion
    public function normalize($value): ?string
```

tests/FieldTypesTest.php
```diff
@@ -31,21 +31,45 @@ public function testEmailBasic()
...
+        $m->save();
+        $this->assertNull($m->get('email'));
+
+        // normal value
         $m->set('email', 'foo@example.com');
         $m->save();
         $this->assertSame('foo@example.com', $m->get('email'));
 
-        // padding removed
+        // padding, spaceing etc removed
```
mvorisek:
```suggestion
        // padding, spacing etc removed
```

tests/FieldTypesTest.php
```diff
@@ -31,21 +31,45 @@ public function testEmailBasic()
...
         $this->assertSame('foo@example.com', $m->get('email'));
 
+        // no domain - go to hell :)
         $this->expectException(ValidationException::class);
         $this->expectExceptionMessage('does not have domain');
         $m->set('email', 'qq');
     }
 
-    public function testEmailMultiple()
+    public function testMultipleEmailFields()
```
mvorisek: I do not think we need this test

src/Field/Email.php
DarkSide666: Parent Field class do not have return value type set. Will that be OK to set it here?

tests/FieldTypesTest.php
DarkSide666: I added this test here to show how it's working (see https://github.com/atk4/data/issues/450).
But probably you're right that this do not solve anything, just adds another test to prove that #450 can be closed :)

mvorisek: addFields is absolutely unrelated with email, I do not think we should test it, it also add LOC to maintain

PR #792, opened 2020-12-01
Add testcase for never_persist vs never_save

### LINKED ISSUE #408: Model::addExpression typecast/normalization issues with int = 0 and never_persist = true
Hi there,
I just discovered this:
In one of my models, I add an Expression like this:

```
 $this->model->addExpression('ctc_is_approved', [$this->model->expr("(SELECT `is_approved` FROM `company_to_calendar` WHERE `company_id`= [] AND `calendar_id` = `calendar`.`id`)", [$this->app->auth->user->get('id')]), 'type' => 'integer', 'system' => true]);
```
   
In this case, if `is_approved` is set to 0 (zero) in mysql, it gets casted/normalized correctly and represented as int(0) in PHP. However, adding `never_persist => true` to this expression changes this:

```
 $this->model->addExpression('ctc_is_approved', [$this->model->expr("(SELECT `is_approved` FROM `company_to_calendar` WHERE `company_id`= [] AND `calendar_id` = `calendar`.`id`)", [$this->app->auth->user->get('id')]), 'type' => 'integer', 'system' => true, 'never_persist' => true]);
```

Now, if `is_approved` is set to 0 (zero) in mysql, it gets casted/normalized to `NULL` instead of int(0).
Didnt dig deeper into this yet, maybe `never_persist => true` skips persistence typecasting/normalization when loading?

Best regards
Philipp

DarkSide666:
Adds testcase for fix #408 to show that it's as expected.

### DIFF
--- a/tests/ExpressionSqlTest.php
+++ b/tests/ExpressionSqlTest.php
@@ -35,15 +35,25 @@ class ExpressionSqlTest extends \atk4\schema\PhpunitTestCase
         $db = new Persistence\Sql($this->db->connection);
         $i = (new Model($db, 'invoice'));
 
-        $i->addExpression('is_approved', [$i->expr("0"), 'type' => 'integer', 'system' => true]);
-        $i->addExpression('is_sent', [$i->expr("0"), 'type' => 'integer', 'system' => true, 'never_persist' => true]);
+        $i->addExpression('is_bad', [$i->expr("0"), 'type' => 'integer', 'system' => true]);
+        $i->addExpression('is_worse', [$i->expr("0"), 'type' => 'integer', 'system' => true, 'never_save' => true]);
+        $i->addExpression('is_worst', [$i->expr("0"), 'type' => 'integer', 'system' => true, 'never_persist' => true]);
+        $i->addExpression('is_good', [$i->expr("1"), 'type' => 'integer', 'system' => true]);
+        $i->addExpression('is_better', [$i->expr("1"), 'type' => 'integer', 'system' => true, 'never_save' => true]);
+        $i->addExpression('is_best', [$i->expr("1"), 'type' => 'integer', 'system' => true, 'never_persist' => true]);
         $i->loadAny();
 
-var_dump($i->get());
-        $this->assertSame(0, $i->get('is_approved'));
-        $this->assertSame(0, $i->get('is_sent'));
+        // normal fields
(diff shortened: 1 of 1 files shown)


tests/ExpressionSqlTest.php
```diff
@@ -239,4 +239,37 @@ public function testExpressionActionAlias()
         $q = $m->action('fx0', ['sum', 'x']);
         $this->assertEquals([0 => ['sum_x' => 5]], $q->getRows());
     }
+
+
+    public function test408()
```
mvorisek: very bad name

also two blank NL before - fix CS

tests/ExpressionSqlTest.php
```diff
@@ -239,4 +239,37 @@ public function testExpressionActionAlias()
...
+        $this->setDb([
+            'invoice' => [
+                ['foo' => 'bar'],
+            ],
+        ]);
+
+        $db = new Persistence\Sql($this->db->connection);
+        $i = (new Model($db, 'invoice'));
+
+        $i->addExpression('is_bad', [$i->expr("0"), 'type' => 'integer', 'system' => true]);
```
mvorisek: use better names, like `zero_basic`, `zero_never_save`, ... `one_never_persist`

tests/ExpressionSqlTest.php
```diff
@@ -239,4 +239,37 @@ public function testExpressionActionAlias()
...
+    public function test408()
+    {
+        $this->setDb([
+            'invoice' => [
+                ['foo' => 'bar'],
+            ],
+        ]);
+
+        $db = new Persistence\Sql($this->db->connection);
+        $i = (new Model($db, 'invoice'));
```
mvorisek: no need for outer ()

tests/ExpressionSqlTest.php
mvorisek:
```suggestion
    public function testNeverSaveNeverPersist()
```

tests/ExpressionSqlTest.php
```diff
@@ -239,4 +239,37 @@ public function testExpressionActionAlias()
         $q = $m->action('fx0', ['sum', 'x']);
         $this->assertEquals([0 => ['sum_x' => 5]], $q->getRows());
     }
+
```
mvorisek, line 242:
```suggestion
```

PR #800, opened 2020-12-09
Add hintable fields support

mvorisek:
see how to use in https://github.com/atk4/ui/pull/1559

then we can convert all tests to it

### DIFF
--- a/tests/ContainsManyTest.php
+++ b/tests/ContainsManyTest.php
@@ -18,6 +18,9 @@ use Atk4\Data\Model;
 
 /**
  * Invoice model.
+ *
+ * @property Line2  $lines       @Atk\RefMany()
+ * @property string $total_gross @Atk\Field()
  */
 class Invoice2 extends Model
 {
@@ -35,7 +38,7 @@ class Invoice2 extends Model
         $this->containsMany('lines', ['model' => [Line2::class], 'caption' => 'My Invoice Lines']);
 
         // total_gross - calculated by php callback not by SQL expression
-        $this->addCalculatedField('total_gross', function ($m) {
+        $this->addCalculatedField($this->fieldName()->total_gross, function (self $m) {
             $total = 0;
             foreach ($m->ref('lines') as $line) {
                 $total += $line->get('total_gross');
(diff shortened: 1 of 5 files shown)


mvorisek:
@georgehristov can you approve? I will then add to more tests and adjust the ui PR.

DarkSide666:
What license is for `mahalux/atk4-hintable` ?

mvorisek:
> 
> 
> What license is for `mahalux/atk4-hintable` ?

free for use as lib, donations welcomed if helpful for corporate clients :)

tests/ContainsManyTest.php
```diff
@@ -51,8 +52,8 @@ protected function init(): void
         // discounts_total_sum - calculated by php callback not by SQL expression
         $this->addCalculatedField('discounts_total_sum', function (self $m) {
             $total = 0;
-            foreach ($m->ref('lines') as $line) {
-                $total += $line->get('total_gross') * $line->get('discounts_percent') / 100;
+            foreach ($m->lines as $line) {
+                $total += $line->total_gross * $line->get('discounts_percent') / 100;
```
DarkSide666: could be also `$line->discounts_percent`, right?

mvorisek: Of course! About 1000 more replaces ;-)

Working with PHPStan team to fix Model iterator support, then probably replace everything with Rector.

DarkSide666 (approved):
Real magic, but looks good to use and interesting as a bit different coding example.
Also important that it's 100% BC, so if you don't want to use this magic in your code - no worries.

georgehristov:
Really good @mvorisek. 
My only comment would be that it is preferrable to have this into data directly to avoid dependency on external package.

DarkSide666:
> Really good @mvorisek.
> My only comment would be that it is preferrable to have this into data directly to avoid dependency on external package.

Probably not in data directly, but in atk4/core because we can maybe use similar approach for atk4/ui controls too.

tests/ContainsManyTest.php
mvorisek: discussion here: https://github.com/phpstan/phpstan/discussions/4220#discussioncomment-191882

PR #801, opened 2020-12-09
[fix] Model::duplicate

georgehristov:


src/Model.php
```diff
@@ -1178,12 +1178,14 @@ public function reload()
      */
     public function duplicate($newId = null)
     {
-        $this->setId(null);
+        $this->entityId = null;
```
mvorisek: Entity approach must satisty several axioms. One of them is to never allow to assing different ID to the same instance.

So I think we should remove duplicate completely or clone before at the beginning of this method and return the cloned model then.

tests/RandomTest.php
```diff
@@ -497,6 +497,49 @@ public function testNewInstance()
...
+        $m->load(1)->duplicate()->save();
+
+        $this->assertSame([
+            ['dat' => '18/12/12'],
+            ['dat' => '12/12/12'],
+            ['dat' => '18/12/12'],
+        ], $m->export(['dat']));
+    }
+
+    public function testDuplicate2()
```
mvorisek: can we have 3rd test? No, not really, use better test method names...

src/Model.php
georgehristov: Removal of method is not good option as functionality is useful.
Cloning model and returning it is a good option but `$entityId` is private. how to clear it best?

georgehristov: I came up with following code but it does not work as even if model unloaded `$entityId` value remains saved. I think this is not the correct behavior

```
    public function duplicate($newId = null): self
    {
        $duplicate = clone $this;

        $duplicate->unload();

        $duplicate->setMulti($this->data);

        if ($duplicate->id_field) {
            $duplicate->setId($newId);
        }

        return $duplicate;
    }
```

mvorisek: simply set entityId to null after clone, + then set whateer needed (unload, reset dirty, do your research)

tests/RandomTest.php
```diff
@@ -497,6 +497,49 @@ public function testNewInstance()
         $this->assertTrue(isset($a->persistence));
     }
 
+    public function testDuplicateSaveNew()
```
mvorisek, line 500: 👍

src/Model.php
georgehristov: I can make it work but code is not consistent.
The problem I see is that clearing `entityId` should be done with `unload`. What is unload doing otherwise?

mvorisek: Definitely not in unload, as this will open up entity which we do not want.

georgehristov: But unload is exactly clearing the model from the entity.
I know it is "transitional" but something is not right.

mvorisek: really?

DarkSide666: Is entityId cleared when cloning model? Probably should be.

DarkSide666 (approved):
If it works like this (also clears entityId) then it's fine with me.

BUT I think that probably we have to finish refactoring this Model / Entity situation and then quite many internal things will change anyway. Currently this Model/Entity approach looks wrong and hard to use and understand.

src/Model.php
mvorisek: clone should not normally do these things

as a design rule, otherwise there is not a real issues

but to satisfy standards as well, we should do that in special method

now in duplciate() :)

src/Model.php
```diff
@@ -1173,20 +1173,20 @@ public function reload()
      * record in the database.
      *
      * @param mixed $newId
-     *
-     * @return $this
      */
-    public function duplicate($newId = null)
+    public function duplicate($newId = null): self
```
mvorisek: annotate with `@return static` and remove `: self`, otherwise `static` return type is not recognized by most IDEs/analysers (until we can support native PPH 8 `static` return type)

src/Model.php
```diff
@@ -1173,18 +1173,20 @@ public function reload()
...
+        $duplicate = (clone $this);
 
-        if ($this->id_field) {
-            $this->setId($newId);
+        $duplicate->dirty = $this->data;
+
+        $duplicate->entityId = null;
+
+        if ($duplicate->id_field) {
+            $duplicate->setId($newId);
```
mvorisek: this is super questionable, as by "allowing ID to be set easily", we "allow to overwrite existing record with that ID by simply allowing to call save()"

mvorisek (changes requested):

src/Model.php
georgehristov: Yes, we do but this is existing functionality. We should otherwise remove the `$newId` parameter and allow saving as new record only.

mvorisek: I would highly prefer that - throw if more than 0 args are given

src/Model.php
georgehristov: I am still of the opinion that if we support `unload` then entityId should be cleared. All other values are cleared.

mvorisek: You got me for 1, 2 seconds. The answer is clear no.

> Entity approach must satisty several axioms. One of them is to never allow to assing different ID to the same instance.
> 
> So I think we should remove duplicate completely or clone before at the beginning of this method and return the cloned model then.

because if we allow to unset `entityId`, it can be easily changed, and this axiom will be violated. This any code that clears `entityId` is dangerous (except you know what you are dooing or do that on a fresh clone)

georgehristov: Then we should remove `unload` method as it is not working properly at the moment or just use it to return a fresh object without entity loaded.

tests/RandomTest.php
```diff
@@ -497,6 +497,43 @@ public function testNewInstance()
...
+            'rate' => [
+                ['dat' => '18/12/12', 'bid' => 3.4, 'ask' => 9.4],
+                ['dat' => '12/12/12', 'bid' => 8.3, 'ask' => 9.2],
+            ],
+        ]);
+
+        $db = new Persistence\Sql($this->db->connection);
+        $m = new Model_Rate($db);
+
+        $m->load(1)->duplicate()->save();
```
mvorisek, line 512: this is weird with word `duplicate` as save will create a new record

we should rename it to `duplicateAndUnload` at least. Then this PR LGTM.

src/Field.php
```diff
@@ -18,10 +18,10 @@
  */
 class Field implements Expressionable
 {
-    use TrackableTrait;
     use DiContainerTrait;
```
mvorisek: merge develop into this PR instead, see https://github.com/atk4/data/pull/812

tests/RandomTest.php
georgehristov: The whole point of duplicating is to use the existing data for a new record. We are not unloading, only clearing the entityId.
What we can do is include the `save` already in the method.
But the idea is that we can duplicate data, change one or 2 fields and then save.

src/Model.php
georgehristov: Done

mvorisek (approved):
main issues (cloning, no `$newId`) resolved

tests/RandomTest.php
mvorisek: I see, data are retained, LGTM

src/Model.php
mvorisek: `unload` unloads data, but keep entity bound to a specific ID

I think there are usecases for it.

PR #802, opened 2020-12-10
Fix reverse join inconsistencies and defaults

### LINKED ISSUE #799: Reverse join not saving
Following situation:

Table 1 (contact_table): 
contact_id (primary), tbl1_name, tbl1_address

Table_2 (additional_info_table): 
additional_id (primary), additional_contact_id (relation), additional_content

Let's assume I have a contact table which is loaded always (Table 1). 
I want to read, if there's an entry available from an additional table (Table 2). Table 2 always references to Table 1.

```
class myModel extends \atk4\data\Model {
    use \atk4\core\TrackableTrait;
    use \atk4\core\AppScopeTrait;

    public $table = "contact_table";
    public $id_field = "contact_id";
    
    function init(): void {
        parent::init();

        $this->leftJoin($this->'additional_info_table', [
          'master_field' => 'contact_id',
          'foreign_field' => 'additional_contact_id',
          'reverse' => true
        ]);
    }
}
```

When using this model in a form, it reads properly, but it just saves data from the main table. The data of the joined table is neither inserted nor updated.
Is this how it's intended to work? If yes, then how to deal with a situation like this?

DarkSide666:
Also explains, fix #799

### DIFF
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -185,7 +185,6 @@ class Join
         }
 
         if ($this->reverse === true) {
-
             if (isset($this->master_field) && $this->master_field !== $id_field) {
                 throw (new Exception('You are trying to link tables on non-id fields. This is not implemented yet'))
                     ->addMoreInfo('condition', $this->getOwner()->table . '.' . $this->master_field . ' = ' . $this->foreign_table . '.' . $this->foreign_field);
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -185,7 +185,7 @@ class Join
         }
 
         if ($this->reverse === true) {
-            if (isset($this->master_field) && $this->master_field !== $id_field) {
+            if (isset($this->master_field) && $this->master_field !== $id_field) { // TODO not implemented yet, see https://github.com/atk4/data/issues/803
                 throw (new Exception('Joining tables on non-id fields is not implemented yet'))
                     ->addMoreInfo('condition', $this->getOwner()->table . '.' . $this->master_field . ' = ' . $this->foreign_table . '.' . $this->foreign_field);
             }
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -167,30 +167,36 @@ class Join
     {
         $this->_init();
 
+        // owner model should have id_field set
+        if (!$this->getOwner()->id_field) {
+            throw (new Exception('Joins owner model should have id_field set'))
+                ->addMoreInfo('model', $this->getOwner());
+        }
+        $id_field = $this->getOwner()->id_field;
+
         // handle foreign table containing a dot
         if (is_string($this->foreign_table) && strpos($this->foreign_table, '.') !== false) {
+            // split by LAST dot in foreign_table name
+            [$this->foreign_table, $this->foreign_field] = preg_split('/\.+(?=[^\.]+$)/', $this->foreign_table);
+
             if (!isset($this->reverse)) {
                 $this->reverse = true;
-                if (isset($this->master_field)) {
+                if (isset($this->master_field) && $this->master_field !== $id_field) {
                     // both master and foreign fields are set
 
-                    // master_field exists, no we will use that
+                    // master_field exists, now we will use that
                     // if (!is_object($this->master_field) && !$this->getOwner()->hasField($this->master_field)) {
                     throw (new Exception('You are trying to link tables on non-id fields. This is not implemented yet'))
-                        ->addMoreInfo('condition', $this->getOwner()->table . '.' . $this->master_field . ' = ' . $this->foreign_table);
+                        ->addMoreInfo('condition', $this->getOwner()->table . '.' . $this->master_field . ' = ' . $this->foreign_table . '.' . $this->foreign_field);
                     // } $this->reverse = 'link';
                 }
             }
 
-            // split by LAST dot in foreign_table name
-            [$this->foreign_table, $this->foreign_field] = preg_split('/\.+(?=[^\.]+$)/', $this->foreign_table);
-
             if (!$this->master_field) {
-                $this->master_field = 'id';
+                $this->master_field = $id_field;
             }
         } else {
             $this->reverse = false;
-            $id_field = $this->getOwner()->id_field ?: 'id';
             if (!$this->master_field) {
                 $this->master_field = $this->foreign_table . '_' . $id_field;
             }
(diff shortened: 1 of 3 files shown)


tests/JoinArrayTest.php
```diff
@@ -44,6 +44,7 @@ public function testDirection()
         $this->assertSame('test_id', $this->getProtected($j, 'master_field'));
         $this->assertSame('id', $this->getProtected($j, 'foreign_field'));
 
+        $this->expectException(Exception::class);
```
mvorisek, line 47: move `expectException` as close as possible to the statement which throw, if the next statement  throws, delete unneeded assertions

mvorisek:
```suggestion
        $this->expectException(Exception::class); // TODO not implemented yet, see https://github.com/atk4/data/pull/802
```

src/Model/Join.php
```diff
@@ -167,30 +167,39 @@ protected function init(): void
...
+        $id_field = $this->getOwner()->id_field;
+        if (!$id_field) {
+            throw (new Exception('Joins owner model should have id_field set'))
+                ->addMoreInfo('model', $this->getOwner());
+        }
+
+        // handle foreign table containing a dot - that will be reverse join
         if (is_string($this->foreign_table) && strpos($this->foreign_table, '.') !== false) {
+            // split by LAST dot in foreign_table name
+            [$this->foreign_table, $this->foreign_field] = preg_split('/\.+(?=[^\.]+$)/', $this->foreign_table);
```
mvorisek:
```suggestion
            [$this->foreign_table, $this->foreign_field] = preg_split('~\.+(?=[^.]+$)~', $this->foreign_table);
```

if I got the regex correctly...

src/Model/Join.php
```diff
@@ -167,30 +167,39 @@ protected function init(): void
...
-                    // } $this->reverse = 'link';
-                }
             }
+        }
 
-            // split by LAST dot in foreign_table name
-            [$this->foreign_table, $this->foreign_field] = preg_split('/\.+(?=[^\.]+$)/', $this->foreign_table);
+        if ($this->reverse === true) {
+            if (isset($this->master_field) && $this->master_field !== $id_field) {
+                throw (new Exception('You are trying to link tables on non-id fields. This is not implemented yet'))
```
mvorisek:
```suggestion
                throw (new Exception('Joining tables on non-id fields is not implemented yet'))
```

mvorisek: I wonder what is the main issue

DarkSide666: If i remember correctly that was because we can't rely on fields if at least one of them is not unique id field. Something along these lines. Anyway this PR tries not to solve that.

src/Model/Join.php
```diff
@@ -167,30 +167,39 @@ protected function init(): void
...
-                        ->addMoreInfo('condition', $this->getOwner()->table . '.' . $this->master_field . ' = ' . $this->foreign_table);
-                    // } $this->reverse = 'link';
-                }
             }
+        }
 
-            // split by LAST dot in foreign_table name
-            [$this->foreign_table, $this->foreign_field] = preg_split('/\.+(?=[^\.]+$)/', $this->foreign_table);
+        if ($this->reverse === true) {
+            if (isset($this->master_field) && $this->master_field !== $id_field) {
```
mvorisek:
```suggestion
            if (isset($this->master_field) && $this->master_field !== $id_field) { // TODO not implemented yet, see https://github.com/atk4/data/pull/802
```

src/Model/Join.php
mvorisek: > we can't rely on fields if at least one of them is not unique id field.

 is this important if all work is done by SQL?

DarkSide666: All work is not done by SQL I think. At least that wasn't the case time ago. It's important in what order you insert records (reverse or not) and afterwards you have to insert respective record in other table too.

Anyway I specifically didn't address that here!
This PR is just about adding more tests and a bit _normalizing_ method how we define reverse joins. Before there was only a one and not intuitive way how to define them. And as soon as you, for example, explicitly said that join is reverse or even set master-field='id' yourself (same as default value set by ATK) it started to fail with _You are trying to link tables on non-id fields. This is not implemented yet_ exception.

src/Model/Join.php
DarkSide666: This is not related to this PR. It's just to ignore case when user explicitly set master_field which is the same as default one set by ATK. See my other comment above.

src/Model/Join.php
mvorisek: > I specifically didn't address that here

but 1/2 of the added LOC are exceptions for that, better to solve the ID limitation if possible, I am ok with merging this, but with proper issue for that limitation and adding a link to it to the code (like I tried with this PR, replace them)

DarkSide666: These two added expected exceptions in tests was added because previously tests was wrong. They "tested" cases which was not supported, but didn't get proper exception because of slightly wrong exception implementation in Join class (which is now fixed there).
OK created issue https://github.com/atk4/data/issues/803 for that and will see what can be done but outside of this PR.

tests/JoinArrayTest.php
```diff
@@ -44,6 +44,7 @@ public function testDirection()
         $this->assertSame('test_id', $this->getProtected($j, 'master_field'));
         $this->assertSame('id', $this->getProtected($j, 'foreign_field'));
 
+        $this->expectException(Exception::class); // TODO not implemented yet, see https://github.com/atk4/data/pull/802
```
mvorisek:
```suggestion
        $this->expectException(Exception::class); // TODO not implemented yet, see https://github.com/atk4/data/issues/803
```

tests/JoinSqlTest.php
```diff
@@ -41,6 +41,7 @@ public function testDirection()
         $this->assertSame('test_id', $this->getProtected($j, 'master_field'));
         $this->assertSame('id', $this->getProtected($j, 'foreign_field'));
 
+        $this->expectException(Exception::class); // TODO not implemented yet, see https://github.com/atk4/data/pull/802
```
mvorisek:
```suggestion
        $this->expectException(Exception::class); // TODO not implemented yet, see https://github.com/atk4/data/issues/803
```

src/Model/Join.php
mvorisek:
```suggestion
            if (isset($this->master_field) && $this->master_field !== $id_field) { // TODO not implemented yet, see https://github.com/atk4/data/issues/803
```

mvorisek:
@georgehristov can you review/merge?

georgehristov (approved):
LGTM

PR #806, opened 2020-12-11
Implicit ID have default type of integer and is required by default

mvorisek:
slightly a BC break, but not in standard cases


mvorisek:
@DarkSide666 @georgehristov wdyt?

georgehristov (approved):
I'd support this

PR #811, opened 2020-12-17
Change HasOne type to integer by default

mvorisek:
related with https://github.com/atk4/data/pull/806


PhilippGrashoff (approved):
LGTM

mvorisek:
I belive we should throw when DB value is not numeric to prevent enforced cast (by `(int)` cast) of non numerical values.

georgehristov (approved):

PR #817, opened 2020-12-29
Add `Model\AggregateModel` model

georgehristov:
extracted from atk4/report and original https://github.com/atk4/data/pull/677


tests/Model/Client.php
```diff
@@ -6,6 +6,8 @@
 
 class Client extends User
 {
+    public $table = 'client';
```
mvorisek, line 9: previously not needed, and now it is?

src/Model/AggregatesTrait.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Provides aggregation methods.
+ */
+trait AggregatesTrait
+{
+    /**
+     * @see Aggregate::withAggregateField.
+     */
+    public function withAggregateField($name, $seed = []): Model
```
mvorisek: add phpdoc/strong types, then pass these params, `...func_get_args()` can not be analysed

src/Model/Aggregate.php
```diff
@@ -0,0 +1,316 @@
...
+    {
+        $condition = $condition ?? $this->scope();
+
+        if (!$condition->isEmpty()) {
+            // peel off the single nested scopes to convert (((field = value))) to field = value
+            $condition = $condition->simplify();
+
+            // simple condition
+            if ($condition instanceof Model\Scope\Condition) {
+                $query->having(...$condition->toQueryArguments());
```
mvorisek: `having` is an interesting argument for `Aggregate` class, but we may use HAVING instead or WHERE when grouping is active in general (before we can analyse if HAVING is really needed)

src/Model/Aggregate.php
```diff
@@ -0,0 +1,316 @@
...
+     * @var bool
+     */
+    public $read_only = true;
+
+    /**
+     * Aggregate does not have ID field.
+     *
+     * @var string
+     */
+    public $id_field;
```
mvorisek: `read_only`, `id_field`, ... should be probably set in init

src/Model/Aggregate.php
```diff
@@ -0,0 +1,316 @@
...
+        parent::__construct($baseModel->persistence, $defaults);
+
+        // always use table prefixes for this model
+        $this->persistence_data['use_table_prefixes'] = true;
+    }
+
+    /**
+     * Specify a single field or array of fields on which we will group model.
+     *
+     * @param array $fields    Array of field names
```
mvorisek: never use array in phpdoc, always be more specific like `string[]`

tests/Model/Client.php
georgehristov: Was discussed before but now I modified code to avoid changing this model

tests/Model/Client.php
```diff
@@ -6,8 +6,6 @@
 
 class Client extends User
 {
-    public $table = 'client';
```
mvorisek: well, I am ok with table here - it is probably how you should model Models, it was a question if this change is related with refactoring (and therefore a separate PR for it is better/clearer) or needed with this PR

src/Model/Aggregate.php
```diff
@@ -0,0 +1,316 @@
...
+     * @param array $aggregate Array of aggregate mapping
+     *
+     * @return $this
+     */
+    public function groupBy(array $fields, array $aggregate = []): Model
+    {
+        $this->group = $fields;
```
mvorisek: I think we should allow to call this method multiple times.

`$aggregate` param - what is the advantage of it instead of `addExpression` or this nodel or simillar method?

src/Model/Aggregate.php
georgehristov: moved to constructor

src/Model/Aggregate.php
georgehristov: > I think we should allow to call this method multiple times.

I did modify it already

> $aggregate param - what is the advantage of it instead of addExpression or this nodel or simillar method?

I played a bit to use addExpression natively but it mixes things up as aggregate expressions use base model fields. Then also it messes up with base model addExpression and you cannot do `$model->addExpression('xx')->groupBy(['yy'])` but need to do `$model->groupBy` first to create the aggregate. So multiple issues in this approach.

src/Model/Aggregate.php
georgehristov: updated

src/Model/Aggregate.php
```diff
@@ -0,0 +1,299 @@
...
+    /**
+     * Specify a single field or array of fields on which we will group model.
+     *
+     * @param mixed[] $aggregateExpressions Array of aggregate expressions with alias as key
+     *
+     * @return $this
+     */
+    public function groupBy(array $fields, array $aggregateExpressions = []): Model
+    {
+        $this->groupByFields = array_unique($this->groupByFields + $fields);
```
mvorisek: array_unique(array_merge()) ?

georgehristov: Merge should be better (I used origunal code here)

src/Model/AggregatesTrait.php
georgehristov: Done

src/Model/Aggregate.php
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    /**
+     * Specify a single field or array of fields on which we will group model.
+     *
+     * @param mixed[] $aggregateExpressions Array of aggregate expressions with alias as key
+     *
+     * @return $this
+     */
+    public function groupBy(array $fields, array $aggregateExpressions = []): Model
```
mvorisek:
```suggestion
    public function groupBy(array $fields, array $aggregateExpressions = [])
```

as then phpdoc does not work in some IDEs (static type)

tests/Model/Client.php
georgehristov: No breaking changes but anyway I moved setup to the new code

src/Model/Aggregate.php
```diff
@@ -0,0 +1,299 @@
...
+            $this->addExpression($name, $seed);
+        }
+
+        return $this;
+    }
+
+    /**
+     * Return reference field.
+     *
+     * @param string $link
```
mvorisek: to strong type

src/Model/Aggregate.php
```diff
@@ -0,0 +1,299 @@
...
+
+                // select but no need your fields
+                $query = $this->baseModel->action($mode, [false]);
+
+                $this->initQueryFields($query, array_unique($fields + $this->groupByFields));
+                $this->initQueryOrder($query);
+                $this->initQueryGrouping($query);
+                $this->initQueryConditions($query);
+
+                $this->hook(self::HOOK_INIT_SELECT_QUERY, [$query]);
```
mvorisek, line 197: why to not use `Persistence\Sql::HOOK_INIT_SELECT_QUERY` (just neutral question, no sudgestion for now)

src/Model/Aggregate.php
mvorisek: great!

docs/aggregates.rst
```diff
@@ -0,0 +1,51 @@
...
+     [
+       'country',
+       'count'=>'count',
+       'total_amount'=>['sum', 'amount']
+     ],
+   );
+   
+   foreach ($aggregate as $row) {
+     var_dump(json_encode($row));
+     // ['country'=>'UK', 'count'=>20, 'total_amount'=>123.20];
```
mvorisek: this is not an output of var_dump ;-)

tests/ModelAggregateTest.php
```diff
@@ -0,0 +1,318 @@
...
+    protected $invoice;
+    /** @var \Atk4\Data\Model|null */
+    protected $invoiceAggregate;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->setDB($this->init_db);
+
+        $this->invoice = new Model\Invoice($this->db);
```
mvorisek: move to `createInvoice` method, then we do not need clone too

src/Model/Aggregate.php
georgehristov: Done

src/Model/Aggregate.php
georgehristov: Done

src/Model/Aggregate.php
georgehristov: Done

tests/ModelAggregateTest.php
georgehristov: Done

tests/ModelAggregateTest.php
```diff
@@ -27,33 +28,28 @@ class ModelAggregateTest extends \Atk4\Schema\PhpunitTestCase
...
-
     protected function setUp(): void
     {
         parent::setUp();
         $this->setDB($this->init_db);
+    }
 
-        $this->invoice = new Model\Invoice($this->db);
-        $this->invoice->getRef('client_id')->addTitle();
+    protected function createInvoice()
```
mvorisek:
```suggestion
    protected function createInvoice(): Model\Invoice
```

docs/aggregates.rst
georgehristov: updated

src/Model/Aggregate.php
mvorisek: I am ok with this for now, as related only with this method... But keep this unresolved in GH discussion.

src/Model/Aggregate.php
mvorisek: Ok for now if we will integrate it for wrapping like now... But keep this unresolved in GH discussion.

tests/ModelAggregateTest.php
```diff
@@ -0,0 +1,314 @@
...
+            'select count(*) from ((select 1 from "invoice" group by "client_id")) der',
+            $aggregate->action('count')->render()
+        );
+    }
+
+    public function testAggregateFieldExpression()
+    {
+        $aggregate = $this->createInvoiceAggregate();
+
+        $aggregate->groupBy(['abc'], [
```
DarkSide666: It creates new field 'abc' in baseModel too?
I don't think we should allow to group by fields which do not already exist in baseModel.

Also in select it doesn't add table alias in `group by abc` and don't even escape abc.

src/Model/Aggregate.php
```diff
@@ -0,0 +1,294 @@
...
+ * but when querying it will use the original model to calculate the query, then add grouping and aggregates.
+ *
+ * If you wish you can add more fields, which will be passed through:
+ * $aggregate->addField('middle');
+ *
+ * If this field exist in the original model it will be added and you'll get exception otherwise. Finally you are
+ * permitted to add expressions.
+ *
+ * The base model must not be Union model or another Aggregate model, however it's possible to use Aggregate model as nestedModel inside Union model.
+ * Union model implements identical grouping rule on its own.
```
DarkSide666: Would be good to allow Union inside Aggregate and nested aggregate models, but ... that's maybe for future.

src/Model/Aggregate.php
```diff
@@ -0,0 +1,294 @@
...
+     * @param mixed[] $aggregateExpressions Array of aggregate expressions with alias as key
+     *
+     * @return $this
+     */
+    public function groupBy(array $fields, array $aggregateExpressions = [])
+    {
+        $this->groupByFields = array_unique(array_merge($this->groupByFields, $fields));
+
+        foreach ($fields as $fieldName) {
+            $this->addField($fieldName);
```
DarkSide666: Impossible to pass seed for these fields to define field type for example.
It could be in format `['fieldName' => ['type' => 'money']]`.

src/Model/Aggregate.php
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        return $field
+            ? parent::addField($name, $field)->setDefaults($seed)
+            : parent::addField($name, $seed);
+    }
+
+    public function setLimit(int $count = null, int $offset = 0)
+    {
+        $this->baseModel->setLimit($count, $offset);
```
DarkSide666: I think this is wrong. It should set limit on aggregated itself model not baseModel.

src/Model/Aggregate.php
```diff
@@ -0,0 +1,294 @@
...
+                $query = $this->baseModel->action($mode, $args);
+
+                $query->reset('field')->field($this->expr('1'));
+                $this->initQueryGrouping($query);
+
+                $this->hook(self::HOOK_INIT_SELECT_QUERY, [$query]);
+
+                return $query->dsql()->field('count(*)')->table($this->expr('([]) der', [$query]));
+            case 'field':
+            case 'fx':
```
DarkSide666: Not sure if `field` should be allowed and also how `fx` works. Need to see some usecase example in SQL and tests.

src/Model/Aggregate.php
DarkSide666: I guess `having` is always needed in place of `where` in case model is grouped and condition is applied on such aggregate model itself.

src/Model/Aggregate.php
```diff
@@ -0,0 +1,294 @@
...
+            // peel off the single nested scopes to convert (((field = value))) to field = value
+            $condition = $condition->simplify();
+
+            // simple condition
+            if ($condition instanceof Model\Scope\Condition) {
+                $query->having(...$condition->toQueryArguments());
+            }
+
+            // nested conditions
+            if ($condition instanceof Model\Scope) {
```
DarkSide666, line 267: would `elseif` be better here?

DarkSide666 (changes requested):
Reviewed. Comments added.
Overall I like implementation, but few things are questionable:

1. don't think that `withAggregateField` method is needed at all because that's the same as 1st argument of `groupBy`. Also `addField` does all that work anyway, so what's the point of having another method.
2. as result not sure if we need trait for just adding one method - `groupBy`.
3. Aggregate model `action()` method partly duplicates code from other class. Maybe we can somehow at least limit code duplication?
4. `setLimit` is wrong.
5. not sure about `addField` implementation. Why some fields are added as fields, but others as expressions? I think `addField` should always clone baseModel field and attach it to aggregate model.
6. few more tests needed - for `setLimit`, for `fx` action, for `order by`, for some nested / more complex conditions etc.

src/Model/Aggregate.php
georgehristov: Done

src/Model/Aggregate.php
mvorisek: not if we wrap as subquery...

src/Model/Aggregate.php
```diff
@@ -0,0 +1,299 @@
...
+     *
+     * $model->groupBy(['abc'])->withAggregateField('xyz');
+     *
+     * and
+     *
+     * $model->withAggregateField('xyz')->groupBy(['abc']);
+     */
+    public function withAggregateField(string $name, $seed = []): Model
+    {
+        static::addField($name, $seed);
```
mvorisek: I wonder why not `$this->addField`.

Let's finish all feedback to this PR and merge it.

src/Model/Aggregate.php
mvorisek: Related with https://github.com/atk4/data/pull/853/files#diff-7e7810cb7d196a13e4c0b10d1c737e5ead3e3169bd82f6b4c1149dcf67d726f0R384 , I belive wrapping is the only solution, `having` requires `group by` across all columns for some DB vendors (even if grouped by some unique column)

src/Model/AggregatesTrait.php
```diff
@@ -0,0 +1,35 @@
...
+trait AggregatesTrait
+{
+    /**
+     * @param array|object $seed
+     *
+     * @return Aggregate
+     *
+     * @see Aggregate::withAggregateField.
+     */
+    public function withAggregateField(string $name, $seed = []): Model
```
mvorisek, line 21: should return `static`, eg. the `Aggregate` model wrapped in current Model to support hinting

also, I think this will fail with DB vendors as some require all non-unique fields to be specified to be somehow aggregated

mvorisek:
@georgehristov can you please make the tests passing?

src/Model/Aggregate.php
mvorisek: TODO after https://github.com/atk4/data/pull/946 - traversing may be implemented using model wrapping

but it will require "virtualized Fields" - maybe we should always impl. different persistence fields using specific subobjects

src/Model/Aggregate.php
mvorisek: TODO after https://github.com/atk4/data/pull/946

src/Model/Aggregate.php
```diff
@@ -0,0 +1,298 @@
...
+        $this->table = $baseModel->table;
+
+        // this model does not have ID field
+        $this->id_field = null;
+
+        // this model should always be read-only
+        $this->read_only = true;
+
+        parent::__construct($baseModel->persistence, $defaults);
+    }
```
mvorisek, line 74: `$this->persistence_data['use_table_prefixes'] = true;` line was here, if present, tests do not pass

mvorisek: fine, was an issue in "Model2" testing code

src/Model/AggregatesTrait.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return (new Aggregate($this))->withAggregateField($name, $seed);
+    }
+
+    /**
+     * @return Aggregate
+     *
+     * @see Aggregate::groupBy.
+     */
+    public function groupBy(array $fields, array $aggregateExpressions = []): Model
```
mvorisek: function name is brutally confusing as it returns a new Model, it should be named like `createAggregate`

alternatively, the Aggregate model can be added to a property and honored

tests/ModelAggregateTest.php
mvorisek: > Also in select it doesn't add table alias in `group by abc` and don't even escape abc.

fixed - now correctly escaped

> It creates new field 'abc' in baseModel too? I don't think we should allow to group by fields which do not already exist in baseModel.

TODO you are right, grouping by a non-selecting field does not make a lot of sense, as then it is unknown to which the result is

docs/aggregates.rst
```diff
@@ -0,0 +1,43 @@
...
+--------
+
+Aggregate model can be used for grouping::
+
+   $aggregate = $orders->groupBy(['country_id']);
+
+`$aggregate` above is a new object that is most appropriate for the model's persistence and which can be manipulated 
+in various ways to fine-tune aggregation. Below is one sample use::
+
+   $aggregate = $orders->withAggregateField('country')->groupBy(['country_id'], [
```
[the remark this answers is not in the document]
mvorisek: marking as resolved as it is a duplicate of Imants review point 1

src/Model/Aggregate.php
```diff
@@ -0,0 +1,294 @@
...
+ * $aggregate = new Aggregate($mymodel);
+ * $aggregate->groupBy(['first','last'], ['salary'=>'sum([])'];
+ *
+ * your resulting model will have 3 fields:
+ *  first, last, salary
+ *
+ * but when querying it will use the original model to calculate the query, then add grouping and aggregates.
+ *
+ * If you wish you can add more fields, which will be passed through:
+ * $aggregate->addField('middle');
```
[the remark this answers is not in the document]
mvorisek: it is the same, doc is correct here

src/Model/Aggregate.php
mvorisek: TODO update docs before merge, thanks to https://github.com/atk4/data/pull/946 unlimited/unrestricted nesting is possible

src/Model/Aggregate.php
mvorisek: I belive seed should be not allowed. Instead a "virtualized Field" from the source model must be used. Or the field must be added using regular `addField` method.

src/Model/Aggregate.php
mvorisek: fixed

src/Model/Aggregate.php
mvorisek: fixed

src/Model/Aggregate.php
mvorisek: method removed

src/Model/AggregatesTrait.php
mvorisek: we should probably rename to `setGroupBy` to be consistent with Model methods naming

mvorisek: fixed

README.md
```diff
@@ -182,16 +182,16 @@ $grid->setModel($data);
...
 ``` php
 $chart = new \Atk4\Chart\BarChart();
 $data = new JobReport($db);
 
 // BarChart wants aggregated data
 $data->addExpression('month', 'month([date])');
-$aggregate = new \Atk4\Report\GroupModel($data);
-$aggregate->groupBy('month', ['profit_margin' => 'sum']);
+$aggregate = new AggregateModel($data);
+$aggregate->setGroupBy('month', ['profit_margin' => 'sum']);
```
mvorisek: should be `addGroupBy` & /wo 2nd param

at least one group by expression should be required

mvorisek:
@georgehristov this PR is almost done, please let me know if you have any feedback

tests/ModelAggregateTest.php
mvorisek: not much useful, but valid in sense of usage/sql

PR #819, opened 2020-12-29
Table must be specified with table key in defaults

mvorisek:


DarkSide666 (approved):
LGTM, just need to fix tests appropriately.

PR #820, opened 2020-12-30
Require array in reference/join defaults

mvorisek:
hasOne/hasMany was nice with string/object, but allowing string violates seed definition

if we however allow a seed for Model directly (without `['model' => ...]`) then it can not be distinguished from type only (array vs. array)

thus this impl. is strict, but the only one simple and reliable

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ class JobReport extends Job {
     $invoice_lines = $invoice->ref('Lines');
 
     // Build relation between job and invoice line
-    $this->hasMany('InvoiceLines', $invoice_lines)
+    $this->hasMany('InvoiceLines', ['model' => $invoice_lines])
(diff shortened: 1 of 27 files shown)


DarkSide666 (approved):
This is quite big BC break, but for consistency and good code sake it should be like that.

One thing to also think about is maybe rename join master_field -> our_field and foreign_field -> their_field? Like we have in references? That would be more consistent then too. Of course that would also be BC break, but ...

georgehristov:
> One thing to also think about is maybe rename join master_field -> our_field and foreign_field -> their_field? Like we have in references? That would be more consistent then too. Of course that would also be BC break, but ...

This idea I would support. For consistency we should also consider `ourField` and `theirField` names in both references and joins

mvorisek:
I have no resources to pursue another renames, but I like that idea to.

PR #821, opened 2020-12-30
Fix PHPStan 4267 issue until fixed officially

mvorisek:
related with https://github.com/phpstan/phpstan/issues/4267

otherwise `foreach ($model as $itemModel)` `$itemModel` is assumed as `Model` instead of `static`


mvorisek:
@georgehristov please approve

georgehristov (approved):

PR #822, opened 2020-12-30
Convert hardcoded strings to hintable fields in Contains{One, Many}Test

mvorisek:
related with https://github.com/atk4/data/pull/800

### DIFF
--- a/src/Model/ReferencesTrait.php
+++ b/src/Model/ReferencesTrait.php
@@ -80,7 +80,7 @@ trait ReferencesTrait
      *
      * @return Reference\HasOne
      */
-    public function hasOne(string $link, array $defaults = []): Reference
+    public function hasOne(string $link, array $defaults = []) //: Reference
     {
         return $this->_hasReference($this->_default_seed_hasOne, $link, $defaults); // @phpstan-ignore-line
     }
@@ -90,7 +90,7 @@ trait ReferencesTrait
      *
      * @return Reference\HasMany
      */
-    public function hasMany(string $link, array $defaults = []): Reference
+    public function hasMany(string $link, array $defaults = []) //: Reference
     {
         return $this->_hasReference($this->_default_seed_hasMany, $link, $defaults); // @phpstan-ignore-line
     }
@@ -100,7 +100,7 @@ trait ReferencesTrait
      *
      * @return Reference\ContainsOne
      */
-    public function containsOne(string $link, array $defaults = []): Reference
+    public function containsOne(string $link, array $defaults = []) //: Reference
     {
         return $this->_hasReference($this->_default_seed_containsOne, $link, $defaults); // @phpstan-ignore-line
     }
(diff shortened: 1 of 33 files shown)


tests/ContainsManyTest.php
```diff
@@ -19,8 +19,9 @@
 /**
  * Invoice model.
  *
- * @property Line2  $lines       @Atk\RefMany()
- * @property string $total_gross @Atk\Field()
+ * @property Line2  $lines               @Atk\RefOne()
```
mvorisek: always use `RefOne`, `RefMany` should be used once `load()` returns another instance (otherwise you will not be able to laod a model, you must get it thru iterator first)

tests/ContainsManyTest.php
```diff
@@ -35,13 +36,13 @@ protected function init(): void
...
         // will contain many Lines
-        $this->containsMany('lines', ['model' => [Line2::class], 'caption' => 'My Invoice Lines']);
+        $this->containsMany($this->fieldName()->lines, ['model' => [Line2::class], 'caption' => 'My Invoice Lines']);
 
         // total_gross - calculated by php callback not by SQL expression
         $this->addCalculatedField($this->fieldName()->total_gross, function (self $m) {
             $total = 0;
-            foreach ($m->ref('lines') as $line) {
-                $total += $line->get('total_gross');
+            foreach ($m->lines as $line) {
```
mvorisek: this is one strong advantage of hintable magic, thanks to https://github.com/atk4/data/pull/821 foreach value has correct type for phpstan as the iterated model

src/Model/ReferencesTrait.php
```diff
@@ -80,7 +80,7 @@ public function addRef(string $link, array $defaults): Reference
      *
      * @return Reference\HasOne
      */
-    public function hasOne(string $link, array $defaults = []): Reference
+    public function hasOne(string $link, array $defaults = []) //: Reference
```
mvorisek, line 83: otherwise autohinting (from phpdoc) does not work

PhilippGrashoff: Why not add HasOne to Method Signature? 
```
public function hasOne(string $link, array $defaults = []): Reference\HasOne
```

mvorisek: because it can be implemented by another class not based on this one

this is almost the no1 issue with Atk

mvorisek:
@georgehristov can we merge?

georgehristov (approved):

src/Model/ReferencesTrait.php
PhilippGrashoff: so you mean a child class could e.g. implement 
`public function hasOne(string $link, array $defaults = []): Some\Other\Reference`
and that would fail if we set the parent signature to Reference\HasOne?

mvorisek: yes

PhilippGrashoff: grml this is ugly. Do you really think adding `Reference\HasOne` would be an issue? I see return types in method signatures as an important part of good code. While this PR helps IDEs and phpstan and so on using comments, it weakens the method signatures which was something I liked about almost any of your PR: You created stricter, clearer method signatures.

PR #826, opened 2021-01-16
Remove Model::addFields() method

mvorisek:
and `addField` should probably return Model for chaining, field is named (1st param)...

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,8 @@ class Client extends \Atk4\Data\Model {
   function init(): void {
     parent::init();
 
-    $this->addFields(['name','address']);
+    $this->addField('name');
+    $this->addField('address');
 
     $this->hasMany('Project', ['model' => [Project::class]]);
   }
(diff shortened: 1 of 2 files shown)


acicovic:
@mvorisek what's your reasoning for removing addFields? It does reduce code repetition a lot when you have many fields.

mvorisek:
you can assign complex seed to a variable first if needed

acicovic:
Could  you demonstrate?

mvorisek:
Of course!

```
$this->addFields(['name','address'], ['you must get it now!']);
```
to
```
$seed = ['you must get it now!'];
$this->addField('name', $seed);
$this->addField('address', $seed);
```

georgehristov:
I would not support this change

acicovic:
Personally me neither, but it's up to you guys to decide.

DarkSide666:
No I don't support this.
I see no advantage in removing that method. It only removes existing short-hand functionality.

mvorisek:
not pursuing this for now

PR #830, opened 2021-01-19
Check hasMany guess early

mvorisek:
Join has simillar guessing, but as it operates on `foreign_table` instead of Model, we can not check early there

### DIFF
--- a/src/Reference/HasMany.php
+++ b/src/Reference/HasMany.php
@@ -23,13 +23,13 @@ class HasMany extends Reference
         // this is pure guess, verify if such field exist, otherwise throw
         // TODO probably remove completely in the future
         $ourModel = $this->getOurModel();
-        $theirField = $ourModel->table . '_' . $ourModel->id_field;
-        if (!$this->createTheirModel()->hasField($theirField)) {
+        $theirFieldName = $ourModel->table . '_' . $ourModel->id_field;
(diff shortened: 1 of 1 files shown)


src/Reference/HasMany.php
```diff
@@ -14,6 +14,24 @@
...
+        // this is pure guess, verify if such field exist, otherwise throw
+        // TODO probably remove completely in the future
+        $ourModel = $this->getOurModel();
+        $theirField = $ourModel->table . '_' . $ourModel->id_field;
+        if (!$this->createTheirModel()->hasField($theirField)) {
+            throw (new Exception('Their model does not contain fallback field'))
+                ->addMoreInfo('their_fallback_field', $theirField);
+        }
+
+        return $theirField;
```
georgehristov: Maybe assign it to `$this->their_field`?

georgehristov (commented):
This version I like

georgehristov (approved):

src/Reference/HasMany.php
georgehristov: And rename `$theirField` to `$theirFieldName` for consistency...

mvorisek: no - we should not materialize guesses if not done on construction time

PR #831, opened 2021-01-20
Nice caption with debug {actual, hintable} prefix

mvorisek:
To address https://github.com/atk4/ui/pull/1559#issuecomment-762305894

`atk_fp_xxx__` = field prefix (for testing hintable in ui)
`atk_afp_xxx__` = actual field prefix (for testing `actual` field name for `Field`)

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -610,7 +610,7 @@ class Field implements Expressionable
      */
     public function getCaption(): string
     {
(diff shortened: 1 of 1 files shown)


DarkSide666 (approved):
Probably you know what you're doing. I have no idea about why we need all these hintable things there :D

PR #833, opened 2021-01-22
fix CS

DarkSide666:


.github/workflows/unit-tests.yml
```diff
@@ -140,7 +140,7 @@ jobs:
...
+          if [ "${{ matrix.type }}" == "Phpunit Burn" ]; then sed -i 's/ *public function runBare(): void/public function runBare(): void { gc_collect_cycles(); $mem0 = memory_get_usage(); for ($i = 0; $i < '"$(if [ \"$GITHUB_EVENT_NAME\" == \"schedule\" ]; then echo 50; else echo 5; fi)"'; ++$i) { $this->_runBare(); gc_collect_cycles(); if ($i === 0) { $mem1 = memory_get_usage(); } } $mem2 = memory_get_usage(); if ($mem2 - 3072 * 1024 > $mem0 || $mem2 - 128 * 1024 > $mem1) { $this->onNotSuccessfulTest(new AssertionFailedError("Memory leak detected! (" . round($mem0 \/ (1024 * 1024), 3) . " + " . round(($mem1 - $mem0) \/ (1024 * 1024), 3) . " + " . round(($mem2 - $mem1) \/ (1024 * 1024), 3) . " MB, " . $i . " iterations)")); } } private function _runBare(): void/' vendor/phpunit/phpunit/src/Framework/TestCase.php && cat vendor/phpunit/phpunit/src/Framework/TestCase.php | grep '_runBare(' ; fi
```
mvorisek: no - we need this for catch connections issues

DarkSide666: Our scheduled burn test already executes more than 1h. It's to much. Should be less than 1h.

mvorisek: What is wrong with 1 hour? :D

mvorisek:
```suggestion
          if [ "${{ matrix.type }}" == "Phpunit Burn" ]; then sed -i 's/ *public function runBare(): void/public function runBare(): void { gc_collect_cycles(); $mem0 = memory_get_usage(); for ($i = 0; $i < '"$(if [ \"$GITHUB_EVENT_NAME\" == \"schedule\" ]; then echo 100; else echo 5; fi)"'; ++$i) { $this->_runBare(); gc_collect_cycles(); if ($i === 0) { $mem1 = memory_get_usage(); } } $mem2 = memory_get_usage(); if ($mem2 - 3072 * 1024 > $mem0 || $mem2 - 128 * 1024 > $mem1) { $this->onNotSuccessfulTest(new AssertionFailedError("Memory leak detected! (" . round($mem0 \/ (1024 * 1024), 3) . " + " . round(($mem1 - $mem0) \/ (1024 * 1024), 3) . " + " . round(($mem2 - $mem1) \/ (1024 * 1024), 3) . " MB, " . $i . " iterations)")); } } private function _runBare(): void/' vendor/phpunit/phpunit/src/Framework/TestCase.php && cat vendor/phpunit/phpunit/src/Framework/TestCase.php | grep '_runBare(' ; fi
```

mvorisek (approved):

.github/workflows/unit-tests.yml
DarkSide666: Because scheduled each hour, so as result it's non-stop testing.

mvorisek: Yes, there is an overlap. But in reality, what is wrong with that?

PR #834, opened 2021-02-25
Add additional Join test with field actual names and prefixes

DarkSide666:


tests/JoinSqlTest.php
```diff
@@ -664,4 +664,87 @@ public function testJoinReverseOneOnOne()
...
+        $m_u2 = (clone $m_u)->load(1);
+        $m_u2->set('name', 'John 2');
+        $m_u2->set('j1_phone', '+555');
+        $m_u2->set('j2_salary', 111);
+        $m_u2->save();
+
+        $this->assertEquals(
+            [
+                'user' => [
+                    1 => ['id' => 1, 'first_name' => 'John 2', 'cid' => 1],
```
mvorisek, line 709: I belive this is wrong, as `actual` name should be used for low level SQL only, otherwise field name (not `actual` name) should be used.

This line should be:
```
1 => ['id' => 1, 'first_name' => 'John 2', 'contact_id' => 1],
```

DarkSide666: Not really. This is just direct dump from DB, so there will be actual names here.

mvorisek: You are right, but in export and other non-low-level dumps it should not have the `actual` name.

DarkSide666: yes in import / export should not have actual names.

But as you see in Join class it still will use actual names when linking tables. See I added comment there in this PR.
I don't think we have to change that, but having additional note about that could be good.

mvorisek (approved):
join use `actual` only, but that seems enough for now....

PR #835, opened 2021-02-26
Fix title field name guessing

mvorisek:
for https://github.com/atk4/ui/pull/1595

guessing is always a compromise, but if `id_field` is prefixed, removing `'id'` should be ok as a fallback

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -195,7 +195,7 @@ class Reference
             $ourModel = $this->getOurModel();
 
             $this->table_alias = $this->link;
-            $this->table_alias = preg_replace('~_(' . $ourModel->id_field . '|id)~', '', $this->table_alias);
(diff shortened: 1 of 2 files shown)


src/Reference.php
```diff
@@ -195,8 +195,7 @@ protected function initTableAlias(): void
             $ourModel = $this->getOurModel();
 
             $this->table_alias = $this->link;
-            $this->table_alias = preg_replace('/_' . ($ourModel->id_field ?: 'id') . '/', '', $this->table_alias);
-            $this->table_alias = preg_replace('/([a-zA-Z])[a-zA-Z]*[^a-zA-Z]*/', '\1', $this->table_alias);
```
mvorisek, line 199: @DarkSide666 @georgehristov what was purpose of this regex replace?

DarkSide666 (approved):
LGTM

PR #837, opened 2021-03-03
Add model in ValidationExceptions

DarkSide666:


mvorisek:
shouldn't we make the model parameter required then?

DarkSide666:
I didn't want to make some breaking changes. Theoretically field can exist without model. So I think this PR is good as it is.

Also theoretically possible that fields there could be from multiple models, because first argument can be array of fields=>errors.

PR #838, opened 2021-03-03
Feature/fix conditions

DarkSide666:


mvorisek (approved):
LGTM

src/Persistence/Sql/Join.php
```diff
@@ -250,8 +250,10 @@ public function doDelete($id): void
             return;
         }
 
-        $id = $this->reverse ? $this->getOwner()->getId() : $this->getOwner()->get($this->master_field);
+        $model = $this->getOwner();
+        $query = $this->dsql();
```
PhilippGrashoff, line 254: I dont see the use of putting these 2 in variables. Whats the idea behind this? Better code readability?

DarkSide666: Yeah, this is not functional change. Just made it a bit more similar to other methods above :)

PR #843, opened 2021-03-22
Remove ui property from UserAction

mvorisek:
finish after https://github.com/atk4/ui/pull/1550

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -16,9 +16,6 @@ parameters:
     ignoreErrors:
         - '~^Unsafe usage of new static\(\)\.$~'
 
-        # TODO remove once https://github.com/atk4/ui/issues/1548 is resolved
-        - '~^Class Atk4\\Ui\\(Button|Icon) not found\.$~'
-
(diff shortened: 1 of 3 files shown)


mvorisek:
@ibelar please finish this PR related to the ui PR

I tried to delete all `ui` props from this data repo, but there are quite a lot of uses and I am not sure if all of them can be removed...

mvorisek:
@ibelar these props
image
stay, right?

ibelar:
@mvorisek - I have removed UserAction ui property setting from Model.
I have also delete the $ui property from UserAction Class. Was not sure if we should keep it for backward compatibility or not and mark it as obsolete. If so, please add it back.

mvorisek:
> I have also delete the $ui property from UserAction Class. Was not sure if we should keep it for backward compatibility or not and mark it as obsolete. If so, please add it back.

removing is ok as it will go hand in hand with 2.4 ui rel

please approve/merge this PR, the changes are from you, but I am the initiating author and can not merge it...

ibelar (approved):
LGTM

PR #846, opened 2021-04-06
Deduplicate load methods and assert only one row can be loaded if expected

mvorisek:
no bc break

also add `Model::loadOne()` and `Model::tryLoadOne()` methods


src/Persistence/Sql.php
```diff
@@ -30,6 +30,9 @@ class Sql extends Persistence
     /** @const string */
     public const HOOK_BEFORE_DELETE_QUERY = self::class . '@beforeDeleteQuery';
 
+    /** @const string */
+    protected const ID_LOAD_ANY = self::class . '@idLoadAny';
```
mvorisek, line 34: better constant than `null` to prevent typos/errors (like unassigned variable...)

mvorisek: maybe we should support `one` (with checked only one record match) and `any` (one record of not strictly one)

src/Persistence/Sql.php
```diff
@@ -662,117 +662,52 @@ public function action(Model $model, string $type, array $args = [])
...
-    }
-
-    /**
-     * Tries to load any one record.
-     */
-    public function tryLoadAny(Model $model): ?array
-    {
-        $load = $model->action('select');
-        $load->limit(1);
+        $query->limit($id === self::ID_LOAD_ANY ? 1 : 2);
```
DarkSide666, line 678: smart move with limit(2) :)
I think that will not be very bad for performance if it tries to select 2 records instead of one.
Only question is - what if table will actually have only one record? Then it will not be able to detect issue here.

DarkSide666 (approved):
Quite a lot of changes, but LGTM

src/Persistence/Sql.php
mvorisek: yes, performance should be fine

> Only question is - what if table will actually have only one record? Then it will not be able to detect issue here.

that is expected and not only if table have not one record, but also if conditions match only one record :)

PR #848, opened 2021-04-10
Use strict tryLoadOne in ContainsOne

mvorisek:
to enforce stronger checks from https://github.com/atk4/data/pull/846

regex to find possible places to fix: `(try)?loadAny`

### DIFF
--- a/src/Reference/ContainsOne.php
+++ b/src/Reference/ContainsOne.php
@@ -100,8 +100,7 @@ class ContainsOne extends Reference
             });
         }
 
-        // try to load any (actually only one possible) record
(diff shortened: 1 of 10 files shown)


src/Reference/ContainsOne.php
```diff
@@ -100,8 +100,7 @@ public function ref(array $defaults = []): Model
             });
         }
 
-        // try to load any (actually only one possible) record
-        $theirModel->tryLoadAny();
+        $theirModel->tryLoadOne();
```
mvorisek, line 103: This is the main change

DarkSide666 (approved):
LGTM

PR #851, opened 2021-04-15
Update hintable dep to 1.2.x

mvorisek:
now we also allow only `1.2.*` instead of `1.*`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,13 +42,13 @@
     "php": ">=7.3.0",
     "ext-intl": "*",
     "atk4/dsql": "dev-develop",
-    "mahalux/atk4-hintable": "^1.1.4"
+    "mahalux/atk4-hintable": "~1.2.0"
   },
   "require-release": {
     "php": ">=7.3.0",
     "ext-intl": "*",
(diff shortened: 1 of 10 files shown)


georgehristov (approved):
LGTM

mvorisek:
@georgehristov any idea how to fix https://github.com/mvorisek/atk4-hintable-mirror/blob/develop/tests/Data/HintableModelTest.php#L102 ? (related with references and scopes)

currently it requires persistence set before ref() can be even returned

georgehristov:
> @ghristov#5652 any idea how to fix https://github.com/mvorisek/atk4-hintable-mirror/blob/develop/tests/Data/HintableModelTest.php#L102 ? (related with references and scopes)
>
> currently it requires persistence set before ref() can be even returned

@mvorisek you referenced somebody else :)

Anyway I will have a look

mvorisek:
@georgehristov did you find a solution?

update: link with fixed version/commit is: https://github.com/mvorisek/atk4-hintable-mirror/blob/1.2.1/tests/Data/HintableModelTest.php#L100

PR #853, opened 2021-04-15
Fix traversal when parent model is loaded

mvorisek:
finish after #855, #860, #861

discussion here: https://github.com/atk4/data/commit/c8cbce3e0ceceab616d990763a6b7f28abadded3#r49718142

fixes https://github.com/mvorisek/atk4-hintable-mirror/blob/1.2.1/tests/Data/HintableModelTest.php#L120

```
$model = new Model\Standard($db);
$model->load(2)->simpleMany->loadAny()->id;
```

must traverse from model with ID = 2 only (before this PR, ID = 2 condition was not applied and it was required to be added manually, now with entity concept we can assume locked ID)

note: this works also with `$model->load{One, Any}()->simpleMany` and also with `Model::duplicate()` which will unset the ID correctly.

### DIFF
--- a/src/Persistence/Array_.php
+++ b/src/Persistence/Array_.php
@@ -385,7 +385,15 @@ class Array_ extends Persistence
      */
     protected function applyScope(Model $model, Action $action): void
     {
-        $action->filter($model->scope());
+        $scope = $model->scope();
+
+        // add entity ID to scope to allow easy traversal
+        if ($model->id_field && $model->getId() !== null) {
+            $scope = new Model\Scope([$scope]);
+            $scope->addCondition($model->getField($model->id_field), $model->getId());
+        }
+
+        $action->filter($scope);
     }
 
     /**
--- a/src/Persistence/Sql.php
+++ b/src/Persistence/Sql.php
@@ -14,6 +14,8 @@ use Atk4\Dsql\Exception as DsqlException;
 use Atk4\Dsql\Expression;
 use Atk4\Dsql\Query;
 use Doctrine\DBAL\Platforms;
+use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Platforms\SQLServer2012Platform;
 
 class Sql extends Persistence
 {
@@ -369,6 +371,18 @@ class Sql extends Persistence
     public function initQueryConditions(Model $model, Query $query): void
     {
         $this->_initQueryConditions($query, $model->scope());
+
+        // add entity ID to scope to allow easy traversal
+        if ($model->id_field && $model->getId() !== null) {
+            $query->group($model->getField($model->id_field));
+            if ($this->getDatabasePlatform() instanceof SQLServer2012Platform
+                || $this->getDatabasePlatform() instanceof OraclePlatform) {
+                foreach ($query->args['field'] as $alias => $field) {
+                    $query->group(is_int($alias) ? $field : $alias);
+                }
+            }
+            $query->having($model->getField($model->id_field), $model->getId());
+        }
     }
 
     private function _initQueryConditions(Query $query, Model\Scope\AbstractScope $condition = null): void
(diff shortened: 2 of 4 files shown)


src/Persistence/Array_.php
```diff
@@ -332,7 +332,15 @@ protected function setLimitOrder(Model $model, \Atk4\Data\Action\Iterator $actio
      */
     public function applyScope(Model $model, \Atk4\Data\Action\Iterator $iterator)
     {
-        return $iterator->filter($model->scope());
+        $scope = $model->scope();
+
+        // add entity ID to scope to allow easy traversal
+        if ($model->id_field && $model->getId() !== null) {
```
mvorisek, line 338: should be probably deduplicated in Model

mvorisek:
@georgehristov I need your feedback on this

tests/RandomTest.php
```diff
@@ -380,7 +380,7 @@ public function testGetTitle()
 
         $m->load(2);
         $this->assertSame('Sue', $m->getTitle()); // loaded returns title_field value
-        $this->assertSame([1 => 'John', 2 => 'Sue'], $m->getTitles()); // all titles
+        $this->assertSame([2 => 'Sue'], $m->getTitles()); // traverse loaded
```
DarkSide666: I don't think this is correct. `getTitles()` should return names of all records taking applied conditions into account, but it's not related to is model loaded or not. It's like `$m->export([$m->title_field])`.

src/Persistence/Sql.php
```diff
@@ -364,12 +364,23 @@ protected function setLimitOrder(Model $model, Query $query)
     }
 
     /**
-     * Will apply a condition defined inside $condition or $model->scope() onto $query.
+     * Will apply $model->scope() conditions onto $query.
      */
-    public function initQueryConditions(Model $model, Query $query, Model\Scope\AbstractScope $condition = null): void
+    public function initQueryConditions(Model $model, Query $query): void
```
DarkSide666, line 371: Probably should be `_initQueryConditions` prefixed with underscore

DarkSide666 (changes requested):
Overall I don't think this is correct approach.
Loading record shouldn't add condition to model because that will limit how `getTitles()`, `export()` and other dataset-wise methods will work when model is or is not loaded. That shouldn't happen.

src/Persistence/Sql.php
mvorisek: why do you think so?

DarkSide666: Oh forget about this. I didn't notice other method below.

mvorisek:
All changes reverted and addressed differently, `getTitles()` and other confusing behaviour with entity (loaded model) now thow because of https://github.com/atk4/data/pull/855

src/Persistence/Sql.php
```diff
@@ -369,6 +369,12 @@ protected function setLimitOrder(Model $model, Query $query)
     public function initQueryConditions(Model $model, Query $query): void
     {
         $this->_initQueryConditions($query, $model->scope());
+
+        // add entity ID to scope to allow easy traversal
+        if ($model->id_field && $model->getId() !== null) {
+            $query->group($model->getField($model->id_field));
+            $query->having($model->getField($model->id_field), $model->getId());
```
mvorisek, line 384: FYI: `having` with `group by` is like `where`, but it does not apply after aggregate functions.

DarkSide666 (approved):
Approved in advance. Need to fix this approach for MS SQL

src/Persistence/Array_.php
```diff
@@ -185,38 +197,35 @@ public function add(Model $model, array $defaults = []): Model
...
-        if (!isset($this->data[$model->table])) {
-            throw (new Exception('Table was not found in the array data source'))
-                ->addMoreInfo('table', $model->table);
-        }
+        $table = $this->seedDataAndGetTable($model);
 
         if ($id === self::ID_LOAD_ONE || $id === self::ID_LOAD_ANY) {
-            if (count($this->data[$model->table]) === 0) {
+            $action = $this->action($model, 'select');
+            $action->generator->rewind(); // needed for some reasons!
```
mvorisek, line 204: why is this needed, should this be done in the action itself?

PR #855, opened 2021-04-18
No iterate over entity

mvorisek:
If you want to iterate over entity (loaded model), use `Model::duplicate` before `Model::export` or other iterating functions (like iterator in `foreach`).

### DIFF
--- a/tests/LimitOrderTest.php
+++ b/tests/LimitOrderTest.php
@@ -22,7 +22,7 @@ class LimitOrderTest extends \Atk4\Schema\PhpunitTestCase
         $i = (new Model($this->db, ['table' => 'invoice']))->addFields(['total_net', 'total_vat']);
         $i->addExpression('total_gross', '[total_net]+[total_vat]');
         $i->getField($i->id_field)->system = false;
-        $i->id_field = false;
+        $i->id_field = null;
 
         $i->setOrder('total_net');
         $i->onlyFields(['total_net']);
@@ -46,7 +46,7 @@ class LimitOrderTest extends \Atk4\Schema\PhpunitTestCase
         $ii = (new Model($this->db, ['table' => 'invoice']))->addFields(['total_net', 'total_vat']);
         $ii->addExpression('total_gross', '[total_net]+[total_vat]');
         $ii->getField($ii->id_field)->system = false;
-        $ii->id_field = false;
+        $ii->id_field = null;
 
         $i = clone $ii;
         $i->setOrder(['total_net' => 'desc', 'total_gross' => 'desc']);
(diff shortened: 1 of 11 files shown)


tests/LimitOrderTest.php
```diff
@@ -22,6 +22,7 @@ public function testBasic()
         $i = (new Model($this->db, ['table' => 'invoice']))->addFields(['total_net', 'total_vat']);
         $i->addExpression('total_gross', '[total_net]+[total_vat]');
         $i->getField($i->id_field)->system = false;
```
DarkSide666, line 24: these lines here and below probably are no more needed if you set id_field=null

tests/PersistentArrayOfStringsTest.php
```diff
@@ -46,7 +46,7 @@ public function testTypecasting()
             'array' => ['foo' => 'bar', 'int' => 123, 'rows' => ['a', 'b']],
             'object' => (object) ['foo' => 'bar', 'int' => 123, 'rows' => ['a', 'b']],
         ]);
-        $m->saveAndUnload();
+        (clone $m)->saveAndUnload();
```
DarkSide666, line 49: $m is not loaded here so why clone ?

tests/RandomTest.php
```diff
@@ -506,7 +508,7 @@ public function testDuplicateSaveNew()
         $db = new Persistence\Sql($this->db->connection);
         $m = new Model_Rate($db);
 
-        $m->load(1)->duplicate()->save();
+        (clone $m)->load(1)->duplicate()->save();
```
DarkSide666, line 497: why clone?

tests/RandomTest.php
```diff
@@ -399,7 +398,10 @@ public function testGetTitle()
         $m->title_field = 'my_name';
         $m->load(2);
         $this->assertEquals(2, $m->getTitle()); // loaded returns id value
-        $this->assertEquals([1 => 1, 2 => 2], $m->getTitles()); // all titles (my_name)
+
+        $this->expectException(Exception::class);
```
DarkSide666, line 402: No. `getTitles()` should work and if it needs duplicate, then that should be done in getTitles() or export() methods.

DarkSide666 (changes requested):

tests/LimitOrderTest.php
mvorisek: They are needed, only test does not pass - this does test `onlyFields`, but ID must be undefined if not flagged as a system. (normally, you keep it flagged as system field and this is not needed)

tests/PersistentArrayOfStringsTest.php
mvorisek: I kept `saveAndUnload` (`save` whould be enough), but we need clone because `save` makes model entity as well (which is consistent, as model is associated with DB, no matter if from laod or save)

tests/RandomTest.php
mvorisek: because later needed for a full export (which iterates over model and can not longer over entity (as entity = model restricted with one ID, thus iterating over one model does not make sense)) ;-)

DarkSide666 (approved):
LGTM

tests/RandomTest.php
mvorisek: image

PR #856, opened 2021-04-19
Remove Model::{asModel, saveAs, newInstance} methods

mvorisek:
The removed methods were more or less confusing...

### `Model::asModel` replacement:

see removed code, adjust for your usecase (that is why we removed this method, sometimes ID or default values should be copied, sometimes not)

### `Model::saveAs(string $class, array $options = [])` replacement:

```
$model->asModel($class, $options)->save();
```

### `Model::newInstance(string $class = null, array $options = [])` replacement:

```
new $class(null, $options);
```

This works even with `$class` beiing an object - https://3v4l.org/EfOiW

This PR also improve/simplify `Model::asModel()` method.

No change in `atk4/ui` repo needed.

### DIFF
--- a/docs/model.rst
+++ b/docs/model.rst
@@ -35,7 +35,7 @@ and even perform operations on multiple records (See `Persistence Actions` below
    $m->addCondition('expired', true);
 
    $m->action('delete')->execute(); // performs mass delete, hooks are not executed
-   
+
    $m->each(function () use ($m) { $m->delete(); }); // deletes each record, hooks are executed
 
 When data is loaded from associated Persistence, it is automatically converted into
@@ -434,17 +434,6 @@ This introduces a new business object, which is a sub-set of User. The new class
 inherit all the fields, methods and actions of "User" class but will introduce one new
 action - `send_gift`.
 
-There are some advanced techniques like "SubTypes" or class substitution,
-for example, this hook may be placed in the "User" class init()::
-
-   $this->onHookShort(Model::HOOK_AFTER_LOAD, function() {
-      if ($this->get('purchases') > 1000) {
-         $this->breakHook($this->asModel(VipUser::class);
-      }
-   });
-
-See also :php:class:`Field\\SubTypeSwitch`
-
 
 Associating Model with Database
 ===============================
@@ -498,14 +487,6 @@ Creates a duplicate of a current model and associate new copy with a specified
 persistence. This method is useful for moving model data from one persistence
 to another.
 
-.. php:method:: asModel($class, $options = [])
-
-Casts current model into another class. The new model class should be compatible
-with $this - you can do `$user->asModel(VipUser::class)` but converting `$user`
-into `Invoice::class` is a bad idea.
-
-Although class is switched, the new model will retain current record data, replace all
-fields/actions and will combine conditions (avoiding identical conditions).
 
 Populating Data
 ===============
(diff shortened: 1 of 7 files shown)


src/Model.php
```diff
@@ -1330,16 +1313,15 @@ public function saveAndUnload(array $data = [])
...
     {
-        $m = $this->newInstance($class, $options);
+        $m = new $class(null, $options);
 
         foreach ($this->data as $field => $value) {
-            if ($value !== null && $value !== $this->getField($field)->default && $field !== $this->id_field) {
-                // Copying only non-default value
-                $m->set($field, $value);
-            }
+            $m->set($field, $value);
```
DarkSide666: You can do simply `$m->set($this->data);` here without loop.
But I'm not sure why you removed condition lines above. There was some logic - to not copy null and default values and also of course not copy id.

src/Model.php
```diff
@@ -1350,24 +1332,7 @@ public function asModel(string $class, array $options = []): self
...
-     * as $this.
-     *
-     * @return static
-     */
-    public function newInstance(string $class = null, array $options = [])
-    {
-        $model = (self::class)::fromSeed([$class ?? static::class], $options);
-
-        if ($this->persistence) {
-            return $this->persistence->add($model); // @phpstan-ignore-line
```
DarkSide666, line 1362: This is the essence of newInstance(). At least for me.
It creates new model object and already links it with same persistence of old object.

mvorisek: I am aware, I can not be fixed in the constructor as construtor is not aware of the previous instance/persistence.

The solution is to call `new $class($model->persistence)`, eg. pass the persistence manually when needed.

Safety is handled, all DB operations throws if persistence is not set.

src/Model.php
mvorisek: - ignoring null - why? (I think too much crazy logic)
- ignoring v = default - saves only a call, nothin else :) (should be removed)
- ignoring ID - why (seems wrong to me, different model usually still targets the same data/table)

We would then need loadAs... This is too much methods for nothing.

docs/model.rst
```diff
@@ -434,17 +434,6 @@ This introduces a new business object, which is a sub-set of User. The new class
 inherit all the fields, methods and actions of "User" class but will introduce one new
 action - `send_gift`.
 
-There are some advanced techniques like "SubTypes" or class substitution,
```
DarkSide666, line 437: And this was quite a nice feature we used in one project :)

mvorisek: See code below, it tries to return different class from load method.

Notice, that a lot of custom code is needed already, you can adjust the code for the removed method easily. Usually, in the case below, you want to retain ID!

DarkSide666 (approved):
Is George around in Discord sometime? I would love to hear his opinion too on this.
I kind of understand that we can live without these methods (including asModel), but on other hand - why should we? Everything can be done manually in project-specific way, but it's also good if framework does some work for you (at least in one specific way and if you don't like that, then you can overwrite or extend).

So to wrap this up - I will approve this PR, but please try to discuss it also with George before merging. Maybe he can come up with better example of use-case then I could.

mvorisek:
@georgehristov Please approve as well, `asModel` is not good as the copy conditions depends on use case and all core functions must behave strictly from the name itself, which is not possible here. `newInstance` is more or less like an alias for constructor and confusing for the developer. Even for me, and I know a lot of internal things... We cleared this with @DarkSide666, but we want to make sure our decisions are aligned together. Thanks.

georgehristov (approved):
Ok, approving

PR #859, opened 2021-04-21
Fix next AI for array persistence

mvorisek:
merge after https://github.com/atk4/data/pull/710

originally AI was based on the max ID present, but if it was deleted, ID was lowered

now fixed to behave consistently like MySQL

### DIFF
--- a/src/Persistence/Array_.php
+++ b/src/Persistence/Array_.php
@@ -21,13 +21,14 @@ class Array_ extends Persistence
     /** @var array<string, array> */
     private $data;
 
-    /**
-     * Array of last inserted ids per table.
-     * Last inserted ID for any table is stored under '$' key.
-     *
-     * @var array
-     */
-    protected $lastInsertIds = [];
+    /** @var array<string, int> */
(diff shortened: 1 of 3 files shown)


src/Persistence/Array_.php
```diff
@@ -225,39 +230,37 @@ public function generateNewId(Model $model)
...
      */
     public function lastInsertId(Model $model = null)
     {
         if ($model) {
-            return $this->lastInsertIds[$model->table] ?? null;
+            return $this->lastInsertIdByTable[$model->table] ?? null;
         }
 
-        return $this->lastInsertIds['$'] ?? null;
+        return $this->lastInsertIdByTable[$this->lastInsertIdTable] ?? null;
```
DarkSide666, line 313: what if `lastInsertIdTable` is null (no inserts was done yet) ?

DarkSide666 (approved):

src/Persistence/Array_.php
mvorisek: image

PR #860, opened 2021-04-21
Array persistence store with strict objects

mvorisek:
There are minor BC breaks like renamed array Action class, but these are normally not used by the end user thus the changes should be BC, also no changes in atk4/ui required or it was a bug previously.

Fixes several bugs like non strictly rectangular data, Array persistence throw if update is called on non existing ID...

### DIFF
--- a/src/Persistence/Array_/Db/Table.php
+++ b/src/Persistence/Array_/Db/Table.php
@@ -6,13 +6,13 @@ namespace Atk4\Data\Persistence\Array_\Db;
 
 use Atk4\Data\Exception;
 
-class TableArray
+class Table
 {
     /** @var string Immutable */
     private $tableName;
     /** @var array<string, string> */
     private $columnNames = [];
-    /** @var array<int, RowArray> */
+    /** @var array<int, Row> */
     private $rows = [];
 
     public function __construct(string $tableName)
@@ -45,7 +45,7 @@ class TableArray
      */
     protected function assertValidValue($value): void
     {
-        if ($value instanceof self || $value instanceof RowArray) {
+        if ($value instanceof self || $value instanceof Row) {
             throw new Exception('Value can not be an ' . get_class($value) . ' object');
         } elseif (!is_scalar($value) && $value !== null) {
             throw (new Exception('Value must be scalar'))
@@ -108,7 +108,7 @@ class TableArray
         return isset($this->rows[$rowIndex]);
     }
 
-    public function getRow(int $rowIndex): RowArray
+    public function getRow(int $rowIndex): Row
     {
         if (!isset($this->rows[$rowIndex])) {
             throw (new Exception('Row with given index was not found'))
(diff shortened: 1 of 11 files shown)


tests/PersistentArrayTest.php
```diff
@@ -392,13 +403,13 @@ public function testConditions()
...
             2 => ['name' => 'ABC8', 'code' => 12, 'country' => 'Ireland', 'active' => 0],
-            3 => ['code' => 13, 'country' => 'Latvia', 'active' => 1],
+            3 => ['name' => null, 'code' => 13, 'country' => 'Latvia', 'active' => 1],
             4 => ['name' => 'ABC6', 'code' => 14, 'country' => 'UK', 'active' => 0],
             5 => ['name' => 'ABC5', 'code' => 15, 'country' => 'UK', 'active' => 0],
             6 => ['name' => 'ABC4', 'code' => 16, 'country' => 'Ireland', 'active' => 1],
             7 => ['name' => 'ABC3', 'code' => 17, 'country' => 'Latvia', 'active' => 0],
             8 => ['name' => 'ABC2', 'code' => 18, 'country' => 'Russia', 'active' => 1],
-            9 => ['code' => 19, 'country' => 'Latvia', 'active' => 1],
+            9 => ['name' => null, 'code' => 19, 'country' => 'Latvia', 'active' => 1],
```
mvorisek, line 412: this PR always make sure the data are rectangular, a positive side effect

src/Persistence/Array_/Db/Table.php
```diff
@@ -0,0 +1,196 @@
...
+    {
+        foreach ($newRowData as $columnName => $newValue) {
+            $this->assertValidValue($newValue);
+
+            // update index here
+        }
+    }
+
+    /**
+     * TODO rewrite with hash index support.
```
DarkSide666, line 182: Yes this is very important for performance

tests/JoinArrayTest.php
```diff
@@ -13,7 +13,18 @@ class JoinArrayTest extends AtkPhpunit\TestCase
 {
     private function getInternalPersistenceData(Persistence\Array_ $db): array
     {
-        return $this->getProtected($db, 'data');
+        $data = [];
+        /** @var Persistence\Array_\Db\Table $table */
+        foreach ($this->getProtected($db, 'data') as $table) {
+            foreach ($table->getRows() as $row) {
+                $rowData = $row->getData();
+                $id = $rowData['id'];
```
DarkSide666, line 21: hard-coded `'id'` field name here?

tests/StaticPersistenceTest.php
```diff
@@ -113,9 +113,9 @@ public function testFieldTypes()
             'name' => 'hello',
             'test_int' => 123,
             'test_float' => 123.45,
-            'test_date' => new \DateTime(),
-            'test_array' => ['a', 'b', 'c'],
-            'test_object' => new \DateInterval('P1Y'),
+            // 'test_date' => new \DateTime(),
```
DarkSide666, line 116: Array persistence no more support non-scalar data types?

DarkSide666 (approved):
LGTM, but few questions

tests/JoinArrayTest.php
mvorisek: yes - it is in specific test class, `id` here is safe to be assumed, the data are newly not indexed by some specific ID.

tests/StaticPersistenceTest.php
mvorisek: It fully does, but you must typecast, static persistence does not typecast seed data.

src/Persistence/Array_/Db/Table.php
mvorisek: Will come in some next PR, previously, the array data were always iterated with the select action anyway.

PR #861, opened 2021-04-22
Fix "field" action - return all records, allow Action as scope condition

mvorisek:
merge after https://github.com/atk4/data/pull/860

also fixes array persistence `tryLoad` which previously was completely ignoring all conditions (they were used only when loading data using iterator)

### DIFF
--- a/src/Persistence/Array_.php
+++ b/src/Persistence/Array_.php
@@ -201,7 +201,7 @@ class Array_ extends Persistence
 
         if ($id === self::ID_LOAD_ONE || $id === self::ID_LOAD_ANY) {
             $action = $this->action($model, 'select');
-            $action->generator->rewind(); // needed for some reasons!
+            $action->generator->rewind(); // TODO needed for some reasons!
 
             $selectRow = $action->getRow();
             if ($selectRow === null) {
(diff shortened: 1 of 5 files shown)


mvorisek:
@georgehristov can you please review the last commit?

mvorisek:
@georgehristov can you please review?

Do you know why https://github.com/atk4/data/pull/861/commits/7d598f19ce77a9356a1388e5e32a86002aa603b3#diff-facfb8206c512f6ada3a34d73c4ed464957c84987a77f986e2b8357bca34ab29R205 is needed to NOT check keys? The scope is for `refId` but the data have `id` keys. Are (our/their) column names/aliases set correctly in HasMany class?

src/Persistence/Array_.php
```diff
@@ -185,38 +197,35 @@ public function add(Model $model, array $defaults = []): Model
...
-        if (!isset($this->data[$model->table])) {
-            throw (new Exception('Table was not found in the array data source'))
-                ->addMoreInfo('table', $model->table);
-        }
+        $table = $this->seedDataAndGetTable($model);
 
         if ($id === self::ID_LOAD_ONE || $id === self::ID_LOAD_ANY) {
-            if (count($this->data[$model->table]) === 0) {
+            $action = $this->action($model, 'select');
+            $action->generator->rewind(); // needed for some reasons!
```
mvorisek: why is this needed, should this be done in the action itself?

DarkSide666 (approved):
LGTM, but would be good if @georgehristov could take a look too

PR #862, opened 2021-04-23
Load methods return a new instance (distinguish strictly between model and entity)

mvorisek:
# For 3.0 release 🚀 

# Definitions

### MODEL

Instance of `Model` class that is supposed to model data/relations, but can not hold/contain any record data.

### ENTITY

Instance of `Model` class that is supposed to hold/contain record data, but they can be modelled only using MODEL (described above).

The instance/object is locked with specific ID once it is loaded/inserted from/to DB.

# Migration

Model data like now. To manipulate with the actual data, you need to create another instance from MODEL using `createEntity()` method to create a new entity or using `{,try}load{,One,Any}()` methods which now return a new instance - ENTITY.

# Mistake safety

If and only if we can distinguish strictly between model and entity, then all wrong usage will throw and no code can keep working if used wrongly.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1930,43 +1930,36 @@ class Model implements \IteratorAggregate
      *
      * @param mixed $id
      *
-     * @return $this
+     * @return static
      */
     public function delete($id = null)
     {
-        if (!$this->isEntity()) {
-            return $this->createEntity()->delete($id);
-        }
-
-        if ($this->read_only) {
-            throw new Exception('Model is read-only and cannot be deleted');
-        }
-
-        if ($this->compare($this->id_field, $id)) {
-            $id = null;
-        }
-
         if ($id !== null) {
-            $thisCloned = clone $this;
-            $thisCloned->load($id)->delete();
+            $this->assertIsModel();
+
+            $this->load($id)->delete();
 
             return $this;
         }
 
-        return $this->atomic(function () {
-            if ($this->loaded()) {
-                if ($this->hook(self::HOOK_BEFORE_DELETE, [$this->getId()]) === false) {
-                    return $this;
-                }
-                $this->persistence->delete($this, $this->getId());
-                $this->hook(self::HOOK_AFTER_DELETE, [$this->getId()]);
-                $this->unload();
+        $this->assertIsEntity();
 
-                return $this;
-            }
-
+        if ($this->read_only) {
+            throw new Exception('Model is read-only and cannot be deleted');
+        } elseif (!$this->loaded()) {
             throw new Exception('No active record is set, unable to delete.');
+        }
+
+        $this->atomic(function () {
+            if ($this->hook(self::HOOK_BEFORE_DELETE, [$this->getId()]) === false) {
+                return;
+            }
+            $this->persistence->delete($this, $this->getId());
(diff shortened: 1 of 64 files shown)


mvorisek:
@georgehristov what do you think of this PR? Any idea why the one test is failing (delete should not set ID in any way)?

src/Model.php
```diff
@@ -1845,7 +1897,7 @@ public function atomic(\Closure $fx)
         try {
             return $this->persistence->atomic($fx);
         } catch (\Exception $e) {
-            if ($this->hook(self::HOOK_ROLLBACK, [$this, $e]) !== false) {
+            if ($this->hook(self::HOOK_ROLLBACK, [$e]) !== false) {
```
mvorisek, line 1938: this was wrong originally

tests/UserActionTest.php
```diff
@@ -206,15 +206,15 @@ public function testDisabled3()
 
     public function testFields()
     {
-        $client = new UaClient($this->pers);
-        $a = $client->addUserAction('change_details', ['callback' => 'save', 'fields' => ['name']]);
+        $m_client = new UaClient($this->pers);
+        $a = $m_client->addUserAction('change_details', ['callback' => 'save', 'fields' => ['name']]);
 
-        $client = $client->load(1);
+        $e_client = $m_client->load(1);
```
mvorisek, line 212: Please keep changes mininal, this is not consistent style with the other changes.

I will fix this, but please do not add more unneded refactoring changes.

src/Model/Scope/Condition.php
```diff
@@ -350,7 +350,7 @@ protected function valueToWords(Model $model, $value): string
         }
 
         // use the referenced model title if such exists
-        if ($field && ($field->reference ?? false) && $field->reference->getOwner()->isEntity()) {
+        if ($field && ($field->reference ?? false)) {
             // make sure we set the value in the Model parent of the reference
             // it should be same class as $model but $model might be a clone
             $field->reference->getOwner()->set($field->short_name, $value);
```
mvorisek: @georgehristov it is not ok to mutate Model from any scope code. What is the correct way to fix this code, eg. clone the reference incl. the parent? On cloned model, we can then call `createEntity()` easily and fix the test.

abbadon1334 (approved):
LGTM, really good and clean work.
IMHO it's okay to have an extra line in code to make sure that all the wrong usage are caught.

src/Model/UserAction.php
```diff
@@ -213,11 +216,20 @@ public function getConfirmation()
...
      */
     public function getModel(): Model
     {
-        return $this->getOwner();
+        return $this->entity ?? $this->getOwner();
+    }
+
+    public function setEntity(Model $entity): void
+    {
+        if ($entity->getModel() !== $this->getOwner() && PHP_MAJOR_VERSION > 0) { // TODO 2nd condition is for phpstan, see https://github.com/phpstan/phpstan/issues/4928
```
mvorisek: issue https://github.com/phpstan/phpstan/issues/4928

mvorisek:
@georgehristov please merge #870 and review this PR, v3.0 comming soon 🚀

PR #864, opened 2021-04-26
Make Model::{data,dirty,scope} props private

mvorisek:
for https://github.com/atk4/data/pull/862, we must have full control over these properties

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -153,7 +153,7 @@ class Model implements \IteratorAggregate
     public $persistence_data = [];
 
     /** @var Model\Scope\RootScope */
-    protected $scope;
+    private $scope;
 
     /**
      * Array of limit set.
@@ -185,7 +185,7 @@ class Model implements \IteratorAggregate
      *
      * @var array
      */
-    public $data = [];
+    private $data = [];
 
     /**
      * After loading an active record from DataSet it will be stored in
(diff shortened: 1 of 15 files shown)


src/Model.php
```diff
@@ -442,6 +442,16 @@ private function initEntityHooks(): void
         $this->onHookShort(self::HOOK_AFTER_SAVE, $checkFx, [], -10);
     }
 
+    public function &getDataRef(): array
```
DarkSide666, line 448: Don't we need to describe (comment) this and below method? It's quite straightforward, but still...

src/Model.php
```diff
@@ -620,7 +630,8 @@ public function isDirty(string $field): bool
     {
         $this->checkOnlyFieldsField($field);
 
-        if (array_key_exists($field, $this->dirty)) {
+        $dirtyRef = &$this->getDirtyRef();
+        if (array_key_exists($field, $dirtyRef)) {
```
DarkSide666, line 640: Here we can simplify to `return array_key_exists($field, $dirtyRef);` don't we?

DarkSide666: And this method looks exaactly the same as `_isset($field)` method below, so to avoid code duplication we could just use `return $this->_isset($field);` here.

DarkSide666 (approved):

src/Model.php
mvorisek: I will add `@internal` phpdoc, method is to easy migration for #862 before we remove this internal state method completely.

src/Model.php
mvorisek: Good catch, should be done later with another PR, probably after #862. The issue is, at least from my POV, why should be use dirty array for `_isset` instead of standard model data?

PR #865, opened 2021-04-28
Fix Field clone with Reference and do not mutate model scope in Condition::toWords

### LINKED ISSUE #867: Cloning Model having fields with reference
When cloning a Model, the fields are cloned as well but Field::$reference if set continues to hold Reference object owned by the original Model and not of the cloned Model.

e.g
```
$cloneModel = clone $originalModel;
// situation now
$cloneModel->getField('field_with_reference')->reference->getOwner() === $originalModel;

// proposed solution
$cloneModel->getField('field_with_reference')->getReference()->getOwner() === $cloneModel;
```

The solution would be to store only the Reference::$link under the Field::$reference (change to `protected`) and introduce Field::getReference() method:

```
public function getReference()
{
    return $this->getOwner()->getRef($this->reference);
}
```

mvorisek:
fixes #867 and https://github.com/atk4/data/blob/2.4.0/src/Model/Scope/Condition.php#L356 issue as well - Scope must not mutate Model in any sense, if something is desired to be set to gather data, it must be done on clone.

BC break:

`Field::reference` property was removed, replaced by `Field::getReference()` method.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -57,14 +57,12 @@ class Field implements Expressionable
     public $values;
 
     /**
-     * If value of this field can be described by a model, this property
-     * will contain reference to that model.
+     * If value of this field is defined by a model, this property
+     * will contain reference link.
      *
-     * It's used more in atk4/ui repository. See there.
-     *
-     * @var Reference|null
+     * @var string|null
      */
-    public $reference;
+    protected $referenceLink;
 
     /**
      * Actual field name.
@@ -530,6 +528,13 @@ class Field implements Expressionable
         };
 
         return $typecastFunc($value) === $typecastFunc($value2);
+    }
+
+    public function getReference(): ?Reference
+    {
+        return $this->referenceLink !== null
+            ? $this->getOwner()->getRef($this->referenceLink)
+            : null;
     }
 
     public function getPersistenceName(): string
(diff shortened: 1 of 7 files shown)


georgehristov:
I am looking into it but it will be a challenge.
This was a workaround for a bigger issue I see now.

georgehristov:
I added a quickfix now. The issue will be resolved when #867 is closed

mvorisek:
@georgehristov thanks, I fixed the Field cloning issue #867 as well as your have proposed.

src/Field.php
```diff
@@ -532,6 +530,13 @@ public function compare($value, $value2 = null): bool
         return $typecastFunc($value) === $typecastFunc($value2);
     }
 
+    public function getReference(): ?Reference
+    {
+        return $this->referenceLink !== null
+            ? $this->getOwner()->getRef($this->referenceLink)
```
mvorisek, line 536: I do not think we need to check if `getOwner()` is set, as Field with reference is created always with a model. If not, invoking `getOwner` will throw. :)

georgehristov (approved):

georgehristov:
Now we need to modify `data/ui`to use `Field::getReference()`

PR #868, opened 2021-04-29
Fix retrieval of Field reference from Model

### LINKED ISSUE #867: Cloning Model having fields with reference

georgehristov:
Fixes #867 

- introduces Field::hasReference / Field::getReference
- uses Field::$reference as link only

### DIFF
--- a/src-schema/Migration.php
+++ b/src-schema/Migration.php
@@ -224,7 +224,7 @@ class Migration
                 'mandatory' => ($field->mandatory || $field->required) && ($persistField->mandatory || $persistField->required),
             ];
 
-            $this->field($field->actual ?: $field->short_name, $options);
+            $this->field($field->getPersistenceName(), $options);
         }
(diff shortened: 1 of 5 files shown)


mvorisek:
@georgehristov already fixed in #865 - please review

PR #871, opened 2021-05-05
Upgrade phpstan to level 6

mvorisek:
## BC break

`Atk4\Schema\PhpunitTestCase::getMigrator()` method renamed to `Atk4\Schema\PhpunitTestCase::createMigrator()`


georgehristov (approved):
DBAL connection `getSchemaManager` is also renamed to `createSchemaManager`

PR #878, opened 2021-05-12
Test with limited max. connections

mvorisek:
related with https://github.com/sebastianbergmann/phpunit/issues/4705

`mysql_close` call seems to be not always synchronous, so use limit of 5

related with https://github.com/php/php-src/pull/5581

### DIFF
--- a/src-schema/PhpunitTestCase.php
+++ b/src-schema/PhpunitTestCase.php
@@ -72,7 +72,17 @@ class PhpunitTestCase extends AtkPhpunit\TestCase
 
     protected function tearDown(): void
     {
-        $this->db = null; // @phpstan-ignore-line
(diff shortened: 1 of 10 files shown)


mvorisek:
@georgehristov simultaneous connections count of 5 should be enough when connections are released properly, any idea how to fix this?

src-schema/PhpunitTestCase.php
```diff
@@ -36,42 +36,17 @@ protected function setUp(): void
...
-                    {
-                    }
-                }
-            );
-        }
     }
 
     protected function tearDown(): void
     {
+        $this->db->connection->connection->close();
```
mvorisek, line 49: needed for some reasons

abbadon1334 (approved):
LGTM

PR #880, opened 2021-05-24
HasOneSql::addField(): Use type from reference Model

PhilippGrashoff:
Hello,

currently, the type of a field is not taken from the referenced model (unlike Caption, this is already pulled from the referenced Model). This leads to:
```
//in ModelA
$this->addField('some_date', ['type' => 'date']);

//in ModelB
$this->hasOne('model_a_id', ['model' => [ModelA::class]]);
$this->getRef('model_a_id')->addField('some_date');
```
at this point, ModelBs field 'some_date' is of type string, not of type date as defined in ModelA. This PR would change this.


PhilippGrashoff:
RTM, can't assign labels

mvorisek (approved):

PR #886, opened 2021-08-27
Fix rel typo

mvorisek:


mvorisek:
closed in favor of https://github.com/atk4/data/pull/887

PR #888, opened 2021-09-15
Test with PHP 8.1

mvorisek:
merge once PHP 8.1 is stable

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "ext-intl": "*",
         "ext-pdo": "*",
         "atk4/core": "dev-develop",
-        "doctrine/dbal": "^2.10 || ^3.0",
+        "doctrine/dbal": "^2.13.3 || ^3.0",
         "mahalux/atk4-hintable": "~1.3.2"
     },
     "require-release": {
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -47,7 +47,7 @@
         "ext-pdo": "*",
         "atk4/core": "~3.1.0",
         "doctrine/dbal": "^2.10 || ^3.0",
-        "mahalux/atk4-hintable": "~1.3.1"
+        "mahalux/atk4-hintable": "~1.3.2"
```
DarkSide666, line 50: mahalux or mvorishek/atk4-hintable? I seen mvorisek/atk4-hintable in some other PR

DarkSide666 (approved):

composer.json
mvorisek: Repo renamed, mvorisek now, both are working - mahalux can not be actually deleted from Packagist as there were too many downloads already :D (nice feature of Packagist to keep repos available!)

mvorisek: but change will be part of that another PR

PR #891, opened 2021-09-29
Deduplicate Field properties

mvorisek:
`HasOne` is defining the same properties for seeding, deduplicate them. They were already out of date.

no BC break (except when `Field` seed properties are accessed from `HasOne` after `init()` on it was called)


abbadon1334 (approved):
LGTM

PR #894, opened 2021-09-29
Fix PR 891

mvorisek:
fix https://github.com/atk4/data/pull/891, serialize was used here as a fallback originally, atk4/ui `Scopebuilder` sets one field to a `Scope` object, which is probably wrong, but let's allow it for now. Issue is present only in Behat testing.

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -352,7 +352,9 @@ class Field implements Expressionable
                 return null;
             }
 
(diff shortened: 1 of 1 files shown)


DarkSide666 (approved):

mvorisek:
@ibelar do we really need to compare objects? It seems like a bug to me in a ScopeBuilder and this fix should not be needed.

PR #897, opened 2021-10-02
Fix field typecast and drop Boolean field class

mvorisek:
This fixes several bugs and always invoke DBAL typecast on non-generic/non-standard DBAL types.

Fixes several bugs like value of different type than `string` was casted to `null` which was wrong for types like `text`. There should be minimal hardcoded type specific handling.

BC break:

We also drop `Boolean` field class as it provided non-standard enum behaviour which should be provided by a custom DBAL type if really needed.


abbadon1334 (approved):
LGTM

PR #898, opened 2021-10-02
Require field to be present when typecasting a row

mvorisek:
Important for a security, we can not leave anything uncasted if not found.


mvorisek:
@georgehristov you know the data more internally, please help with this PR.

DarkSide666 (approved):
pre-approving

PR #899, opened 2021-10-03
Fix Field mandatory & normalize in row save typecast

### LINKED ISSUE #575: "mandatory" null validation is not working
See https://github.com/atk4/data/blob/1c5c361c2a8a34d0038fe9bf4e95f247b9f6997d/tests/FieldTest.php#L924
`b`/`mandatory` field should throw when null is set like the `c`/`required` field there.

As confirmed by @DarkSide666, the `set()` method is expected to fully normalize/validate the field at the time of setting the value (ie. immediatelly as `required` does).

Considering this as a Major as this is core functionality and many developers may rely on that.

### LINKED ISSUE #576: "mandatory" and "required" field properties need to be refactored
See https://github.com/atk4/data/pull/569#discussion_r414498175

Currently they are defined like:
- `mandatory` - prevent null
- `required` - prevent null and anything that evaluates with `empty()` to `true`

Currently there are two major issues:
- the names are not self explanatory
- two properties does not cover all use cases (we can not allow nullable non-empty field) and the values are not always valid (when `required` is set, `mandatory` can be false and can not be used for checking if the field can be `null` or not)
- the currently implementation does not even work like it is defined - see #575

I and @DarkSide666 propose to fix all these issues by refactoring this functionality to these fields:
- `notNull` - if true, field can not be set `null` or saved/exported when `null` is set by default
- `notEmpty` - if true, field can not be empty. Emptines should be evaluated to true when `(string)$v === ''` or for numerical types if `(float)$v === 0.0`.

With this proposal strings like ` ` (space) are considered non-empty. This is good to not solve multiple functionalities at once and it was so also originally as `empty(' ')` is evaluated to `false`. This should be fixed by normalization https://github.com/atk4/data/issues/577 which should be processed first.

These properties are very commonly used and therefore maximum care need to be provided to them.

mvorisek:
fixes #575
fixes #576

this also fixes atk4/ui to show `mandatory` error below form inputs (other errors like `required` are shown)

### DIFF
--- a/src/Persistence.php
+++ b/src/Persistence.php
@@ -170,10 +170,6 @@ abstract class Persistence
         foreach ($row as $fieldName => $value) {
             $field = $model->getField($fieldName);
 
-            // TODO move to value (instead of row) typecast
-            // TODO should we drop field normalization completely?
-            $value = $field->normalize($value);
-
             $result[$field->getPersistenceName()] = $this->typecastSaveField($field, $value);
         }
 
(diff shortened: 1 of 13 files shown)


mvorisek:
@DarkSide666 @georgehristov please guys comment on this change

src/Persistence.php
```diff
@@ -174,14 +170,11 @@ public function typecastSaveRow(Model $model, array $row): array
         foreach ($row as $fieldName => $value) {
             $field = $model->getField($fieldName);
 
-            $value = $this->typecastSaveField($field, $value);
+            // TODO move to value (instead of row) typecast
+            // TODO should we drop field normalization completely?
+            $value = $field->normalize($value);
```
mvorisek, line 175: TODO

DarkSide666 (approved):

PR #900, opened 2021-10-05
Adjust to CS fixer v3.2 rules

mvorisek:
also unify CI files with https://github.com/atk4/ui/pull/1675


src/Model.php
```diff
@@ -1382,7 +1382,7 @@ public function saveAndUnload(array $data = [])
      */
     public function withPersistence(Persistence $persistence, $id = null, string $class = null)
     {
-        $class = $class ?? static::class;
+        $class ??= static::class;
```
DarkSide666, line 1385: This is new syntax, right? Never noticed it before.

DarkSide666 (approved):

src/Model.php
mvorisek: shorthand by php, needed for default cs fixer config

PR #909, opened 2021-10-22
Fix table name source with Join

samsouthardjr:


mvorisek:
Code to reproduce:

```
<?php

use \Atk4\Data\Model;

require 'vendor/autoload.php';

class Par extends Model
{
    public $table = 'parent';
    
    {
        parent::init();
        $this->addField('name');
        $this->hasMany('children', ['model' => [Child::class]]);
    }
}

class Child extends Model
{
    public $table = 'child';
    
    {
        parent::init();
        $j = $this->join('jointable');
        $j->hasOne('parent_id', ['model' => [Par::class]]);
    }
}


$a = new App();

$p = new Par($a->db);
$e = $p->createEntity();
$e->save(['name' => 'foo']);
$e2 = $e->ref('children')->createEntity();
$e2->save();
$e->ref('children')->action('delete')->execute();
```

samsouthardjr:
> Code to reproduce:

Due to the join, the execute example is bad -- it will still produce an error.  A better example is:

```
foreach ($e->ref('children') as $c) {
    $c->delete();
}
```

src/Persistence/Sql/Query.php
```diff
@@ -279,9 +279,15 @@ protected function _render_table($add_alias = true): ?string
...
     }
+    
+    /**
+     * @var boolean $deAliasMainTable   Should we de-alias the main table in [where] and [having] clauses?
+     */
+    private $deAliasMainTable = false;
 
     protected function _render_table_noalias(): ?string
     {
+        $this->deAliasMainTable = true;
```
mvorisek, line 288: this relies on a fact that the `_render_table_noalias` is called first which is very bad assumption as it must not be always true like when WITH clause is used

IMHO we should check why some clauses use non-aliased table name in general

samsouthardjr:
Probably true – I’m a relatively new SQL person.  The alternative would seem to be something like having [noalias] in the initial part of the template definition, which resulted in no aliases used in the entire query.  However, I’m not expert enough to know if that would break something else (i.e., is it necessary to use aliases in deletes in sub queries, perhaps?).

 

However, the code example to repeat the bug does seem to demonstrate something seriously wrong.  It seems to be impossible to use ATK4 to delete a record if a join is involved.  That seems fairly serious.

src/Persistence/Sql/Query.php
```diff
@@ -671,7 +677,18 @@ protected function _sub_render_condition(array $row): string
             throw new \InvalidArgumentException();
         }
 
+        $oldMainAlias = $owner = null;
```
mvorisek:
```suggestion
        $owner = null;
        $oldMainAlias = null;
```

src/Persistence/Sql/Query.php
```diff
@@ -279,9 +279,15 @@ protected function _render_table($add_alias = true): ?string
 
         return implode(', ', $ret);
     }
+    
+    /**
+     * @var bool Should we de-alias the main table in [where] and [having] clauses?
+     */
```
mvorisek:
```suggestion
    /** @var bool Should we de-alias the main table in [where] and [having] clauses? */
```

src/Persistence/Sql/Query.php
samsouthardjr: Would a better version be to do something like changing a template to:
'[noalias][with]update [table_noalias] set [set] [where]' (and perhaps if [noalias] was at the beginning, [table_noalias] would be  redundant)?  With that syntax, we could just have a _render_noalias function that did nothing but set deAliasMaintable to true.

mvorisek:
How is the code supposed to work when there is no `jointable` defined?

here is a testable demo https://github.com/atk4/data/compare/fix_upd_join_w_alias#diff-92f78239a1e95b6ddebce64dec3b75d71cfa81ee4073bb544f9f8f491a37218cR736

please make it passing, eg. add extra 1:1 model /w `jointable`

samsouthardjr:
I’m not sure what you mean.  That code was a demonstration of the problem when a jointable is defined – not defining a jointable would completely miss the point.

 

Also, the “$deleteAction = $e->ref('children')->action('delete');” code at line 709/710 won’t work properly because of the join.  I just had it to force it to generate a SQL Query dump that I could see was causing the problem.  The correct code for a test would be

 

foreach ($e->ref(‘children’) as $child)

              $child->delete();

 

-- Sam

mvorisek:
these two lines from your example: https://github.com/atk4/data/compare/fix_upd_join_w_alias#diff-92f78239a1e95b6ddebce64dec3b75d71cfa81ee4073bb544f9f8f491a37218cR700-R701

create only `parent` and `child` table names only, add 3rd model with `jointable`

you can fork the branch from my experiment and you can see the actual queries executed - https://github.com/atk4/data/runs/4166653222?check_suite_focus=true

samsouthardjr:
Oh, I think I see the confusion.  Jointable refers to a database table, not a separate model.  Isn’t that the syntax for $this->join(), referring to a table, not a model?

 

In fact, the public table = ‘child’ in the Child class definition is unnecessary, and probably a red herring.  There is no ‘child’ table in the database.  Notice that I don’t actually have any $this->hasOne()s in Child’s init?

 

What is going on here is that I basically have a transaction table, with many different types of transactions.  All of the transactions have a common set of fields (think timestamp, user, etc.), but each type of transaction has a number of fields that are specific to only that kind of transaction.  The Par class would be for one of those kinds of transactions (so there would also be Par2, Par3, etc.).  But it’s a one-to-one relationship between Par and Child – so the hasMany in your Par model should be as hasOne.  I think that’s the confusion.

 

-- Sam

mvorisek:
> 
> Oh, I think I see the confusion. Jointable refers to a database table, not a separate model. Isn’t that the syntax for $this->join(), referring to a table, not a model?

yes, table, but we need a model to create the table using migrator...

mvorisek:
closing in favor of https://github.com/atk4/data/pull/784 , thank you for the proposed solution, the dealiasing condition is correct

PR #911, opened 2021-11-05
Normalize ID in array persistence and fix type guessing in static persistence

### LINKED ISSUE #907: Persistence\Array_ not allowing zero id, throwing error
1) Create a model from an array with `$m = new Model(new Array_($t));` where `$t= [
    ['name' => 'January', 'sales' => 20000, 'purchases' => 10000],
    ['name' => 'February', 'sales' => 23000, 'purchases' => 12000]]`

2) Iterate through model with `foreach ($model as $row) { }`

Atk4\Data throws an error on iteration as it does not accept id to be zero in \Atk4\Data\Field Line 311.
                    
`$value = (int) $value;
                    if ($this->required && empty($value)) {
                        throw new ValidationException([$this->name => 'Must not be a zero'], $this->getOwner());
                    }`

@mvorisek Probably due to stricter type validation, but that way Persistence\Array_ is broken.

mvorisek:
fixes #907 (as a side effect, invalid ID is now detected immediately)


DarkSide666 (approved):

src/Persistence/Static_.php
```diff
@@ -40,19 +40,23 @@ public function __construct(array $data = null)
...
-        });
-
         if (!is_array($row1)) {
             // convert array of strings into array of hashes
+            $allKeysInt = true;
             foreach ($data as $k => $str) {
                 $data[$k] = ['name' => $str];
+
+                if (!is_int($k)) {
+                    $allKeysInt = false;
```
PhilippGrashoff, line 50: A `break;` here would speed things up :)

PR #912, opened 2021-11-06
Deduplicate model properties and Field/Join from entity

mvorisek:
## BC break

`Field` is no longer bound to an entity, thus methods like `Field::get`, `Field::set` requires entity provided as the 1st argument like `$field->get($entity)`

BC/magic mapping for model properties and methods on entity is provided


src/Field.php
```diff
@@ -41,6 +43,16 @@ public function __construct(array $defaults = [])
         }
     }
 
+    /**
+     * @param Model $owner
+     */
+    public function setOwner(object $owner)
```
mvorisek, line 49: DEBUG only, Field should never be bound with an entity

src/Model/EntityField.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Atk4\Data\Model;
+
+use Atk4\Data\Field;
+use Atk4\Data\Model;
+
+class EntityField
```
mvorisek, line 10: test should be added

DarkSide666:
Not so sure about this. Looks like this makes working with models/entities even more complex than it was before.
I'll have to focus and think more about this before approving.

mvorisek:
> 
> 
> Not so sure about this. Looks like this makes working with models/entities even more complex than it was before. I'll have to focus and think more about this before approving.

internally a lot of `getModel(true)` calls must be added to places where model/entity usage is mixed (persistence/joins), from user perspective, there should be no ambiguity - entity should not provide functionality to operate on model

currently, any modifications to model properties on cloned to entity are wrong as this creates discrepancy between model and entity which is undesired as the should be single model/source of truth for any size of it's entities

src/Field.php
```diff
@@ -272,15 +288,10 @@ public function normalize($value)
     /**
      * Casts field value to string.
      *
-     * @param mixed $value Optional value
+     * @param mixed $value
      */
-    public function toString($value = null): string
+    public function toString($value): string
     {
-        $value = ($value === null /* why not func_num_args() === 1 */ ? $this->get() : $this->normalize($value));
```
mvorisek, line 279: FYI called in typecast too, no need to call twice...

DarkSide666 (approved):
OK let's go for it in 3.1

PR #914, opened 2021-11-09
Unify common methods

mvorisek:


src/Model.php
```diff
@@ -1889,7 +1864,11 @@ public function atomic(\Closure $fx)
     // {{{ Support for actions
 
     /**
-     * Execute action.
+     * Create persistence action.
+     *
+     * TODO Rename this method to stress this method should not be used by used
```
mvorisek, line 1869:
```suggestion
     * TODO Rename this method to stress this method should not be used
```

src/Model/Join.php
```diff
@@ -136,6 +137,13 @@ class Join
     public function __construct(string $foreign_table = null)
     {
         $this->foreign_table = $foreign_table;
+
+        // handle foreign table containing a dot - that will be reverse join
+        if (strpos($this->foreign_table, '.') !== false) {
+            // split by LAST dot in foreign_table name
+            [$this->foreign_table, $this->foreign_field] = preg_split('~\.+(?=[^.]+$)~', $this->foreign_table);
```
mvorisek, line 144: FYI `foreign_table` table is used as an index when added to a Model, must be set prior init

src/Model/Join.php
```diff
@@ -136,6 +137,13 @@ class Join
...
+            $this->reverse = true;
```
DarkSide666, line 145: Can we still overwrite $reverse property while initializing join?

src/Model/JoinsTrait.php
```diff
@@ -29,11 +31,19 @@ public function join(string $foreignTable, array $defaults = []): Join
     {
         $defaults[0] = $foreignTable;
 
-        return $this->add(Join::fromSeed($this->_default_seed_join, $defaults));
+        $join = Join::fromSeed($this->_default_seed_join, $defaults);
+
+        if ($this->hasElement($name = $join->getDesiredName())) {
```
DarkSide666, line 36: Can we join same table multiple times?

DarkSide666 (approved):

src/Model/Join.php
mvorisek: yes

src/Model/JoinsTrait.php
mvorisek: only one Join can be added which can be retrived by `getJoin`, same as for fields and references (even if the setup methods are named `hasOne`/`hasMany`)

PR #915, opened 2021-11-11
Deduplicate Reference from entity

mvorisek:
## BC break

`Reference` is no longer bound to an entity, thus methods `Reference::ref`, `Field::refModel`, `Field::refLink` requires entity provided as the 1st argument like `$ref->ref($entity)`. No change required if traversing from model/entity directly.


DarkSide666 (approved):
lgtm

PR #917, opened 2021-11-12
Add binary/blob column support for MSSQL and Oracle

mvorisek:
- fixes various inconsistencies so all DB vendors behave the same from the DSQL API
- adds full UTF-8 support for all DB vendors
- adds auto drop tables after each test


src/Persistence.php
```diff
@@ -45,20 +45,11 @@ public static function connect($dsn, string $user = null, string $password = nul
         $dsn = \Atk4\Data\Persistence\Sql\Connection::normalizeDsn($dsn, $user, $password);
 
         switch ($dsn['driverSchema']) {
+            case 'sqlite':
             case 'mysql':
-            case 'oci':
-            case 'oci12':
```
DarkSide666, line 50: For some reason which I don't remember now we had separated Oracle 12 support.

src/Persistence/Sql/Connection.php
```diff
@@ -212,6 +212,16 @@ protected static function connectDbalConnection(array $dsn)
         if (isset($dsn['pdo'])) {
             $pdo = $dsn['pdo'];
         } else {
+            $enforceCharset = [
+                'mysql' => 'utf8mb4',
+                'oci' => 'AL32UTF8',
+            ][$dsn['driverSchema']] ?? null;
+
+            if ($enforceCharset !== null) {
+                $dsn['dsn'] = preg_replace('~; *charset=[^;]+~i', '', $dsn['dsn'])
```
DarkSide666, line 221: So we always enforce utf8? What if user database is not utf8? Previously utf8 was used by default, but user could use different charset too.

src/Persistence/Sql/Oracle/Query.php
```diff
@@ -4,8 +4,24 @@
...
+        if ($this->mode === 'select' && $this->main_table === null) {
+            $this->table('DUAL');
+        }
+
+        return parent::render();
+    }
+
+    public function groupConcat($field, string $delimiter = ',')
+    {
+        return $this->expr('listagg({field}, []) within group (order by {field})', ['field' => $field, $delimiter]);
```
DarkSide666, line 22: Which is lowest Oracle version which supports LISTAGG?

tests/Persistence/Sql/WithDb/SelectTest.php
```diff
@@ -304,14 +267,34 @@ public function testExecuteException(): void
...
 
+    public function testUtf8mb4Support(): void
+    {
+        // remove once https://jira.mariadb.org/browse/MDEV-27050 is fixed
+        if (str_contains($_ENV['DB_DSN'], 'mariadb')) {
+            $this->markTestSkipped('MariaDB has broken support of utf8mb4 identifiers');
+        }
+
+        $this->assertSame(
+            ['❤' => 'žlutý_😀'],
```
DarkSide666, line 278: wow 🚀 :D

DarkSide666 (approved):
LGTM. Pls check comments.

src/Persistence.php
mvorisek: There was a different compatibility layer previously for Oracle 12, but I purged some time ago it in favor of generic layer for Oracle 11 and all tests are run againt this version in CI.

src/Persistence/Sql/Connection.php
mvorisek: Because of many hidden issues. UTF-8 is defacto standard. This is about the connection charset, you can store any data you want in binary fields, actually, this is one of the feature thi PR adds and fixes.

src/Persistence/Sql/Oracle/Query.php
mvorisek: 11g2 https://www.toolbox.com/tech/oracle/blogs/oracle-sql-listagg-in-oracle-9i-022118/

I belive we do not use it Model/Ref thus this version is not strictly required.

PR #928, opened 2021-11-23
Do not dump query twice on SQL exception

mvorisek:


src/Reference/HasOneSql.php
```diff
@@ -29,7 +28,9 @@ public function addField(string $ourFieldName, string $theirFieldName = null, ar
         $defaults['caption'] ??= $refModelField->caption;
         $defaults['ui'] ??= $refModelField->ui;
 
-        $fieldExpression = $ourModel->addExpression($ourFieldName, array_merge(
+        $defaults['ui'] = array_merge($defaults['ui'], ['editable' => false]);
```
mvorisek, line 31: hasOne field seems to be not saveable

PR #932, opened 2021-11-25
Add virtual IsEntity/IsLoaded interfaces for phpstan

mvorisek:
needs https://github.com/phpstan/phpstan/issues/7344 (implemented)
needs https://github.com/phpstan/phpstan/issues/7385 - https://phpstan.org/developing-extensions/type-specifying-extensions should be the workhorse
needs https://github.com/phpstan/phpstan/discussions/7386 (solution found)

TODO check if https://github.com/phpstan/phpstan-src/pull/2609/files can help

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -73,7 +73,7 @@ jobs:
         if: matrix.type == 'StaticAnalysis'
         run: |
           echo "memory_limit = 2G" > /usr/local/etc/php/conf.d/custom-memory-limit.ini
-          vendor/bin/phpstan analyse -v
+          vendor/bin/phpstan analyse -v --debug
 
   unit-test:
     name: Unit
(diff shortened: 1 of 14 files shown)


composer.json
```diff
@@ -39,15 +39,15 @@
         "ext-pdo": "*",
         "atk4/core": "dev-develop",
         "doctrine/dbal": "^2.13.5 || ^3.0",
-        "mvorisek/atk4-hintable": "~1.7.1"
+        "mvorisek/atk4-hintable": "~1.7.3 || dev-develop"
```
mvorisek, line 42: TODO remove before merge

mvorisek:
there must be a better way to do it - https://github.com/phpstan/phpstan/discussions/12101

mvorisek:
rebased from 1.x phpstan

Hard to decide, but we should probably not fake an interface as at least for final classes, this might resolve to never.

PR #933, opened 2021-11-26
`HasOneSQL::addField` does not support save (editable)

### LINKED ISSUE #929: `HasOneSql::addField` adds field without `->ui['editable'] = false`, but it is not saveable
field added using `HasOneSql` reference seems to be non-saveable, but currently if added using `HasOneSql::addField` it is NOT marked with `->ui['editable'] = false` thus in the main atk4/filestore demo `index.php`, the 2 imported fields (actually 4, imported twice) in https://github.com/atk4/filestore/blob/2.4.0/src/Field/File.php#L99 are shown in Form/CRUD and make user fake feel they can be edited. Currently, they are not saved. I have verified this again original atk4/filestore v2.4.0.

I belive `->ui['editable'] = false` should be added to the Field seed in `HasOneSql::addField` or make the field trully saveable from the reference side.

https://github.com/atk4/filestore/blob/develop/docs/images/crud.png historical (from version pre 2.4) screenshot confirms the fields are intended to be non-editable

Current behaviour:

image

mvorisek:
fixes https://github.com/atk4/data/issues/929

expected behaviour seems described in https://github.com/atk4/data/blob/5c411d2cd8430c84e7472219c0ec62909176d7ea/src/Model/FieldPropertiesTrait.php#L97-L105

but there are some hooks I do not fully understand their general purpose

https://github.com/atk4/data/blob/develop/src/Reference/HasOneSql.php#L49-L59

and `HasOne` class has these

https://github.com/atk4/data/blob/develop/src/Reference/HasOne.php#L77-L114

but these are not used for `HasOneSql::addField`, as field is created using `refLink` nad `ref` is not called at all

@georgehristov @DarkSide666 @ibelar what do you think, what should be the expected behaviour and how to resolve the issue above?

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -454,7 +454,7 @@ class Field implements Expressionable
      */
     public function isEditable(): bool
     {
-        return $this->ui['editable'] ?? !$this->read_only && !$this->never_persist && !$this->system;
+        return $this->ui['editable'] ?? !$this->read_only && !$this->never_persist && !$this->never_save && !$this->system;
     }
 
     /**
--- a/src/Reference/HasOneSql.php
+++ b/src/Reference/HasOneSql.php
@@ -28,6 +28,8 @@ class HasOneSql extends HasOne
         $defaults['caption'] ??= $refModelField->caption;
         $defaults['ui'] ??= $refModelField->ui;
 
+        $defaults['ui'] = array_merge($defaults['ui'], ['editable' => false]);
+
         $fieldExpression = $ourModel->addExpression($fieldName, array_merge(
             [
                 'expr' => function (Model $ourModel) use ($theirFieldName) {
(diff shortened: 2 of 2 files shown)


src/Field.php
```diff
@@ -450,7 +450,7 @@ public function getQueryArguments($operator, $value): array
      */
     public function isEditable(): bool
     {
-        return $this->ui['editable'] ?? !$this->read_only && !$this->never_persist && !$this->system;
+        return $this->ui['editable'] ?? !$this->read_only && !$this->never_persist && !$this->never_save && !$this->system;
```
mvorisek, line 457: what is the real difference between `read_only` and `never_save`?

mvorisek: @DarkSide666 @abbadon1334 @georgehristov any idea?

mvorisek: suggestion from chat:

```
return $this->ui['editable'] ?? !$this->read_only && !$this->system;
```

mvorisek: I tried the condition above and it does not fix the atk4/filestore issue described in https://github.com/atk4/data/issues/929

PR #935, opened 2021-11-26
Add WITH support for MSSQL

mvorisek:


src/Reference.php
```diff
@@ -200,7 +200,7 @@ public function createTheirModel(array $defaults = []): Model
                 $theirModel = ($this->model)($this->getOurModel(null), $this, $defaults);
             } else {
                 // if model is set, then use clone of this model
-                $theirModel = clone $this->model;
+                $theirModel = Factory::factory(clone $this->model, $defaults);
```
mvorisek, line 203: I would consider this as a bug

PR #940, opened 2021-12-03
Model only properties should be settable only from non-entity

mvorisek:


src/Field.php
```diff
@@ -380,14 +380,10 @@ public function compare($value, $value2): bool
         return $this->getValueForCompare($value) === $this->getValueForCompare($value2);
     }
 
-    public function getReference(Model $entity = null): ?Reference
+    public function getReference(): ?Reference
```
mvorisek, line 383: Reference is model only, accepting entity was non-sense

PR #946, opened 2021-12-31
Add support for Model nesting

mvorisek:
when saving, all inner models hooks are executed

no BC break

Array persistence support will be added in another PR - https://github.com/atk4/data/pull/961

### DIFF
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -182,6 +182,15 @@ class Join
         );
     }
 
+    private function getModelTableString(Model $model): string
+    {
+        if (is_object($model->table)) {
+            return $this->getModelTableString($model->table);
+        }
+
+        return $model->table;
+    }
+
     /**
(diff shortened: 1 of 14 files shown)


src/Model/Join.php
```diff
@@ -182,6 +182,15 @@ static function (Model $entity) use ($name): self {
         );
     }
 
+    private function getModelTableString(Model $model): string
+    {
+        if (is_object($model->table)) {
```
DarkSide666, line 187: maybe better `$model->table instanceof Model`

DarkSide666 (approved):
LGTM

src/Model/Join.php
mvorisek: answered on Skype, `is_object` is fine now as it is the only possible object type code path, Expression(able) in table prop is not supported

PR #948, opened 2022-01-01
Replace Model::persistence property with getter/setter

mvorisek:
previously `Persistence::add()` was not called when `Model::persistence` was set and it implied no `Persistence::HOOK_AFTER_ADD` hook was called

this PR fixes it by replacing `Model::persistence` property with persistence getter/setter

with the newly introduced persitence `Model::setPersistence()` setter, `Persistence:add()` call does not make any logical sense ( persistence can be set only once and never unset) and thus direct `Persistence:add()` call newly throws an exception

Here are several examples how to upgrade your code when setting persistence:

regex to locate old usage: `(?<!\w)(db|persistence)(?!\w).*->add\(`

```diff
 $m = new Invoice($db);
 // no change required
```

```diff
-$m = $db->add(new Invoice());
+$m = new Invoice($db);
```

```diff
 $m = new Invoice();
 ...
-$db->add($m);
+$m->setPersistence($db);
```

When accessing the persistence, `Model::getPersistence()` getter must be used and it throws when persistence is not set. To check if persistence is set, use `Model::issetPersistence()` method.

regex to locate old usage: `->persistence(?!\w)`

```diff
-$m = new Invoice($this->persistence);
+$m = new Invoice($this->getPersistence());
```


DarkSide666 (commented):
Looks good. Have to fix tests.

mvorisek:
what tests you mean?

PR #952, opened 2022-01-06
Add support for native MySQL driver (mysqli)

mvorisek:
to allow support of simultaneous queries :)

### DIFF
--- a/src/Persistence/Sql/Expression.php
+++ b/src/Persistence/Sql/Expression.php
@@ -493,6 +493,41 @@ class Expression implements Expressionable, \ArrayAccess
         return $arr;
     }
 
+    protected function hasNativeNamedParamSupport(): bool
+    {
+        return true;
+    }
+
+    /**
+     * @param array{string, array<string, mixed>} $render
+     *
+     * @return array{string, array<string, mixed>}
+     *
+     * @internal
+     */
+    protected function updateRenderBeforeExecute(array $render): array
+    {
+        [$sql, $params] = $render = $render;
+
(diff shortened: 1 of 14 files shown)


src/Persistence/Sql/Expression.php
```diff
@@ -493,6 +493,41 @@ public function __debugInfo(): array
...
+    /**
+     * @param array{string, array<string, mixed>} $render
+     *
+     * @return array{string, array<string, mixed>}
+     *
+     * @internal
+     */
+    protected function updateRenderBeforeExecute(array $render): array
+    {
+        [$sql, $params] = $render = $render;
```
DarkSide666: probably bug here

mvorisek: no bug, bug typo :)

mvorisek: fixed and cheked all other code with `\$(\w+) = \$\1(?!\w|\(|->|\[| \?)` regex

src/Persistence/Sql/Mysql/Query.php
```diff
@@ -16,6 +18,6 @@ class Query extends BaseQuery
 
     public function groupConcat($field, string $delimiter = ',')
     {
-        return $this->expr('group_concat({} separator [])', [$field, $delimiter]);
+        return $this->expr('group_concat({} separator \'' . str_replace('\'', '\'\'', $delimiter) . '\')', [$field]);
```
mvorisek, line 21: https://github.com/php/php-src/issues/7903

was actually a hidden issue

PR #960, opened 2022-01-13
Use Model for Join write operations

### LINKED ISSUE #955: Join class must not update with DSQL to trigger Model hooks
slighly related with https://github.com/atk4/data/issues/803

Join class should use Model instead of table directly

related with https://github.com/atk4/data/pull/946

mvorisek:
take advantage of https://github.com/atk4/data/pull/946 and https://github.com/atk4/data/pull/961

and fix https://github.com/atk4/data/issues/955 (model is still faked, or `Model\Join::createFakeForeignModel()` method needs to be overridden to return non-faked model with real hooks)


src/Model/Join.php
```diff
@@ -132,6 +131,44 @@ public function __construct(string $foreign_table = null)
         }
     }
 
+    /**
+     * Create fake foreign model, in the future, this method should be removed
+     * in favor of always requiring an object model.
+     */
+    protected function createFakeForeignModel(): Model
```
mvorisek, line 117: FYI if foreign model hooks needs to be executed, override this method to provide non-faked Model for `$this->foreign_table` table

src/Model/Join.php
```diff
@@ -31,8 +31,7 @@ class Join
     }
 
     /**
-     * Name of the table (or collection) that can be used to retrieve data from.
-     * For SQL, This can also be an expression or sub-select.
+     * Foreign model or WITH/CTE alias when used with SQL persistence.
      *
      * @var string
```
DarkSide666, line 35: hint

DarkSide666 (approved):
LGTM

PR #971, opened 2022-03-02
Connection DSN can omit password

DarkSide666:
We can have connection string with empty password and in such case password is not set in dsn array.
This should fix that.

For example
```
'dsn' => "mysql://root@localhost:3307/my_dbase;charset=utf8mb4",
```

### DIFF
--- a/src/Persistence.php
+++ b/src/Persistence.php
@@ -52,7 +52,7 @@ abstract class Persistence
             case 'pdo_sqlsrv':
             case 'pdo_oci':
             case 'oci8':
-                $persistence = new Persistence\Sql($dsn, $dsn['user'], $dsn['password'], $args);
+                $persistence = new Persistence\Sql($dsn, $dsn['user'] ?? null, $dsn['password'] ?? null, $args);
 
                 return $persistence;
             default:


src/Persistence.php
```diff
@@ -52,7 +52,7 @@ public static function connect($dsn, string $user = null, string $password = nul
             case 'pdo_sqlsrv':
             case 'pdo_oci':
             case 'oci8':
-                $persistence = new Persistence\Sql($dsn, $dsn['user'], $dsn['password'], $args);
+                $persistence = new Persistence\Sql($dsn, $dsn['user'] ?? null, $dsn['password'] ?? null, $args);
```
mvorisek, line 55: > connection string with empty password and in such case password is not set in dsn array

why `$dsn['user'] ?? null` then?

mvorisek: also, why not to use `http://username:@example.com` as DSN for empty password?

mvorisek:
PR LGTM, especially for sqlite DSN

src/Persistence.php
DarkSide666: ```http://username:@``` also don't set `password` key in dsn array and using `?? null` to avoid exception in case `password` (or user) keys are not set in dsn array.
Maybe we need to do this validation in `Connection` class instead and always pass user and password keys even if they are null (not set).

PR #972, opened 2022-03-04
Fix phpstan v1.4.7

DarkSide666:


tests/ContainsMany/Invoice.php
```diff
@@ -45,7 +45,7 @@ protected function init(): void
         $this->addCalculatedField($this->fieldName()->discounts_total_sum, function (self $m) {
             $total = 0;
             foreach ($m->lines as $line) {
-                $total += $line->total_gross * $line->discounts_percent / 100;
+                $total += (float) $line->total_gross * $line->discounts_percent / 100;
```
mvorisek: `total_gross` should be typed correctly instead

tests/ContainsMany/Invoice.php
```diff
@@ -12,7 +12,7 @@
  * @property string $ref_no              @Atk4\Field()
  * @property float  $amount              @Atk4\Field()
  * @property Line   $lines               @Atk4\RefMany()
- * @property string $total_gross         @Atk4\Field()
+ * @property float  $total_gross         @Atk4\Field()
```
mvorisek, line 15: the field type must be declared as well, this is for hintable only

PR #974, opened 2022-03-13
Fix long string param support for Oracle

mvorisek:
read needs https://github.com/php/php-src/pull/8018 for pdo_oci driver, oci8 driver is working

tested /w large value with size of 15 MiB, both text and blob DBAL types (~150 MB memory needed for Sqlite/Mysql, ~500 MB for Oracle and test took over 3 minutes to finish (vs. 1.5 second on Mysql))

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -77,7 +77,7 @@ jobs:
     name: Unit
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/mvorisek/image-php:${{ matrix.php }}
+      image: ghcr.io/mvorisek/image-php:${{ matrix.php }}-debian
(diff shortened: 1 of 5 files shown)


.github/workflows/test-unit.yml
```diff
@@ -77,7 +77,7 @@ jobs:
     name: Unit
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/mvorisek/image-php:${{ matrix.php }}
+      image: ghcr.io/mvorisek/image-php:${{ matrix.php }}-debian
```
mvorisek, line 82: Alpine support for newer pdo_oci is broken, see https://github.com/mlocati/docker-php-extension-installer/issues/523

PR #996, opened 2022-05-29
`Model::tryLoad` returns null when not found

mvorisek:
the return type was already documented as `null`, but the implementation returned unloaded entity

NOTE: `hintable` newly returns `null` when traversing entity 1:1 and the referenced record is not found - see https://github.com/mvorisek/atk4-hintable-mirror/commit/6a1863283b47afe20bb43f942544a8a7fd615460


mvorisek:
BC breaking change, but listing in MAJOR section CHANGELOG is enough

PR #1002, opened 2022-05-30
Strip schema from table for HasMany::their_name guess

### LINKED ISSUE #998: hasMany referencing a table in another db generates ciritcal error from lack of fallback field
With two classes:

```
class Parent extends Model
{
   public $table = 'db1.parent';

   {
      parent::init();
      $this->addField('name');
      $this->hasMany('children', ['model' => [Child::class]]);
    }
}

class Child extends Model
{
   public $table = 'db2.child';

   {
      parent::init();
      $this->addField('name');
      $this->hasOne('parent_id', ['model' => [Parent::class]]);
   }
}
```

With those classes, the code:

```
$p = new Parent($db);
$c = $p->ref('children');
```

will generate an exception "Atk4\Data\Exception: Their model does not contain fallback field" as it will attempt to reference a their_fallback_field of 'db1.parent_id' (it should be simply 'parent_id').

Adding "their_field' => 'parent_id' to the hasMany definition resolves this, but shouldn't be necessary

samsouthardjr:
fixes https://github.com/atk4/data/issues/998

### DIFF
--- a/src/Reference/HasMany.php
+++ b/src/Reference/HasMany.php
@@ -29,7 +29,11 @@ class HasMany extends Reference
         // this is pure guess, verify if such field exist, otherwise throw
         // TODO probably remove completely in the future
         $ourModel = $this->getOurModel(null);
-        $theirFieldName = $this->getModelTableString($ourModel) . '_' . $ourModel->id_field;
+        $theirFieldName = $this->getModelTableString($ourModel);
+        if (($dbSepPos = strpos($theirFieldName, '.')) === false) {
+            $dbSepPos = -1;
+        }
+        $theirFieldName = substr($theirFieldName, $dbSepPos + 1) . '_' . $ourModel->id_field;
         if (!$this->createTheirModel()->hasField($theirFieldName)) {
             throw (new Exception('Their model does not contain fallback field'))
                 ->addMoreInfo('their_fallback_field', $theirFieldName);
--- a/src/Reference/HasMany.php
+++ b/src/Reference/HasMany.php
@@ -29,11 +29,7 @@ class HasMany extends Reference
         // this is pure guess, verify if such field exist, otherwise throw
         // TODO probably remove completely in the future
         $ourModel = $this->getOurModel(null);
-        $theirFieldName = $this->getModelTableString($ourModel);
-        if (($dbSepPos = strpos($theirFieldName, '.')) === false) {
-            $dbSepPos = -1;
-        }
-        $theirFieldName = substr($theirFieldName, $dbSepPos + 1) . '_' . $ourModel->id_field;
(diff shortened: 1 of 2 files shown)


src/Reference/HasMany.php
```diff
@@ -29,7 +29,11 @@ public function getTheirFieldName(): string
         // this is pure guess, verify if such field exist, otherwise throw
         // TODO probably remove completely in the future
         $ourModel = $this->getOurModel(null);
-        $theirFieldName = $this->getModelTableString($ourModel) . '_' . $ourModel->id_field;
+        $theirFieldName = $this->getModelTableString($ourModel);
+        if (($dbSepPos = strpos($theirFieldName, '.')) === false) {
+            $dbSepPos = -1;
+        }
+        $theirFieldName = substr($theirFieldName, $dbSepPos + 1) . '_' . $ourModel->id_field;
```
mvorisek:
```suggestion
        $theirFieldName = preg_replace('~^.+?\.~', '', $this->getModelTableString($ourModel)) . '_' . $ourModel->id_field;
```

mvorisek: is such guessing present solely for HasMany?

tests/ReferenceTest.php
```diff
@@ -115,4 +115,15 @@ public function testCustomRef(): void
...
+
+    public function testOtherDbFieldRef(): void
+    {
+        $user = new Model($this->db, ['table' => 'thisdb.user']);
+        $order = new Model($this->db, ['table' => 'otherdb.orders']);
+        $order->addField('user_id');
+
+        $user->hasMany('Orders', ['model' => $order, 'caption' => 'My Orders']);
+        $user->ref('Orders');
+        $this->assertSame('noexception', 'noexception');
```
mvorisek: assert `$user->getRef('Orders')->getTheirFieldName()`

src/Reference/HasMany.php
samsouthardjr: I keep reminding myself to get better at regex.....and I was so focused on eliminating the exception that I never noticed getTheirFieldName being public so that I could use it in the tests.

Both items have been addressed.  I did check, and only found that issue in HasMany -- it wouldn't apply to HasOne, which specifies our field name directly and -- unless you say otherwise -- uses the Id field of the other Model, so there's no guessing involved.  The comments in getTheirFieldName seem to indicate that someone doesn't like the guessing at all, and that it may eventually go away.  From my point of view, though, well-documented standards (i.e., using 'user_id' as the default field name for the id of a record in the User model) that make people's lives easier are a good thing, as long as they don't inhibit functionality (which this doesn't, since you can always override the field if you want to).

The pull request has been updated and passes all tests (the smoke test issues in other code remains, but that's from items not in this pull request).

mvorisek:
Thank you

PR #1003, opened 2022-05-31
Strip schema from table for Join::foreign_field guess

mvorisek:
related with https://github.com/atk4/data/pull/1002, at least Join probably needs a simillar fix

/cc @samsouthardjr

### DIFF
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -351,12 +351,6 @@ abstract class Join
      */
     public function hasMany(string $link, array $defaults = [])
     {
-        $id_field = $this->getOwner()->id_field;
-        $defaults = array_merge([
-            'our_field' => $id_field,
-            'their_field' => $this->getModelTableString($this->getOwner()) . '_' . $id_field,
-        ], $defaults);
-
         return $this->getOwner()->hasMany($link, $defaults);
     }
 
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -103,6 +103,8 @@ abstract class Join
         $this->foreign_table = $foreignTable;
 
         // handle foreign table containing a dot - that will be reverse join
+        // TODO this table split condition makes JoinArrayTest::testForeignFieldNameGuessTableWithSchema test
+        // quite unconsistent - drop it?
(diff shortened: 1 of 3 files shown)


samsouthardjr:
> related with #1002, at least Join probably needs a simillar fix
> 
> /cc @samsouthardjr

I'm not an expert at Join or github, but you're right.  The exact same fix should be made at line 357 of Join.php.  Are you going to put that into this branch, or were you looking for me to do so?  I've never contributed to a pull request that someone else has started, but if you're looking to me to do it I can always try and learn.

mvorisek:
Technically you can open a PR into `join_name_guess_from_table_with_schema` branch, but I fixed it - the only thing to fix is the test and you can offer (such small) fix using a GH suggestion:

image

samsouthardjr:
> Technically you can open a PR into `join_name_guess_from_table_with_schema` branch, but I fixed it - the only thing to fix is the test and you can offer (such small) fix using a GH suggestion:
> 
> image

I couldn't find that button -- is it in githubdesktop, maybe?

In any case, I'll just comment on it here, since it's your pull.  The current pull doesn't address that the same change is needed in Join's hasMany() method, at line 357.

src/Model/Join.php
```diff
@@ -238,7 +238,7 @@ protected function init(): void
             }
 
             if (!$this->foreign_field) {
-                $this->foreign_field = $this->getModelTableString($this->getOwner()) . '_' . $id_field;
+                $this->foreign_field = preg_replace('~^.+?\.~', '', $this->getModelTableString($this->getOwner())) . '_' . $id_field;
```
samsouthardjr, line 243: Needs the same change on line 357 of this file

mvorisek: yes! - fixed, I think the seeding on l357 was just a duplicated/unneeded code, so I removed it

mvorisek: > I couldn't find that button -- is it in githubdesktop, maybe?

available only in "Files changed" tab and near the changed lines only

mvorisek:
Impl. fixed, but Join should be dropped sooner or later with Model impl. to support full/nested modelling.

PR #1004, opened 2022-05-31
Forbid ContainsXxx unmanaged data modification

### LINKED ISSUE #881: ContainsMany does not fire ref model hook when set or save
Consider code below:

```
class Client extends \Atk4\Data\Model
{
    public $table = 'client';
    public $caption = 'Client';

    {
        parent::init();

        $this->addField('name');
        $this->containsMany('Accounts', ['model' => [Account::class]]);
    }
}

class Account extends \Atk4\Data\Model
{
    public $caption = ' ';

    {
        parent::init();

        $this->addField('email', [
            'required' => true,
            'ui' => ['multiline' => [Multiline::INPUT => ['icon' => 'envelope', 'type' => 'email']]],
        ]);
    }
}
````

When changing Accounts reference value, then the Account model hook should be fire. Actually id does not.

Ex:
```
$m = new Client($app->db);
// loading client that has 3 related account record.
$clientEntity = $m->load(6);

$ac = $clientEntity->get('Accounts');
array_pop($ac);
$clientEntity->set('Accounts', $ac);
$clientEntity->save();
```

Then, when setting Account or saving Client, the Account model hook should be fired.

### PR DESCRIPTION
**Pull request by mvorisek**
closes https://github.com/atk4/data/issues/881

we forbid the hidden feature to prevent unwanted behaviour (no hooks, normalization, ...)

when the direct unmanaged data modification is really wanted, simply add a field using `$m->addField()` instead of `$m->containsOne()` (or `$m->containsMany()`)

### DIFF
--- a/src/Reference/ContainsOne.php
+++ b/src/Reference/ContainsOne.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Atk4\Data\Reference;
 
+use Atk4\Data\Exception;
+use Atk4\Data\Field;
 use Atk4\Data\Model;
 use Atk4\Data\Persistence;
 use Atk4\Data\Reference;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Reference/ContainsOne.php [hunk 1]
```diff
@@ -61,6 +63,25 @@ protected function init(): void
...
+            $ourRef = $field->getReference();
+            if ($ourRef === null || $field->shortName !== $this->getOurFieldName() || $value === null) {
+                // this code relies on Field::$referenceLink set
+                // also, allowing null value to be set will not fire any HOOK_BEFORE_DELETE/HOOK_AFTER_DELETE hook
+                return;
+            }
+
+            foreach (array_slice(debug_backtrace(\DEBUG_BACKTRACE_IGNORE_ARGS), 1) as $frame) {
+                if (($frame['class'] ?? null) === static::class) {
+                    return null; // allow load/save from ContainsOne hooks
```
mvorisek:
```suggestion
                    return; // allow load/save from ContainsOne hooks
```

PR #1008, opened 2022-06-08
Add Model::executeCountQuery() method

mvorisek:
after a long discussion, I propose `Model::executeCountQuery` name

I do not want to name it `Model::getCount`, as such name might imply it is stored and/or no query is sent to the DB

it is not only shorter than `$m->action('count')->getOne()` but the value is also casted properly to an `integer`


mvorisek:
Another method name could be `loadCount`

PR #1009, opened 2022-06-09
Improve datetime types typecasting

mvorisek:
also drop `Field::persist_format` and `Field::persist_timezone` properties (`datetime` is always stored (and converted to) in UTC)

if a custom persist format/timezone is wanted, implement a custom persistence or add a custom/named DBAL type

### DIFF
--- a/src/Persistence.php
+++ b/src/Persistence.php
@@ -439,19 +439,18 @@ abstract class Persistence
                 $format = preg_replace('~(?<=H:i:s)(?![. ]*u)~', '.u', $format);
             }
 
-            if ($field->type === 'datetime') {
-                $value = \DateTime::createFromFormat('!' . $format, $value, new \DateTimeZone('UTC'));
-                if ($value !== false) {
+            $valueOrig = $value;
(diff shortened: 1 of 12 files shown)


src/Persistence.php
```diff
@@ -435,30 +434,23 @@ protected function _typecastLoadField(Field $field, $value)
...
 
-            if ($field->type === 'datetime') {
-                $value = \DateTime::createFromFormat($format, $value, new \DateTimeZone($field->persist_timezone));
-                if ($value !== false) {
+            $valueOrig = $value;
+            $value = \DateTime::createFromFormat('!' . $format, $value, new \DateTimeZone('UTC'));
+            if ($value !== false) {
+                if ($field->type === 'datetime') {
                     $value->setTimezone(new \DateTimeZone(date_default_timezone_get()));
                 }
```
mvorisek, line 447:
```suggestion
                } elseif ($field->type === 'date') {
                    // set time to 12:00 UTC so conversion to another TZ does not change the date
                    $value->setTime(12, 0);
                }
```

this can be interesting implementation, but it complicates assertions

PR #1011, opened 2022-06-24
Fix Model::loadXxx() /w limit/offset set

### LINKED ISSUE #1010: setLimit(1) not working
Steps to reproduce:
1. Select any model and add some conditions which result in multiple records. Optionally also setOrder.
2. Apply setLimit(1) to restrict to one record only
3. Then tryLoadOne()

I get this error: `Atk4\Data\Exception: Ambiguous conditions, more than one record can be loaded`
Code snippet:
```
$customermodel = new \Atk4\Data\Model\Customer($db);
$customermodel->addCondition(\Atk4\Data\Model\Scope::CreateOr(['phone', 'LIKE', '%'.$number.'%'], ['mobile', 'LIKE', '%'.$number.'%']));
$customermodel->setOrder(['id']);
$customermodel->setLimit(1);
$entity = $customermodel->tryLoadOne();
```

What should be expected:
If there was minimum one entity matching the conditions, $entity should contain that (first) one, otherwise $entity == null.

mvorisek:
fixes https://github.com/atk4/data/issues/1010

Array persistence does not need fix, as it wraps the Action

most tests should run /w all persistences but a lot of test redesign is needed

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -143,10 +143,10 @@ class Model implements \IteratorAggregate
     /** @var Model\Scope\RootScope */
     private $scope;
 
-    /** @var array Array of limit set. */
-    public $limit = [];
+    /** @var array */
+    public $limit = [null, 0];
 
-    /** @var array Array of set order by. */
+    /** @var array */
     public $order = [];
 
     /** @var array<string, array{'model': Model, 'mapping': array<int|string, string>, 'recursive': bool}> */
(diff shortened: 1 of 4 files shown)


mvorisek:
@mkrecek234 thank you for the report, the limit offset was also broken, now it should be fixed for both Array/SQL persistences

mvorisek:
before merge, I will also emulate support for MSSQL for limit = 0, all other DBs support it

PR #1038, opened 2022-07-24
Fix HasOneSql imported field update support

### LINKED ISSUE #26: Subsequently adding new files in separate fields not working
Steps to reproduce in the existing demo:
1. Upload a File in the "File" field.
2. Save the form
3. Then, upload a second file in the "File 2" field.
4. Save the form.

What happens: it sets the token of the first "File" field to null, and fills the second "File 2" field with the new filename.

Expected behaviour:
It should keep the "File" token, and only add another token to the "File 2" field.

### LINKED ISSUE #929: `HasOneSql::addField` adds field without `->ui['editable'] = false`, but it is not saveable

mvorisek:
fixes https://github.com/atk4/data/issues/929
closes https://github.com/atk4/data/pull/933
fixes https://github.com/atk4/filestore/issues/26

https://github.com/atk4/data/blob/6657685ec9684286415b4d84d6fcc1ba4f0da1b7/src/Reference/HasOneSql.php#L35-L50

hook code is simply wrong.

In https://github.com/atk4/data/commit/5d49235b39cb5dca6e50fff781737eebe7de09d7 the code was intended to be fixed for changed values for different than `our_field` field. This caused data corruptions like https://github.com/atk4/filestore/issues/26. The data corruption was hotfixed in https://github.com/atk4/data/pull/1037 but looking into this issue deeper, the hook should simply not be there. We cannot infer `our_field` values based on "imported"/"joined" fields, as these fields might not be unique and might represent totaly different relation.

The proposed fix maintains the feature for the setting the `our_field` value for a new record with our field equal to null, under this state, we can determine the foreign/their record based on the imported/joined field without the possibility to mismatch the our field. The relation to foreign/their record can still imply duplicates, but this is a decision by the developer.

Making the field read only after the record is saved will fix also https://github.com/atk4/data/issues/929.

The update issue is obvious in https://github.com/atk4/data/blob/3.1.0/tests/ReferenceSqlTest.php#L562 test. How on Earth can "first name" solely uniquely locate a foreign user record?

### DIFF
--- a/src/Field.php
+++ b/src/Field.php
@@ -141,7 +141,7 @@ class Field implements Expressionable
                 switch ($this->type) {
                     case null:
                     case 'string':
-                        $value = trim(preg_replace('~\r?\n|\r~', ' ', $value)); // remove all line-ends and trim
+                        $value = trim(preg_replace('~\r?\n|\r|\s~', ' ', $value)); // remove all line-ends and trim
 
                         break;
                     case 'text':
--- a/src/Model.php
+++ b/src/Model.php
@@ -1627,21 +1627,14 @@ class Model implements \IteratorAggregate
 
     protected function _insert(array $row): void
     {
-        // Find any row values that do not correspond to fields, and they may correspond to
-        // references instead
+        // find any row values that do not correspond to fields, they may correspond to references instead
         $refs = [];
         foreach ($row as $key => $value) {
-            // and we only support array values
-            if (!is_array($value)) {
+            if (!is_array($value) || !$this->hasReference($key)) {
                 continue;
             }
 
-            // and reference must exist with same name
-            if (!$this->hasReference($key)) {
-                continue;
-            }
-
-            // Then we move value for later
+            // then we move value for later
             $refs[$key] = $value;
             unset($row[$key]);
         }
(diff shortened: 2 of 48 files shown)


mvorisek:
This PR breaks filestore in a diffrent way - upload CRUD edit form cannot be even opened - fix it before merge!

mvorisek:
related historical change https://github.com/atk4/data/commit/5d49235b39cb5dca6e50fff781737eebe7de09d7

PR #1039, opened 2022-07-26
Fix connection close on test error

mvorisek:


mvorisek:
fixed by https://github.com/atk4/core/pull/359

PR #1051, opened 2022-08-15
Adjusted scope validation

mkrecek234:
The introduced scope validation is fine, but too strict:

We have often situations where through a Save command, the entity will leave the scope of the model. This is and should be handled with saveAndUnload. With the latest https://github.com/atk4/data/pull/1044 even $model->saveAndUnload will not permit for example the SoftDelete code example to mark an entity as 'is_deleted' => true. Because it drops out of the model and will result in a record not found.

I generally agree, but only for save, and not for situations where saveAndUnload is used , that means ->reloadAfterSave is false. Otherwise it is a lot of messing around if you let the user change values which make entities leave the model scope.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1510,7 +1510,7 @@ class Model implements \IteratorAggregate
 
     protected function validateEntityScope(): void
     {
-        if (!$this->getModel()->scope()->isEmpty()) {
+        if ($this->reloadAfterSave && !$this->getModel()->scope()->isEmpty()) {
             $this->getPersistence()->load($this->getModel(), $this->getId());
         }
     }


mkrecek234:
@mvorisek The more I think about it - isn't the ->reload call in this line  absolutely sufficient for scope check?
https://github.com/atk4/data/blob/86ee2097a0ab5f26a4979d0dd3d71f262e346c81/src/Model.php#L1604

Scope should prevail after all hooks, updates, deletes were done, and only if reloadAfterSave is true.

mvorisek:
> @mvorisek The more I think about it - isn't the ->reload call in this line absolutely sufficient for scope check?
> 
> https://github.com/atk4/data/blob/86ee2097a0ab5f26a4979d0dd3d71f262e346c81/src/Model.php#L1604
> 
> Scope should prevail after all hooks, updates, deletes were done, and only if reloadAfterSave is true.

The https://github.com/atk4/data/pull/1044 took me quite long time to figure out a solution that works with all pre/post update hooks. I belive the current code cannot be simplified or you will break the tests.

Currently, you already broke the tests, especially for insert. `ModelNestedSqlTest::*` tests are broken too, but these tests may need an adjustment when the persist/hook order is expected to be different.

mvorisek:
Closing as explained on Discord.

Issue #967, opened 2022-02-22
Saving an entity with a modified field which excludes them from model results in error


mkrecek234:
Steps to reproduce:
- You have a model wit emails "Email" and filter it, e.g. by addCondition(unread, true);
- Out of this model you edit an email, changing "unread" to false, and apply an entity->save(['unread' => false]);
- Saving would theoretically work, but as atk4/data by default reloads the entity after save, it is rolled back with an error because that entity no longer exists in the model.

Current workaround: change reaload_after_save of model to false.

Problem for developers:
Saving values to entity which might result that they no longer are part of a model with conditions happens regularly. 

Idea:
The reload will be performed to the model without conditions to ensure entity can be found.

mvorisek:
Model is currently a domain/scope, eg. violation any conditions must always throw/rollback and never allow to save out-of-scope model data.

What you need is to have a Model /wo such filtering conditions and then close it for filtering an unread messages. But when saving such message as read, you must use the unfiltered model or remove the filtering cond before save.

Please post also the traversing back and forth case which might be solveable, but here is nothing we can do.

The only thing I think we can do is introduce a flag to every condition marking if such condition is filtering one or a model/hard one, if non-hard, we can ignore such condition on save.

mvorisek:
Even if we will ignore conditions set outside Model::init(), it is not logical to be able to save a record that cannot be selected later

Typical usecase is softdelete. I belive such usecase, and answer/solution for you?, is to temporary relax/ignore such condition during save in try&finally (remove&restore) approach

mkrecek234:
Reopening as in Discord discussion. Also see branch https://github.com/atk4/data/tree/save_wo_scope

mvorisek:
I am closing this issue. We can discuss it under your PR. No need for 2 places.

PR #1054, opened 2022-08-18
Add native soft-delete support

mvorisek:
replaces #692 and #1058 (issue #967)

### DIFF
--- a/tests/RandomTest.php
+++ b/tests/RandomTest.php
@@ -91,7 +91,7 @@ trait ModelSoftDeleteTrait
 
             $entity->hook(Model::HOOK_AFTER_DELETE);
             $entity->breakHook(false); // this will cancel original Model::delete()
-        });
+        }, [], 100);
     }
 }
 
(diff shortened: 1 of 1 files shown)


tests/RandomTest.php
```diff
@@ -71,6 +75,83 @@ protected function init(): void
...
+    protected function initSoftDelete(): void
+    {
+        $this->addField('is_deleted', ['type' => 'boolean', 'nullable' => false, 'default' => false]);
+        $this->addCondition('is_deleted', false);
+        $this->onHook(Model::HOOK_BEFORE_DELETE, function (Model $entity) {
+            $softDeleteController = new ControllerSoftDelete();
+            $softDeleteController->softDelete($entity);
+
+            $entity->hook(Model::HOOK_AFTER_DELETE);
+            $entity->breakHook(false); // this will cancel original Model::delete()
```
mvorisek, line 93: an alternative could be to replace the persistence during update so the delete hook does not need to be broken

abbadon1334: Both solutions are internal so we can go ahead in this way and add this feature.

mvorisek: Not really, such changes can have effect on overrided delete method or user hooks, very important decision that needs literaly hours of very fine reasoning to decide this correctly.

PR #1058, opened 2022-08-25
Add weak conditions/save support

mkrecek234:
related initial issue #967

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1003,6 +1003,30 @@ class Model implements \IteratorAggregate
     }
 
     /**
+     * Same as addCondition but condition will be dropped on save event and not be enforced.
+     * Therefore, it is valid to save an entity where the saved entity will not comply to the
+     * weak condition anymore.
+     *
+     * Example:
+     *
+     * $messsageModel->addWeakCondition('unread', true);
+     * $messageModel->save(['unread' => false]);
+     *
+     * @param mixed $field
+     * @param mixed $operator
+     * @param mixed $value
+     *
+     * @return $this
+     */
+    public function addWeakCondition($field, $operator = null, $value = null)
+    {
+
+        $this->scope()->addCondition(...func_get_args(), true);
+
+        return $this;
+    }
+
+    /**
      * Adds WITH/CTE model.
      *
      * @return $this
@@ -1341,6 +1365,21 @@ class Model implements \IteratorAggregate
     }
 
     /**
+     * Try to reload model by taking its current ID, otherwise return null.
+     *
+     * @return $this
+     */
+    public function tryReload()
+    {
+        $id = $this->getId();
+        $this->unload();
+
+        $res = $this->_load(true, true, $id);
+
+        return $this;
+    }
+
+    /**
      * Keeps the model data, but wipes out the ID so
      * when you save it next time, it ends up as a new
      * record in the database.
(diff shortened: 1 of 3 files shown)


mvorisek:
Usecases? What is "weak" and when? Sometimes traversing should be weak, sometimes not, we should probably start with all possible usecases and then find a well understandable solution.

src/Model.php
```diff
@@ -1340,6 +1364,21 @@ public function reload()
         return $this;
     }
 
+    /**
+     * Try to reload model by taking its current ID, otherwise return null.
+     *
+     * @return $this
+     */
+    public function tryReload()
```
mvorisek, line 1372: What are the usecases? If a record is expected to be possibly deleted, you can always check with `tryLoad` or handle `reload` exception.

src/Model.php
```diff
@@ -1562,7 +1651,7 @@ public function save(array $data = [])
             if ($this->idField && $this->reloadAfterSave) {
                 $d = $dirtyRef;
                 $dirtyRef = [];
-                $this->reload();
+                $this->tryReload();
```
mvorisek, line 1654: no try - we want strict behaviour as much as possible, not to silently ignore possible issues

src/Model.php
mkrecek234: The use case is as follows - consider a Crud which displays unread messages with `addWeakCondition('unread', true)` . You click on edit, and change the unread flag to false. You click save. 
The crud then runs a normal ->save function. Since save allows that - in case a field is changed that moves it outside of a weak condition, the save command should only try to reload. If it is successful, the crud will be updated, if it is null, then the entity left the scope, and thus the same animation as in a delete function will be run.

src/Model.php
mkrecek234: This is not a possible issue but a designed situation - if you do not have any weak condition then we could think about throwing an exception and not null. I think we should not be too dogmatic, especially if we destroy then logics which are present in Ui.

src/Model.php
mvorisek: This is legit usecase 👍

src/Model.php
mvorisek: I do not understand this, please provide a testcase for it.

mkrecek234: I placed the tryReload for the usecase above (the crud remove unread message from scope on save) - so when a model saves something, the associated view gets a feedback if the entity is still there (in scope, so not null) or if it is gone - so save function only *tries* to reload, and if it is not successful it should return null.

mvorisek: I understand, but we cannot remove this check because of some need.

I have no solution is my head, maybe we need to add some support to ui for it.

In atk4/data, we have currently BEFORE_SAVE and AFTER_SAVE hooks, maybe we need some concept of try/finally SAVE hook.

In the past we discussed an immutable models. I am still not fully decided if a condition removal should be supported at all. I have no plan to disallow it in the near future, but to support the conditions removal fully, we need to implement https://github.com/atk4/data/issues/662 first.

If you want to pursue this feature, please start with a real Crud test in atk4/ui. If you need one or two atk4/data classes modified there, you can extend the atk4/data classes in atk4/ui. Please keep in mind, this feature should work also for other ui components like Multiline.

mkrecek234: Why not rund a proper reload if no weak condition exists - only if a weak condition exists, you only try to reload and return null of it left the scope.

mvorisek: Because we need to be able to rely on reload. If the record became out of scope, it can be some error as well, with `tryReload`, we can be left with old data.

mvorisek:
please make the CS/stan CI happy and unit testing to finish at least

this is good practice so we can focus on the real issues

mvorisek:
see
https://github.com/atk4/data/blob/aacc24708166f39c4595f829105dbccc4467739c/tests/LookupSqlTest.php#L112
I belive a weak approach is bad - any condition (added by an user) will still break the (currently commented) hook code. Maybe a solution can be to backup the conditions after init is invoked and allow it to be used like in https://github.com/atk4/data/pull/1054 `invokeCallbackWithoutUndeletedCondition`.

mvorisek:
closing in favor of #1054

PR #1060, opened 2022-08-31
Xxx::addFields() input must use key for name

mvorisek:
still not sure if `addFields` methods should be presented or not, but they will stay at least for v4.0 - related PR https://github.com/atk4/data/pull/990

### DIFF
--- a/docs/model.rst
+++ b/docs/model.rst
@@ -230,13 +230,15 @@ Read more about :php:class:`Field`
 
 Creates multiple field objects in one method call. See multiple syntax examples::
 
-    $m->addFields(['name'], ['default' => 'anonymous']);
+    $m->addFields([
+        'name' => [],
+    ], ['default' => 'anonymous']);
 
     $m->addFields([
-        'last_name',
+        'last_name' => [],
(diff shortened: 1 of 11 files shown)


mvorisek:
@mkrecek234 AFAK you use and was defending `addFields` method, it stays, wdyt about this change, do you support the consistency?

mkrecek234:
> @mkrecek234 AFAK you use and was defending `addFields` method, it stays, wdyt about this change, do you support the consistency?

I dislike the ['name' => [], 'company' => []] - we want to still ensure that Atk4 remains agile in the sense that with few lines of code you can have a working prototype. So we should keep: addFields(['name', 'company', ... ]) functionality without the need to configure it.

mvorisek:
the original input is problematic, as it does too much transformations to allow us to describe the type for phpstan

also, for SQL HasOne, it allowed the their field to be specified as the 0 => value, but this is breaking seed definition, this key is reserved for the seed class name

mvorisek:
Maybe someday, thank you @mkrecek234 for the feedback.

PR #1061, opened 2022-09-01
Fix phpstan v1.8.3

mvorisek:


src/Model/UserAction.php
```diff
@@ -44,10 +44,10 @@ class UserAction
     /** @var string How this action interact with record */
     public $modifier;
 
-    /** @var \Closure(object, mixed ...$args): mixed|string code to execute. By default will call entity method with same name */
+    /** @var \Closure|string code to execute. By default will call entity method with same name */
```
mvorisek, line 47: needs phpstan fix: https://phpstan.org/r/aa28e93d-b05a-446b-b167-3b5a19b2a8b5

PR #1070, opened 2022-09-30
Model::addUserAction should accept seed /w class

mvorisek:
someday maybe also for hasOne and simillar functions

regex: `function(?! (__construct|add|ref|refmodel|reflink|createTheirModel|addToPersistence|makecsvpersistence|connect)\()((?! use ).)*\$defaults`

### DIFF
--- a/src/Model/UserActionsTrait.php
+++ b/src/Model/UserActionsTrait.php
@@ -19,22 +19,23 @@ trait UserActionsTrait
      * Register new user action for this model. By default UI will allow users to trigger actions
      * from UI.
      *
-     * @param array<string, mixed>|\Closure $defaults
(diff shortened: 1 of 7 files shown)


src/Model/UserActionsTrait.php
```diff
@@ -19,23 +19,18 @@ trait UserActionsTrait
...
     {
         $this->assertIsModel();
 
-        if ($defaults instanceof \Closure) {
-            $defaults = ['callback' => $defaults];
+        if ($seed instanceof \Closure) {
+            $seed = ['callback' => $seed];
         }
 
-        if (!isset($defaults['caption'])) {
```
mvorisek, line 32: it seems the reasons are purely historical and not intentional/needed now

Issue #803, opened 2020-12-10
Join class does not support joining tables by non-id fields


DarkSide666:
Currently we do not support joining tables by non-id fields because we don't know how to save records in such case.

See #802 and "Joining tables on non-id fields is not implemented yet" exception in `Join` class.

mvorisek:
@DarkSide666 can you please provide a minimalistic code to reproduce?

mvorisek:
here is a repro code: https://github.com/atk4/data/blob/f89264b0b7667baacef0aad420b421907d1abe7d/tests/JoinSqlTest.php#L32-L36

mvorisek:
Currently there are these problems:

- `Join::reverse = true` with join on non-ID field is not supported (`Joining tables on non-id fields is not implemented yet` exception is thrown)
- foreign ID field type is hardcoded to `integer`
- join ID field type is hardcoded to `integer`

PR #1077, opened 2022-12-15
Add join support for tables with PK different than "id"

abbadon1334:
related with #803


abbadon1334:
What do you think about the naming and description of this, because that 'idField' is for the FakeModel, not the table itself : 
https://github.com/atk4/data/blob/2426c4d33836a6088f5ebbf1a94d197cf6aa4ade/src/Model/Join.php#L89-L93

i decided to make it nullable because in method `setDefaults($defaults)` nulled values will not be set

src/Model/Join.php
```diff
@@ -85,6 +85,13 @@ abstract class Join
     /** @var array<int, array<string, mixed>> Data indexed by spl_object_id(entity) which is populated here as the save/insert progresses. */
     private array $saveBufferByOid = [];
 
+    /**
+     * Custom definition if $idField for of ForeignModel
+     * By Default ForeignModel idField is equal to 'id'
+     * To be used when the joined table doesn't have a field named 'id'
+     */
+    public ?string $foreignModelIdField = null;
```
mvorisek:
```suggestion
    public ?string $foreignIdField = null;
```

mvorisek: for consistency and move it after `public ?string $foreignField = null;`

tests/JoinSqlTest.php
```diff
@@ -762,4 +763,230 @@ public function testJoinActualFieldNamesAndPrefix(): void
...
+
+        $user->delete(1);
+
+        static::assertSame(null, $master_model->tryLoad(1));
+        static::assertSame(null, $joined_model->tryLoad(1));
+    }
+
+    public function testJoinDeleteForeignCustomIdFieldReverse(): void
+    {
+        $master_model = new \Atk4\Data\Model($this->db, [
```
mvorisek: fix import, use camelCase for variable names

tests/JoinSqlTest.php
```diff
@@ -762,4 +763,230 @@ public function testJoinActualFieldNamesAndPrefix(): void
             ],
         ], $this->getDb());
     }
+
+    public function testJoinSavingForeignCustomIdFieldRaiseException(): void
+    {
+        $this->expectException(\Atk4\Data\Persistence\Sql\ExecuteException::class);
```
mvorisek: always as last-1 line

tests/JoinSqlTest.php
```diff
@@ -762,4 +763,230 @@ public function testJoinActualFieldNamesAndPrefix(): void
...
+
+        $joined_model = new \Atk4\Data\Model($this->db, [
+            'table' => 'contact_join',
+            'idField' => 'not_named_id',
+        ]);
+        $joined_model->addField('contact_id');
+        $joined_model->addField('contact_phone');
+
+        $this->createMigrator($master_model)->create();
+        $this->createMigrator($joined_model)->create();
```
mvorisek: add foreign key, see other tests how, this will detect correct SQL insert/update/delete order

tests/ConditionSqlTest.php
```diff
@@ -286,6 +286,63 @@ public function testExpressionJoin(): void
         static::assertSame('+123 smiths', $mm2->get('contact_phone'));
     }
 
+    public function testExpressionJoinForeignCustomIdField(): void
```
mvorisek: does this test adds anything to the coverage or how it is helpfull?

abbadon1334: Nope, removed

tests/JoinSqlTest.php
abbadon1334: `$joinedModel` created at the start of the test, is only a model and is used to make the import and the final assert.
The `createForeignKey` is added a few lines below on the joined model

mvorisek:
Does this PR really solve https://github.com/atk4/data/issues/803? The issue describes something else, this PR adds support for PK to be different than "id". But if PK is "id" and we join based on a different column, the foreign model still needs to have the knowledge about PK to be "id", for AI (autoincrement) to work at least (not mentioning fake model will never execute proper model hooks).

abbadon1334:
> Does this PR really solve #803? The issue describes something else, this PR adds support for PK to be different than "id". But if PK is "id" and we join based on a different column, the foreign model still needs to have the knowledge about PK to be "id", for AI (autoincrement) to work at least (not mentioning fake model will never execute proper model hooks).

You are right about the hooks, it is a FakeModel so the only functionality of this PR is to build a correct query to avoid raising a DSQL Exception when an update on joined fields occurs, but no hook will be triggered.

So join can be done directly with models? ( like this `$model->join($joinedModel)` )

mvorisek:
that is the direction, see https://github.com/atk4/data/pull/960, however Join class still relies on many hooks as we must hold a reference to joined data and that has limitations currently, imagine like same entity joined/referenced back in one model

abbadon1334:
> that is the direction, see #960, however Join class still relies on many hooks as we must hold a reference to joined data and that has limitations currently, imagine like same entity joined/referenced back in one model

Ok, I will do some tests.

mvorisek:
In this PR, let's focus on the added `foreignIdField` property only, please fix the tests to pass with all DB vendors. It seems the problem is an FK/unique key not added (and it might fix the MySQL issue then).

abbadon1334:
in another repo I solved it in this way:

Define a different Join Model::$_defaultSeedJoin
https://github.com/abbadon1334/atk-wordpress-plugin/blob/ccf5fb538f60e6cc513f4c148de5555bba7edd80/src/Models/PostModel.php#L26

Extend Join class
https://github.com/abbadon1334/atk-wordpress-plugin/blob/master/src/Models/Internal/WpModelJoin.php

Use it in this way:
https://github.com/abbadon1334/atk-wordpress-plugin/blob/ccf5fb538f60e6cc513f4c148de5555bba7edd80/src/Models/Internal/WpModel.php#L48-L64

It works.
What do you think if i try to rewrite the PR following this path?

mvorisek:
The CI issues here are related with DB specifics. They should be fixed and we can merge this PR. Then a different PR to join using real foreign Model can be made and disuccsed.

mvorisek (changes requested):
The src changes LGTM.

Here is a summary of the discussion above:
- https://github.com/atk4/data/pull/1077#issuecomment-1355676134 - no fake model must be used for saving, this needs to reviewed before merge
- https://github.com/atk4/data/pull/1077#issuecomment-1361385588 - in this PR, we probably do not want to do more, but tests need to pass with every DB engine - UPDATE: fixed by #1117

https://github.com/atk4/data/pull/1077#issuecomment-1436398690 - @abbadon1334 what usecases does it solve vs. this PR? Some short example?

mvorisek:
Thank you, @abbadon1334, for this contribution and sorry for letting you wait almost a year. Fixing the DBAL issues needed for proper FK creation - https://github.com/atk4/data/pull/1117 - was not easy.

PR #1084, opened 2023-01-06
Use 1:1 model class name as model caption fallback

mvorisek:
I do not want to use model table name as a fallback as it is often prefixed.

When model class name is used as a fallback, I belive it is better to return it 1:1 (incl. `\`) as it is then must more obvious where the data is comming from. Namespace is almost always present, so if `\` is not desired to be displayed to the end user, simply set model caption manually. If you already do, this change has no impact on your code.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -863,7 +863,7 @@ class Model implements \IteratorAggregate
      */
     public function getModelCaption(): string
     {
-        return $this->caption ?? $this->readableCaption(get_debug_type($this));
+        return $this->caption ?? get_debug_type($this);
     }
 
     /**
--- a/tests/RandomTest.php
+++ b/tests/RandomTest.php
@@ -334,12 +334,12 @@ class RandomTest extends TestCase
 
     public function testModelCaption(): void
     {
-        // caption is not set, so generate it from class name Model
+        // caption is not set, fallback to class name
         $m = new Model($this->db, ['table' => 'user']);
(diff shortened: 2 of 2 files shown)


mkrecek234:
@mvorisek We should not have a caption with leading \ as a default name. This makes UI very ugly (or force you always to set a caption manually which contradicts the agile development approach). I like the current approach as you can easily have multi-word table names which are beautifully displayed like user_address - for model class names we would then have to render spaces in it to transform UserAddress to User Address. Let us not give up the fast prototyping with good-looking UI - I do not see thus the advantage of this PR.

mvorisek:
> leading \

PHP class name (in canonical format) has no leading `\`

I am not fully keen on this PR, but for ex. if your model is internally under `Microsoft\ERP\Models` namespace, is `Microsoft ERP Models Xxx` really a suitaible caption to be ever displayed?

mkrecek234:
@mvorisek No, I was simply referring to today‘s feature that a table called user_address or customer_type is beautifully rendered ti a caption User Address or Customer Type. If we switch to model class name you would render CustomerAddress to Customer Address and so on. Surely I would not include the while class hierarchy in the caption if you switch to class name instead of table name

mvorisek:
I landed https://github.com/atk4/data/pull/1085/commits/8d2772babb2e2434eec5c07962d0e68acb878749 and closing this one. No change for non anonymous classes.

PR #1088, opened 2023-01-24
Allow Model::isset/getPersistence() calls on model only

mvorisek:
Currently Model is tightly coupled with one persistence.

persistence getters on entity are not wanted and `$this->_persistence` in `$entity->getPersistence()` was fetched using `__get`


mkrecek234 (approved):
LGTM

PR #1092, opened 2023-02-09
No implicit IN operator for condition with array value

mvorisek:
Value can be `array` which is not rare for `json` DBAL type, thus we cannot check if value is an array nor infer this from SQL expression.

### DIFF
--- a/docs/conditions.md
+++ b/docs/conditions.md
@@ -122,7 +122,7 @@ If you are building multiple conditions against the same field, you can use this
 format:
 
 ```
-$m->addCondition('name', ['John', 'Joe']);
+$m->addCondition('name', 'in', ['John', 'Joe']);
 ```
 
 For all other cases you can implement them with {php:meth}`Model::expr`:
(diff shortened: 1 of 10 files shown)


docs/conditions.md
```diff
@@ -383,7 +383,7 @@ Creates condition object based on provided arguments. It acts similar to Model::
 
 $key can be Model field name, Field object or Expression object
 $operator can be one of the supported operators >, <, >=, <=, !=, in, not in, like, not like, regexp, not regexp
-$value can be Field object, Expression object, array (interpreted as 'any of the values') or other scalar value
+$value can be Field object, Expression object or any scalar value
```
DarkSide666, line 386: maybe add words "including array" at the end?

DarkSide666 (approved):
LGTM, but it's quite bad BC break because can't be easily fixed by search/replace.

docs/conditions.md
mvorisek: For IN operator yes.

PR #1117, opened 2023-08-18
Fix unique index creation for MSSQL & Oracle

mvorisek:
fix MSSQL https://github.com/doctrine/dbal/issues/5507 - FK cannot be created if the foreign unique index was created with `WHERE xxx IS NOT NULL`

fix Oracle https://github.com/doctrine/dbal/issues/5508 - foreign unique index is not enough to create FK, unique table constraint needs to be present


mvorisek:
Thank you, @abbadon1334, for this contribution and sorry for letting you wait almost a year. Fixing the DBAL issues needed for proper FK creation - https://github.com/atk4/data/pull/1117 - was not easy.

PR #1120, opened 2023-08-22
Improve join temporary data cleanup

mvorisek:
a few changes has been landed sooner in #1119

### DIFF
--- a/src/Model/Join.php
+++ b/src/Model/Join.php
@@ -44,7 +44,7 @@ abstract class Join
     protected $kind;
 
     /** Weak join does not update foreign table. */
-    protected bool $weak = false;
+    public bool $weak = false;
 
     /**
      * Normally the foreign table is saved first, then it's ID is used in the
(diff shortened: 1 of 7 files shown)


src/Model/Join.php
```diff
@@ -545,17 +512,14 @@ public function afterInsert(Model $entity): void
             return;
         }
 
-        $id = $this->hasJoin() ? $this->getJoin()->getId($entity) : $entity->getId();
```
mvorisek, line 548: TODO: conduct more tests of nested joins

src/Model/Join.php
```diff
@@ -271,16 +268,25 @@ protected function init(): void
...
     {
         $this->onHookToOwnerEntity(Model::HOOK_AFTER_LOAD, \Closure::fromCallable([$this, 'afterLoad']));
-        $this->onHookToOwnerEntity(Model::HOOK_AFTER_UNLOAD, \Closure::fromCallable([$this, 'afterUnload']));
+
+        $createHookFxWithCleanup = function (string $methodName): \Closure {
+            return function (Model $entity, &...$args) use ($methodName): void {
+                try {
+                    $this->{$methodName}($entity, ...$args);
+                } finally {
+                    $this->unsetSaveBuffer($entity);
```
mvorisek, line 277: With non-reverse and reverse join, I do not think this is correct, as the non-reverse join will cleanup the save buffer in `Model::HOOK_BEFORE_INSERT` hook.

PR #1125, opened 2023-08-31
Add Model::loadBy() complex condition support

mvorisek:
also introduce `Model::createIteratorBy()` method

both methods accept exactly the same data as `Model::addCondition()` method


PhilippGrashoff:
Hi @mvorisek,

I wonder what the usecase of this is? At the moment, loading with complex conditions can be done by something like
```
$someModel->addCondition(...);
$someModel->addCondition(...);
$someModel->addCondition(...);
$someModel->addCondition(...);
$someModel->addCondition(...);
$someModel->loadAny(); //or loadOne()
```

What is the advantage of providing loadBy() with much more complex arguments?

Best regards
Philipp

mvorisek:
> What is the advantage of providing loadBy() with much more complex arguments?

@PhilippGrashoff the model (and result entity) conditions are not mutated, the "load by" conditions are used to restrict the condition during the lookup only

Issue #1130, opened 2023-09-13
Model::dirty not available in Model::HOOK_AFTER_SAVE spot


PhilippGrashoff:
The current decription of `Model::dirty` indicates that the dirty information is available in all hooks:
`* The $dirty data will be reset after you save() the data but it is still available to all before/after save handlers.`

However, that is not the case. `Model::dirty` is reset before the `Model::HOOK_AFTER_SAVE` spot is executed (taken from current develop):
```
            $dirtyRef = &$this->getDirtyRef();
            $dirtyRef = [];

            if ($this->idField && $this->reloadAfterSave) {
                $this->reload();
            }

            $this->hook(self::HOOK_AFTER_SAVE, [$isUpdate]);
```

It is not enough to move the reset after the hook, as `Model::reload()` also resets dirty. This is where is gets complicated: reloaded model information could be important in the after save hook, as does the dirty information does.

The only fix I can think of is to clone `$dirtyRef`, assign in to `$this` right before the hook is executed and then reset it afterwards. Something like:
```
            $dirtyRef = &$this->getDirtyRef();
            $dirtyRefCache = clone dirtyRef;

            if ($this->idField && $this->reloadAfterSave) {
                $this->reload();
            }

            $dirtyRef = $dirtyRefCache;
            $this->hook(self::HOOK_AFTER_SAVE, [$isUpdate]);
            $dirtyRef = [];
```

WDYT about this? If you are fine with this, I will create a PR with a change like this and add a test.

mvorisek:
You can capture the original dirty data in "before update" hook. The dirtiness is logical, before update, the data are dirty as they are. Once they are saved, they are no longer dirty.

PR #1131, opened 2023-09-14
Update docs about dirty fields and after save hook

PhilippGrashoff:
related with https://github.com/atk4/data/issues/1130

The old PHPDoc implied that $dirty data was available in Model::HOOK_AFTER_SAVE spot. This is not the case.

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -150,15 +150,16 @@ class Model implements \IteratorAggregate
 
     /**
      * After loading an active record from DataSet it will be stored in
-     * $data property and you can access it using get(). If you use
+     * $data property, and you can access it using get(). If you use
      * set() to change any of the data, the original value will be copied
      * here.
      *
      * If the value you set equal to the original value, then the key
      * in this array will be removed.
      *
-     * The $dirty data will be reset after you save() the data but it is
-     * still available to all before/after save handlers.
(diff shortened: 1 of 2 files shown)


src/Model.php
```diff
@@ -150,15 +150,16 @@ class Model implements \IteratorAggregate
...
+     * $data property, and you can access it using get(). If you use
      * set() to change any of the data, the original value will be copied
      * here.
      *
      * If the value you set equal to the original value, then the key
      * in this array will be removed.
      *
-     * The $dirty data will be reset after you save() the data but it is
-     * still available to all before/after save handlers.
+     * Behaviour in Model::save():
```
mvorisek: private variable does not need any docs, as the class code is enought together with the tests to document it

PhilippGrashoff: I am ok with both - either have a correct comment or no comment at all. Do as you like, I would prefer the correct comment as it describes the behaviour and so new users need to read less code.

mvorisek: The problem with too much phpdoc comments is they get easily outdated and only increase LoC.

If you want this to be documented, please update docs/. Please make sure the comment is crisply clear as the current proposal is not much helpful IMHO, as the described behaviour is (generally) expected and probably even already documented in docs/.

PR #1146, opened 2023-10-27
Unify SQL quoted token regex

mvorisek:


tests/Persistence/Sql/ExpressionTest.php
```diff
@@ -180,16 +180,16 @@ public function testNestedParams(): void
     {
         $e1 = $this->e('[] and []', [
             $this->e('++[]', [1]),
-            $this->e('--[]', [2]),
+            $this->e('**[]', [2]),
```
mvorisek, line 183: TODO test if comments and `[xxx]` inside quoted strings/identifiers are untouch

PR #1171, opened 2024-02-24
Store array persistence data using objects

mvorisek:


mvorisek:
WIP, needs usecase and funding, maybe SQLite might be used as internal backend instead.

mvorisek:
hash indexed lookup implemented in #1252

PR #1202, opened 2024-04-23
Impl. binary (case sensitive) LIKE/REGEXP for SQLite

mvorisek:
for binary/blob columns (all SQLite columns /wo COLLATION NOCASE)

the main motivation is to align the behaviour with MySQL

### DIFF
--- a/src/Persistence/Sql/Oracle/Query.php
+++ b/src/Persistence/Sql/Oracle/Query.php
@@ -18,10 +18,10 @@ class Query extends BaseQuery
     protected string $expressionClass = Expression::class;
 
     #[\Override]
-    protected function _renderConditionRegexpOperator(bool $negated, string $sqlLeft, string $sqlRight): string
+    protected function _renderConditionRegexpOperator(bool $negated, string $sqlLeft, string $sqlRight, bool $binary = false): string
     {
         return ($negated ? 'not ' : '') . 'regexp_like(' . $sqlLeft . ', ' . $sqlRight
-            . ', ' . $this->escapeStringLiteral('in') . ')';
+            . ', ' . $this->escapeStringLiteral(($binary ? '' : 'i') . 'n') . ')';
     }
 
     #[\Override]
--- a/src/Persistence/Sql/Oracle/Query.php
+++ b/src/Persistence/Sql/Oracle/Query.php
@@ -21,7 +21,7 @@ class Query extends BaseQuery
     protected function _renderConditionRegexpOperator(bool $negated, string $sqlLeft, string $sqlRight, bool $binary = false): string
     {
         return ($negated ? 'not ' : '') . 'regexp_like(' . $sqlLeft . ', ' . $sqlRight
-            . ', ' . $this->escapeStringLiteral(($binary ? '' : 'i') . 'n') . ')';
+            . ', ' . $this->escapeStringLiteral(($binary ? 'c' : 'i') . 'n') . ')';
     }
 
     #[\Override]
--- a/src/Persistence/Sql/Query.php
+++ b/src/Persistence/Sql/Query.php
@@ -501,6 +501,50 @@ abstract class Query extends Expression
     }
 
     /**
+     * @param \Closure(string, string): string $makeSqlFx
+     */
+    protected function _renderConditionBinaryReuse(
+        string $sqlLeft,
+        string $sqlRight,
+        \Closure $makeSqlFx,
+        bool $allowReuseLeft = true,
+        bool $allowReuseRight = true,
+        string $internalIdentifier = 'reuse'
+    ): string {
+        $nonTrivialSqlRegex = '~\s|\(~';
+        $subqueryLeftColumnSql = $allowReuseLeft && preg_match($nonTrivialSqlRegex, $sqlLeft)
+            ? $this->escapeIdentifier('__atk4_' . $internalIdentifier . '_left__')
+            : null;
+        $subqueryRightColumnSql = $allowReuseRight && preg_match($nonTrivialSqlRegex, $sqlRight)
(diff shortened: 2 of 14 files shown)


src/Persistence/Sql/Oracle/Query.php
```diff
@@ -18,10 +18,10 @@ class Query extends BaseQuery
     protected string $expressionClass = Expression::class;
 
     #[\Override]
-    protected function _renderConditionRegexpOperator(bool $negated, string $sqlLeft, string $sqlRight): string
+    protected function _renderConditionRegexpOperator(bool $negated, string $sqlLeft, string $sqlRight, bool $binary = false): string
     {
         return ($negated ? 'not ' : '') . 'regexp_like(' . $sqlLeft . ', ' . $sqlRight
-            . ', ' . $this->escapeStringLiteral('in') . ')';
+            . ', ' . $this->escapeStringLiteral(($binary ? '' : 'i') . 'n') . ')';
```
DarkSide666: Shouldn't here be 'cn' generated if it's binary (case-sensitive) matching?
As described in https://docs.oracle.com/cd/B13789_01/server.101/b10759/conditions018.htm
_The default case sensitivity is determined by the value of the NLS_SORT parameter._

So I guess we better set `in` if not case-sensitive and `cn` if it should be case-sensitive no matter of Oracle NLS_SORT parameter.

src/Persistence/Sql/Query.php
```diff
@@ -528,10 +572,10 @@ protected function _renderConditionLikeOperator(bool $negated, string $sqlLeft,
             . ' escape ' . $this->escapeStringLiteral('\\');
     }
 
-    protected function _renderConditionRegexpOperator(bool $negated, string $sqlLeft, string $sqlRight): string
+    protected function _renderConditionRegexpOperator(bool $negated, string $sqlLeft, string $sqlRight, bool $binary = false): string
     {
         return ($negated ? 'not ' : '') . 'regexp_like(' . $sqlLeft . ', ' . $sqlRight
-            . ', ' . $this->escapeStringLiteral('is') . ')';
+            . ', ' . $this->escapeStringLiteral(($binary ? '' : 'i') . 's') . ')';
```
DarkSide666, line 578: maybe should force case-sensivity instead of relying on default

src/Persistence/Sql/Sqlite/CreateRegexpLikeFunctionMiddleware.php
```diff
@@ -30,18 +30,13 @@ public function connect(
...
 
-                    if (is_int($value)) {
-                        $value = (string) $value;
-                    } elseif (is_float($value)) {
-                        $value = Expression::castFloatToString($value);
-                    }
+                    $value = CreateRegexpLikeFunctionMiddleware::castScalarToString($value);
 
-                    $isValidUtf8Value = \PHP_VERSION_ID < 80200
-                        ? preg_match('~~u', $value) === 1 // much faster in PHP 8.1 and lower
```
DarkSide666: is this performance flaw fixed in newer PHP 8.2, 8.3 versions?

src/Persistence/Sql/Sqlite/CreateRegexpReplaceFunctionMiddleware.php
```diff
@@ -30,18 +30,14 @@ public function connect(
...
 
-                    if (is_int($value)) {
-                        $value = (string) $value;
-                    } elseif (is_float($value)) {
-                        $value = Expression::castFloatToString($value);
-                    }
+                    $value = CreateRegexpLikeFunctionMiddleware::castScalarToString($value);
 
-                    $isValidUtf8Value = \PHP_VERSION_ID < 80200
-                        ? preg_match('~~u', $value) === 1 // much faster in PHP 8.1 and lower
```
DarkSide666: is this performance flaw fixed in newer PHP 8.2, 8.3 versions?

src/Persistence/Sql/Sqlite/CreateRegexpLikeFunctionMiddleware.php
mvorisek: It is/was a hack I invented for atk4/ui, but here we use it less often. I will recheck the performance here with some real data...

mvorisek: should be fine - https://3v4l.org/aO7XS - 100k strings take ~20ms

wrapping the check in a function with both variants would imply function call overhead - https://3v4l.org/tanVc

src/Persistence/Sql/Sqlite/CreateRegexpReplaceFunctionMiddleware.php
mvorisek: discussed in https://github.com/atk4/data/pull/1202#discussion_r1578233748

src/Persistence/Sql/Sqlite/CreateRegexpLikeFunctionMiddleware.php
mvorisek: should be fine - https://3v4l.org/QaZpS - 100k strings take ~20ms

but I have landed the original speedup hack - https://github.com/atk4/data/pull/1203/commits/b9e189d7ee1389db18c16e92bcc9b04459eb82df

src/Persistence/Sql/Oracle/Query.php
mvorisek: `c` if not-CI added

src/Persistence/Sql/Query.php
mvorisek: impossible - this is generic/non vendored DB Query class and regexp implementations which uses PCRE impl. are CS by default /wo any flag needed, other DBs uses `c` flag, but no all DBs support such flag...

PR #1228, opened 2024-08-05
Build epub and htmlzip docs

Jmos:
Adding epub and html


docs/.readthedocs.yaml
```diff
@@ -14,3 +14,5 @@ python:
 
 formats:
```
mvorisek, line 15: docs: https://docs.readthedocs.io/en/stable/config-file/v2.html#formats

mvorisek:
Why not, please send a PR into https://github.com/atk4/core and https://github.com/atk4/ui as well. Might I know how did you come to this PR, do you prefer epub or htmlzip?

Jmos:
I hope all is done pls Check

PR #1241, opened 2025-02-07
Add SQL select statement in debug info for SQL persistence

DarkSide666:


mvorisek:
I am not fan of this change. `print_r($complexUiComponent)` should not print models incl. queries. The output will be too long to read.

Might I know your usecase, what was hard to debug, some model with `Model::$cteModels`?

DarkSide666:
In my case it was actually a very simple data model with few conditions which I was trying to debug for ~2h because it was so simple that didn't see obvious error. Dumping actual SQL helped me a lot to see what's wrong with it.
Problem was that I completely forgot how to get this debug query by `$model->action('select')->getDebugQuery()`. So I tought that maybe there should be simpler method to be able to see actual query.
At some point we also had `$model->debug=true` option which did dump on screen all selects executed by particular model object ... That was quite straightforward too.

About your comment - I see your concern about very long sql in UI. That's a good point indeed.
On other hand, that's a debug info and normally users should never ever see it in live app.

mvorisek:
I am closing the PR because of the reason explained and also because this is againt OOP - we should not refer specific persistence in Model.

I would however welcome docs improvement or even main README direcly.

PR #1264, opened 2025-09-28
Always use FQN when field is referenced via Model::expr()

mvorisek:


tests/ReferenceSqlTest.php
```diff
@@ -535,8 +535,8 @@ public function testOtherAggregates(): void
                 'items_code' => ['aggregate' => 'count', 'field' => 'code', 'type' => 'integer'], // counts only not-null values
                 'items_star' => ['aggregate' => 'count', 'type' => 'integer'], // no field set, counts all rows with count(*)
                 'items_c:' => ['concat' => '::', 'field' => 'name'],
-                'items_c-' => ['aggregate' => $i->getPersistence()->dsql()->groupConcat($i->expr('[name]'), '-')],
-                'len' => ['aggregate' => $i->expr('SUM(' . $makeLengthSqlFx('[name]') . ')'), 'type' => 'integer'],
+                'items_c-' => ['aggregate' => $i->getPersistence()->dsql()->groupConcat($l->getReference('Items')->createTheirModel()->expr('[name]'), '-')],
```
mvorisek, line 538: Ideally we should preserve the model instance in Expression and assert later that the instance is present in the whole SQL (owner model) - however this will fail after the owner model is cloned...

src/Persistence/Sql.php
```diff
@@ -153,7 +153,12 @@ static function ($matches) use ($model, &$arguments) {
 
                 $identifier = substr($matches[0], 1, -1);
                 if ($identifier !== '' && !isset($arguments[$identifier])) {
-                    $arguments[$identifier] = $model->getField($identifier);
+                    // TODO horrible hack to render the field with a table prefix,
+                    // find a solution how to wrap the field inside custom Field (without owner?)
+                    $modelCloned = clone $model;
+                    $modelCloned->persistenceData['use_table_prefixes'] = true;
+
+                    $arguments[$identifier] = $modelCloned->getField($identifier);
```
mvorisek, line 161: All fields needs to be aliased when CTE is used too - https://dbfiddle.uk/R7lgFv8w.

PR #1266, opened 2025-10-04
Fix Model::getReference() after traversing contained model

mvorisek:
needed for https://github.com/atk4/ui/pull/2297

### DIFF
--- a/src/Reference.php
+++ b/src/Reference.php
@@ -243,9 +243,7 @@ class Reference
         $ourModel = $this->getOurModel();
 
         if ($ourModel->containedInPersistence !== null) {
-            if ($ourModel->containedInEntity !== null) {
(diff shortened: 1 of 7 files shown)


src/Reference.php
```diff
@@ -242,11 +242,12 @@ protected function getDefaultPersistence()
     {
         $ourModel = $this->getOurModel();
 
-        // this is useful for ContainsOne/Many implementation in case when you have
-        // SQL_Model->containsOne()->hasOne() structure to get back to SQL persistence
-        // from Array persistence used in ContainsOne model
-        if ($ourModel->containedInEntity !== null && $ourModel->containedInEntity->getModel()->issetPersistence()) {
-            return $ourModel->containedInEntity->getModel()->getPersistence();
+        if ($ourModel->containedInPersistence !== null) {
```
mvorisek, line 245: contained in contained should be supported

PR #1299, opened 2026-07-07
Parse DSN port as integer

DarkSide666:


DarkSide666:
Otherwise we get this error:
`TypeError: mysqli::real_connect(): Argument #5 ($port) must be of type ?int, string given`

image

mvorisek:
Makes sense! 👍

PR #1300, opened 2026-07-17
Fix array persistence order

PhilippGrashoff:


PhilippGrashoff:
Removing `array_reverse()` in `Model::setOrder()` makes this test work, but others fail. @mvorisek this might be one for you?

mvorisek:
@DarkSide666 what is your opinion on this change? @abbadon1334 and @mkrecek234 tagging you guys as well.

mvorisek:
Thank you for the test case. After #1308 the issue was easy to find and fix.

PR #1303, opened 2026-08-14
Add retry mechanism when connecting to Oracle

DarkSide666:
This fixes issue caused by Oracle listener which we also often see in Atk tests.

On heavier connection load Oracle randomly becomes inaccessible because Oracle listener have some delay before it actually drops connection out of its cache after you disconnect from db. As result listener kind of "overflows" and don't allow to make new connections to DB even if previous connections are all nicely closed and not hit DBs connection limit.

Implementation have no BC breaks as connection retries are done purely _on background_.

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
   schedule:
-    - cron: '0 0/2 * * *'
+    - cron: '0 0/8 * * *'
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -162,7 +162,6 @@ jobs:
           php -r '(new PDO("mysql:host=mysql5", "root", "atk4_pass_root"))->exec("GRANT USAGE ON *.* TO '"'"'atk4_test_user'"'"'@'"'"'%'"'"' WITH MAX_USER_CONNECTIONS 5");'
           php -r '(new PDO("mysql:host=mariadb", "root", "atk4_pass_root"))->exec("ALTER USER '"'"'atk4_test_user'"'"'@'"'"'%'"'"' WITH MAX_USER_CONNECTIONS 5");'
           php -r '(new PDO("pgsql:host=postgres;dbname=atk4_test", "atk4_test_user", "atk4_pass"))->exec("ALTER ROLE atk4_test_user CONNECTION LIMIT 1");'
-          /usr/lib/oracle/setup.sh
           if [ -n "$LOG_COVERAGE" ]; then mkdir coverage; echo "memory_limit = 192M" > /usr/local/etc/php/conf.d/custom-memory-limit.ini; fi
 
       - name: "Run tests: SQLite"
@@ -245,6 +244,12 @@ jobs:
         run: |
           php -d opcache.enable_cli=1 vendor/bin/phpunit --exclude-group none $(if [ -n "$LOG_COVERAGE" ]; then echo --coverage-text; else echo --no-coverage; fi) --fail-on-warning --fail-on-risky $(if vendor/bin/phpunit --version | grep -q '^PHPUnit 9\.'; then echo -v; else echo --fail-on-notice --fail-on-deprecation --display-notices --display-deprecations --display-warnings --display-errors --display-incomplete --display-skipped; fi)
           if [ -n "$LOG_COVERAGE" ]; then mv coverage/phpunit.cov coverage/phpunit-mssql.cov; fi
+
+      - name: "Setup and wait for healthy Oracle service"
+        if: success() || failure()
+        run: |
+          /usr/lib/oracle/setup.sh
+          php -r '$t = microtime(true); while (microtime(true) - $t < 300) { try { (new PDO("oci:host=oracle;dbname=oracle/freepdb1", "system", "atk4_pass_root")); exit; } catch (Exception $e) {} sleep(1); } exit(2);'
 
       - name: "Run tests: Oracle - PDO (only for coverage or cron)"
         if: (success() || failure()) && (env.LOG_COVERAGE || github.event_name == 'schedule')
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -110,16 +110,19 @@ jobs:
           POSTGRES_USER: atk4_test_user
           POSTGRES_PASSWORD: atk4_pass
           POSTGRES_DB: atk4_test
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+        options: --health-cmd=pg_isready --health-interval=10s --health-timeout=5s --health-retries=5
       mssql:
         image: mcr.microsoft.com/mssql/server:${{ matrix.type == 'Phpunit Lowest' && 'latest' || '2017-latest' }}
         env:
           ACCEPT_EULA: Y
           SA_PASSWORD: atk4_pass
+        options: --health-cmd="/opt/mssql-tools/bin/sqlcmd -S 127.0.0.1 -C -U sa -P atk4_pass -b -Q 'SELECT 1' || /opt/mssql-tools18/bin/sqlcmd -S 127.0.0.1 -C -U sa -P atk4_pass -b -Q 'SELECT 1'" --health-interval=10s --health-timeout=5s --health-retries=5
       oracle:
         image: gvenzl/${{ matrix.type == 'Phpunit Lowest' && 'oracle-free:slim-faststart' || 'oracle-xe:18-slim-faststart' }}
         env:
-          ORACLE_PASSWORD: atk4_pass
+          ORACLE_PASSWORD: atk4_pass_root
+          APP_USER: atk4_test_user
+          APP_USER_PASSWORD: atk4_pass
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -246,8 +249,8 @@ jobs:
       - name: "Run tests: Oracle - PDO (only for coverage or cron)"
         if: (success() || failure()) && (env.LOG_COVERAGE || github.event_name == 'schedule')
         env:
-          DB_DSN: "pdo_oci:dbname=oracle/free"
-          DB_USER: system
+          DB_DSN: "pdo_oci:dbname=oracle/freepdb1"
+          DB_USER: atk4_test_user
           DB_PASSWORD: atk4_pass
           NLS_LANG: AMERICAN_AMERICA.AL32UTF8
         run: |
@@ -257,8 +260,8 @@ jobs:
       - name: "Run tests: Oracle - OCI8"
         if: success() || failure()
         env:
-          DB_DSN: "oci8:dbname=oracle/free"
-          DB_USER: system
+          DB_DSN: "oci8:dbname=oracle/freepdb1"
+          DB_USER: atk4_test_user
           DB_PASSWORD: atk4_pass
           NLS_LANG: AMERICAN_AMERICA.AL32UTF8
         run: |
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -120,14 +120,6 @@ jobs:
         image: gvenzl/${{ matrix.type == 'Phpunit Lowest' && 'oracle-free:slim-faststart' || 'oracle-xe:18-slim-faststart' }}
         env:
           ORACLE_PASSWORD: atk4_pass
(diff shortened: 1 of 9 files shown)


src/Persistence/Sql/Oracle/RetryConnectionMiddleware.php
```diff
@@ -0,0 +1,59 @@
...
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Driver\Connection;
+use Doctrine\DBAL\Driver\Middleware;
+use Doctrine\DBAL\Driver\Middleware\AbstractDriverMiddleware;
+use Doctrine\DBAL\Driver\OCI8\Exception\ConnectionFailed;
+
+/**
+ * Retries transient Oracle listener failures during connection establishment.
+ *
+ * Oracle may temporarily return ORA-12516 when the listener has not yet
```
mvorisek: Are you sure? IIRC this issue is present even with very few requests. What is the real connection count limit?

(and btw. this is already an issue with MySQL, but limit of small number of connections instead of strict 1 is enough and reliable)

src/Persistence/Sql/Oracle/RetryConnectionMiddleware.php
```diff
@@ -0,0 +1,59 @@
...
+ * a short delay allows the listener to refresh its handler state.
+ */
+class RetryConnectionMiddleware implements Middleware
+{
+    #[\Override]
+    public function wrap(Driver $driver): Driver
+    {
+        return new class($driver) extends AbstractDriverMiddleware {
+            // ORA-12516: TNS:listener could not find available handler with matching protocol stack
+            private const RETRY_ERROR_CODES = [12516];
```
mvorisek: In Discord you mentioned https://dbamarco.wordpress.com/2023/07/20/suspicous-ora-12516-by-oracle-scan-listener/ . I want to check if this desctribe the issue. In the chat, he wrote they increased the connection limit... Did you see this issue in production somewhen?

.github/workflows/test-unit.yml
```diff
@@ -246,9 +254,10 @@ jobs:
       - name: "Run tests: Oracle - PDO (only for coverage or cron)"
         if: (success() || failure()) && (env.LOG_COVERAGE || github.event_name == 'schedule')
         env:
-          DB_DSN: "pdo_oci:dbname=oracle/free"
-          DB_USER: system
-          DB_PASSWORD: atk4_pass
+          # connection to portable database
+          DB_DSN: "pdo_oci:dbname=oracle/${{ matrix.type == 'Phpunit Lowest' && 'freepdb1' || 'xepdb1' }}"
```
mvorisek: The issues are mostly/solely present with "lowest". Ie. the lowest/oldest deps or the newest db. I would be happy to have this explained - is this because of the newest/never db and did some db limit changed?

.github/workflows/test-unit.yml
```diff
@@ -120,6 +120,14 @@ jobs:
         image: gvenzl/${{ matrix.type == 'Phpunit Lowest' && 'oracle-free:slim-faststart' || 'oracle-xe:18-slim-faststart' }}
         env:
           ORACLE_PASSWORD: atk4_pass
+          APP_USER: atk4_test_user
+          APP_USER_PASSWORD: atk4_user_pass
+        # Provide healthcheck script options for startup
+        options: >-
```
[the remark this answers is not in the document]
DarkSide666: APP user and password are used to run tests with Oracle user with lower privileges and on portable database (freepdb1 or xepdb1) instead of using system user.
Healthcheck script makes sure Oracle is fully started before starting tests phase. In our case theoretically we can live without that as Oracle tests are not the first one to execute, but ... i think we better have this script to make it all a bit more reliable. Normally that takes just a few seconds more.

.github/workflows/test-unit.yml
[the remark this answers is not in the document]
DarkSide666: Tests are more correct to run with normal db user instead of elevated built-in system user.
So say that it may even help with performance a bit (user have less permissions to validate against), but i didn't notice that in my tests. Maybe just unnoticable 0.01% improvement somewhere.

src/Persistence/Sql/Oracle/RetryConnectionMiddleware.php
DarkSide666: Increasing connection limit is simplest workaround. Of course if you have 1000000 connections limit, then you'll not hit it while testing and all should be fine. But normally there shouldn't be such high limit. I've tested that all here in last few days. It's really that - in case of this failure just retry in 10-100ms (rarely a bit more - up till about 0.5s) and listener cache is updated, connections are freed and tests continue.
That wordpress topic is about slightly different kind of case, but still was the one to give me idea where to look at.

src/Persistence/Sql/Oracle/RetryConnectionMiddleware.php
DarkSide666: Current limits were:
Oracle-free (23) - 200 processes, 322 sessions
Oracle-xe (18) - 320 processes, 502 sessions

About 60-90 sessions can be already used internally by Oracle background services.

You can see more stats here https://github.com/atk4/data/actions/runs/31742995074/job/94590708915
Check steps Oracle stats before tests, Oracle stats after tests. These steps were added while testing and not included in this final PR. Also listener hitting limits is not visible in these stats because when listener hits limits, then it doesn't reach DB at all, so it's not visible in DB side of stats. It's only visible in listener log file, but it's stored on container which is not accessible from github job.

.github/workflows/test-unit.yml
DarkSide666: See https://github.com/atk4/data/pull/1303#discussion_r3789124670
I think it's because oracle-fee (v23 which is used in lowest tests) have 322 sessions limit and 200 processes limit by default but old oracle xe has 502 and 320 respectively.

src/Persistence/Sql/Oracle/RetryConnectionMiddleware.php
DarkSide666: Also you can check step "Stress test Oracle connections" in this job https://github.com/atk4/data/actions/runs/31723291851/job/94525337653
You can see there how it hits the limits randomly, then retries one or few times and then continue successfully.

.github/workflows/test-unit.yml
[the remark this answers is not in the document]
mvorisek: Please send this change and also for MSSQL\* as a separate PR and lets measure the CI slowdown. Once tested and landed, lets talk this PR. Thanks.

\* see https://github.com/atk4/ui/actions/runs/31930717744/job/95124971805 and https://github.com/atk4/ui/actions/runs/31776615590/job/94693273172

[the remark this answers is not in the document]
mvorisek: I also wonder if the service's `healthcheck.sh` can be called later, ie. before the specific database step which will give the service time to startup when other DBs are tested.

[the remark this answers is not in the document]
DarkSide666: Ok I'll make separate PR for MSSQL, but not sure what you want me to do with this PR. Should I merge it in? I guess so, right?
Also I'm not sure why you worry so much about adding or not adding healthcheck. There should be healthcheck for (almost) all services as we have to be sure services actually work before we start running tests on them. So MSSQL should have healthcheck too. Probably also Postgre.
Also, speaking about connection retry mechanism, I believe it could be useful for all DB engines to work out of the box.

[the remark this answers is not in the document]
DarkSide666: I've moved healtchecks to separate PRs:
- MSSQL https://github.com/atk4/data/pull/1304
- Oracle https://github.com/atk4/data/pull/1305

.github/workflows/test-unit.yml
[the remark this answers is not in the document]
DarkSide666: Moved to separate PR https://github.com/atk4/data/pull/1306

.github/workflows/test-unit.yml
mvorisek: Github CI is dummy, all services are started one by one and all healtchecks are busy waited before tests. If a service starts slow, it will slowdown the CI.

Thanks and I will check this on the separate PRs. In past, I opted out for the healtchecks as the tests on other DBs were realiably enough without having to busy wait.

Please do not merge this PR yet, I want to land the other changes first and verify if retry mechanism is the only solution, I personally do not like it but on the other side Oracle is very ugly software and it might be the best working solution.

src/Persistence/Sql/Oracle/RetryConnectionMiddleware.php
mvorisek: The proper solution might be to release the connection better. It might be even a bug in php driver.

Does this happen with pdo_oci as well as oci8?

A solution worth to consider might be even to force to release the connection using destructor.

DarkSide666: It's not a bug or something in php.
It's how Oracle Listener works. When you try to connect Oracle, then you first hit the listener and listener keeps track of active sessions too. If listener thinks that there are more active sessions than Oracle DB allows, then it do not pass you trough to database at all.
Listener has a tiny cache of that sessions stats and when you disconnect it doesn't instantly register that in its cache. So in very busy connect/disconnect load listener stays a bit behind.
Simple retrying in few quite **small** intervals is a way to go and solves everything.

.github/workflows/test-unit.yml
mvorisek: I see you are collecting metrics. I did a small test today and it seems many reconnections are not an issue.

I tried 100k connections and never hit the issue.

I would be happy to have small reproducible CI of the "12516 error" first.

I belive this issue is present due some Oracle quirk and we should address it more directly.

If you think the lowered sessions/processes limits are helpful to reproduce it, you might even lower them. Or monitor after which tests they increase rapidly.

mvorisek: I trigged about 1000 pipelines with all Oracle versions and I have found:
- the issue is present as you have written - with any many connections, no special query needed
- the affected versions are >=23.4. With 23.3 or lower I was not able to trigger the issue at all

The retry solution might be the only one. I want to still do some more experiments.

Also, have the 322/502 200/320 limits changed with version 23.3 -> 23.4? If not, what changed?

DarkSide666: Well. That could be true about 23.3 and 23.4 as from 23.4 Oracle introduced AI Vector Search & Data Types - Version 23.4 officially introduced the native VECTOR data type and AI vector search features, which were not fully packaged or active in 23.3 developer builds.
Also they did Container and Setup Adjustments - Docker and container setup scripts (like network binding for tnsnames.ora) were updated and corrected in 23.4 compared to the 23.3 free/developer images.

I can't guarantee that this AI answer is correct, but that's approximately that.
image
image
image
image
image

But anyway, even if any of these solutions work, I still think that retry middleware is good to have as a fallback mechanism. No long retries, just try few times in quite short intervals and that's it.
It could be even helpful for other DB engines too to avoid some small sudden glitches while connecting.

mvorisek: I cannot agree with the last paragraph. If 1 concurrent connection can exceed the total connections limit of 200, there is something very wrong with the database.

If there is a possibility of fixing this by some Oracle configuration, I might prefer it. I mean of course not by raising the concurrency to sky...

DarkSide666: Yes that would be perfect to fix it by Oracle settings of course.
But that doesn't mean that retry mechanism as such is a bad idea.
Normally you would not use Oracle as DB for busy webpages which does connect/disconnect all the time. Then Mysql or even Postgre is much better choice. Oracles power is more in data consistency, complex data cases, large data amounts etc. Also preferably by using persistent connections or connection pool at least.

P.S. Also Oracle uses like maybe 50 connections internally for its own services. So it's not 200 free connections at a start.

src/Persistence/Sql/Oracle/RetryListenerConnectionMiddleware.php
```diff
@@ -0,0 +1,55 @@
...
+
+            #[\Override]
+            public function connect(
+                #[\SensitiveParameter]
+                array $params
+            ): Connection {
+                for ($attempt = 0;; ++$attempt) {
+                    try {
+                        return parent::connect($params);
+                        // @codeCoverageIgnoreEnd
```
DarkSide666: Probably it should be `@codeCoverageIgnoreStart` here

mvorisek: We good catch!

PR #1305, opened 2026-08-16
Add healthcheck for Oracle testing

DarkSide666:


.github/workflows/test-unit.yml
```diff
@@ -120,6 +120,11 @@ jobs:
         image: gvenzl/${{ matrix.type == 'Phpunit Lowest' && 'oracle-free:slim-faststart' || 'oracle-xe:18-slim-faststart' }}
         env:
           ORACLE_PASSWORD: atk4_pass
+        options: >-
```
mvorisek: This change/PR increases the "Initialize containers" CI step by 30-60 seconds.

I would be happy if this healthcheck can be moved to the Oracle testing step directly as it should be alsmost for free there...

DarkSide666:
Moved Oracle healthcheck from initialization phase to right before tests.

mvorisek:
Something like this, it is ugly ATM, I will review this PR later as not immediately needed, as the non-Oracle tests give enough time for Oracle to startup.

mvorisek:
Thank you, I really like all services now covered with healchecks ❤

PR #1306, opened 2026-08-16
Use ordinary user for Oracle testing

DarkSide666:
Use ordinary user and portable database for Oracle tests instead of elevated system user.

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -248,7 +248,7 @@ jobs:
       - name: "Run tests: Oracle - PDO (only for coverage or cron)"
         if: (success() || failure()) && (env.LOG_COVERAGE || github.event_name == 'schedule')
         env:
-          DB_DSN: "pdo_oci:dbname=oracle/free"
+          DB_DSN: "pdo_oci:dbname=oracle/${{ matrix.type == 'Phpunit Lowest' && 'freepdb1' || 'xepdb1' }}"
           DB_USER: atk4_test_user
           DB_PASSWORD: atk4_pass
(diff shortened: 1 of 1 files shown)


.github/workflows/test-unit.yml
```diff
@@ -246,9 +248,10 @@ jobs:
       - name: "Run tests: Oracle - PDO (only for coverage or cron)"
         if: (success() || failure()) && (env.LOG_COVERAGE || github.event_name == 'schedule')
         env:
-          DB_DSN: "pdo_oci:dbname=oracle/free"
-          DB_USER: system
-          DB_PASSWORD: atk4_pass
+          # connection to portable database
+          DB_DSN: "pdo_oci:dbname=oracle/${{ matrix.type == 'Phpunit Lowest' && 'freepdb1' || 'xepdb1' }}"
```
mvorisek: I support the non-elevated user and the whole DB/table lifecycle should be tested with non-elevated user like with other DBs.

But why the DB name change, ie. why to use PDB instead of routed DB/service?

mvorisek: I see, the PDB change is needed for the ordinary user... 👍

DarkSide666:
I think older Oracle XE didn't have `freepdb1` database. I had `xepdb1`.

mvorisek:
The https://github.com/gvenzl/oci-oracle-xe images have, maybe because of my past feature forward-compatibility request ;-)

DarkSide666:
Oh, great !!! I didn't notice that.
Here it is https://github.com/gvenzl/oci-oracle-xe/pull/239

PR #1307, opened 2026-08-31
Add noChanges parameter in AfterSave hook

DarkSide666:
In AfterSave hook there is no way to tell if any data fields were changed (updated) or not as in such case AfterUpdate hook is not fired at all.
So I think that it would be useful to pass $noChanges value, which is already calculated before, as parameter to AfterSave hook.
I've at least one use-case for it already :)

### DIFF
--- a/src/Model.php
+++ b/src/Model.php
@@ -1616,7 +1616,7 @@ class Model implements \IteratorAggregate
                 $this->reload();
             }
 
-            $this->hook(self::HOOK_AFTER_SAVE, [$isUpdate]);
+            $this->hook(self::HOOK_AFTER_SAVE, [$isUpdate, $noChanges]);
 
             if ($this->idField) {
                 $this->validateEntityScope();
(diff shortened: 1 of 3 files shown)


mvorisek:
Might I know the usecase? The info can be get from "HOOK_AFTER_UPDATE" - ie. if called or not.

DarkSide666:
I'm working on small audit library (our atk4/audit is long time deprecated) and it attach itself to multiple model hooks.

Problem is that I have to execute some code in case model entity update was called, but no actual data updated.
So I can do that only in AfterSave hook as AfterUpdate hook will not be called in such case. But AfterSave hook callback knows nothing about dirty fields, old data, was there update or not etc.

Only thing I can imagine how to work around this is:
- in BeforeSave set some custom model property `dataChanged=false`
- in AfterUpdate set property to `dataChanged=true`. It'll execute only if data was changed
- in AfterSave check if dataChanged=false, then no actual data was updated

That's possible as workaround, but involves custom model property and is a bit hacky way to do that.

mvorisek:
Why not :) Thank you!

PR #1309, opened 2026-09-17
Add alter feature to Migrator

DarkSide666:
So far you could only add fields to table when creating new table:
```
        $this->createMigrator()
            ->table('t')
            ->field('a')
            ->create();
```

Now you can also add more fields to already existing table:
```
        $this->createMigrator()
            ->table('t')
            ->field('b', ['type' => 'string'])
            ->alter();
```

Please note that `alter()` intentionally supports only adding new fields in existing table and not modifying or deleting existing fields as these actions are to dangerous to do and you can accidentally loose data in DB.


src/Schema/Migrator.php
```diff
@@ -153,6 +153,54 @@ public function create(): self
         return $this;
     }
 
+    /**
+     * Currently only supports adding new columns as we don't want to risk with destructive actions.
+     */
+    public function alter(): self
```
mvorisek, line 159: What is this for? DBAL can already do diff/alter, we need to only pass the DBAL Table created from our model, don't we?

DarkSide666: You can't call `create` on existing table.

I mean, I had situation when i needed to add new columns to table which already exists DB by migrator. Only thing I could currently do is to drop table and create new one from model. That way I would loose data.

